PR #43, opened 2014-02-28
Fix VectorLayer components look up in wms server

bielfrontera:
Fixes a bug searching the two components in a vector layer request.
We have found this problem after adding QC flags to our dataset.

This is our scenario:

We have variables with these standard names:
- eastward_sea_water_velocity
- eastward_sea_water_velocity status_flag
- northward_sea_water_velocity
- northward_sea_water_velocity status_flag

It follows the CF-conventions (see
http://cf-pcmdi.llnl.gov/documents/cf-conventions/1.6/cf-conventions.html#flags )

In this scenario, sea_water_velocity is listed as an available layer of the dataset.
But when this layer is requested, we get this exception:
    uk.ac.rdg.resc.ncwms.exceptions.LayerNotDefinedException:
    The layer "sea_water_velocity" is not provided by this server

The problem was getting the layer components. The function isComponent was returning
both eastward_sea_water_velocity and eastward_sea_water_velocity status_flag as
sea_water_velocity components.

We have modified isComponent function in order to return only the variables which
match with the layer requested. With these changes, ThreddsDataset.getLayerComponents
is compatible with what WMSUtils.findVectorLayers does. According to this, a better
solution would be to move this functionality to WMSUtils, so both functions would be
in the same package.

Note: you can see this problem at
http://thredds.socib.es/thredds/godiva2/godiva2.html?server=http://thredds.socib.es/thredds/wms/hf_radar/hf_radar_ibiza-scb_codarssproc001/L1/2013/dep0001_hf-radar-ibiza_scb-codarssproc001_L1_2013-12.nc

### DIFF
--- a/tds/src/main/java/thredds/server/wms/ThreddsDataset.java
+++ b/tds/src/main/java/thredds/server/wms/ThreddsDataset.java
@@ -330,35 +330,35 @@ public class ThreddsDataset implements Dataset {
   }
 
   /**
-   * Returns true if the layerName contains some of the standard components prefixes for CF-1.0 or
-   * Grib convention
-   */
+   * Returns true if varAtt is the name of a scalar component of the vector layer layerName
+   *
+   * Check if the result of removing standard components prefixes for CF-1.0 or
+   * Grib convention from varAtt is the layerName.
+   *
+   * Handled standard components:
+   *    - x, y
+   *    - eastward, northward
+   *    - Meridional, Zonal
+   *    - u-component of, v-component of
+   *
+   * They can appear at the begining, in the middle or at the end, separated by _ or space characters
+   *
+   * @param layerName
+   * @param varAtt
+   * @return
+    */
   static boolean isComponent(String layerName, String varAtt) {
 
-    if (varAtt.contains("eastward_") || varAtt.contains("northward_")) {
-      //CF-Conventions
-      String[] tokens = layerName.split("_");
+    String componentRegex = "(x|y|eastward|northward|meridional|zonal|u-component of|v-component of)";
+    String separatorRegex = "(_|\\s+)";
+    String lookBehindRegex = "(?<=_|\\s|^)";
 
-      StringBuilder sb = new StringBuilder();
+    // Regex in two cases:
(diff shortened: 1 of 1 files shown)


kristiansebastian:
Hi,

This fix hasn't be added to the release 4.6.1 as you can see https://github.com/Unidata/thredds/blob/v4.6.1/tds/src/main/java/thredds/server/wms/ThreddsDataset.java

Please, Could you add the fix to the next release?

Best regards,

Kristian

rsignell-usgs:
@ethanrd and @lesserwhirls this is a tiny fix, can you guys merge it?

lesserwhirls:
I think we can do this. Not sure how this fell through the cracks, though. Would it be possible to rebase this on master?

rsignell-usgs:
@joanpau, would one of you SOCIB guys be willing to so this?

dopplershift:
Wow, really *not good* that this PR was closed without comment.

rsignell-usgs:
After getting this one merged, I'll try to help get some of the other SOCIB ncWMS improvements submitted.    They've got a bunch of additional styles that are cool, and those new WMS styles allow them to do nifty stuff like this:
http://apps.socib.es/Leaflet.TimeDimension/examples/example6.html

and more cool client examples using TDS WMS services here:
http://apps.socib.es/Leaflet.TimeDimension/examples/

lesserwhirls:
Just remember that all of the ncWMS stuff will be different in TDS v5.

rsignell-usgs:
Who is the point person on ncWMS now that Marcos is gone?

lesserwhirls:
I suppose that'd be me at this point.

joanpau:
We just rebased the fix onto the master branch.
Please note that, as stated in the commit message and in the description of the pull request,
it would be better to move the modified function to the `WmsUtils` package in ncWMS.
I do not know, however, if it is worth to do it now if release 5.0.0 is moving to ncWMS2.

JohnLCaron:
Not sure why I closed this 2 years ago, but I agree it was a mistake.

Issue #90, opened 2014-12-31
The HTTPSSLProvider is not working


cofinoa:
In the 4.3.22 version the HTTPSSLProvider is working but in 4.5.3 it has stopped to work. 
The command line flag for the JVM
-Dkeystore
-Dkeystorepassword
-Dtruststore
-Dtruststorepassword
are working and the HTTPSSLProvider is registered in the HTTPSession class but the SSL context used is not loading the truststore and the keystore provided. 

If -Djavax.net.debug=all is enabled in the JVM you can see that the truststore loaded is the system default one.  This can be overloaded with the JSSE flags
-Djavax.net.ssl.trustStore
-Djavax.net.ssl.trustStorePassword
-Djavax.net.ssl.keyStore
-Djavax.net.ssl.keyStorePassword
 
and the provided truststore is loaded instead the default one

But the keyStore param it's not working.

PR #91, opened 2015-01-02
workaround for the issue #90 ( HTTPSSLProvider is not working )

cofinoa:
This is a workaround for the issue #90


cwardgar:
Try creating a pull request against Unidata:4.5.4, not Unidata:4.5.3.

DennisHeimbigner:
This is probably a result of our moving to Apache httpclient library 4.2.
I have not had the capability of testing the client side keys issue.

cofinoa:
I made this pull-request against the wrong branch and therefore @cwardgar has closed it.
I have create a new pull request against the 4.5.4 branch. The pull-request is the #92

DennisHeimbigner:
It turns out that equivalent code is already in 4.5.5 and 4.6.0.
You can see it at HTTPSession#setGlobalKeystore().
However, the invocation of this code in HTTPAuthStore has a bug
that prevented it from being invoked.  I have fixed 4.5.5. to invoke the
code properly. However, I still do not know if it works because
of no access to a server to test it.  Please pull 4.5.5 from github
and see if it solves your problem.

PR #92, opened 2015-01-02
workaround for the issue #90 ( HTTPSSLProvider is not working )

cofinoa:
* Workaround for the issue #90. My proposal is that if the HTTPAuthPolicy.PROVIDER parameter is not defined the fallback is try to use the JVM -D properties if they are defined. 
 * This pull is not fixing the issue #90.
 * Also the stdauthentication() method is been used instead returning a null value.
 * Some spelling errors has been fixed: TSL->TLS


cofinoa:
Dear @JohnLCaron and @DennisHeimbigner , 

it would be possible to merge this pull-request into 4.5.4 and 4.6?

For us, the issue #90 which is solved by this pull-request is blocking us to upgrade our client tools to 4.5 version. 

Regards

Antonio

JohnLCaron:
Antonio, can you see if 4.5.5 fixes the problem? thanks!

DennisHeimbigner:
Back port the fix to 4.5.5

JohnLCaron:
we just released 4.5.5 yesterday, so any new work should be on 4.5.6

JohnLCaron:
i think this all works now, so im closing

PR #96, opened 2015-02-04
Implement readTimeseries at ThreddsScalarLayer to improve efficiency

bielfrontera:
We have implemented the method readTimeseries at ThreddsScalarLayer in order to improve efficiency when accessing data values for a single point.

The problem was that the generic method defined at AbstractScalarLayer (in ncwms) was calling readHorizontalPoints for each time included in the timeseries requested and there are some variables that could be calculated only once. With this new method, we avoid repeating this calculations more than one time (CdmUtils.createHorizontalGrid, for example).

Tested in our thredds server, with a significant improvement in the response time of GetFeatureInfo requests (with a time interval as a parameter).

### DIFF
--- a/tds/src/main/java/thredds/server/wms/ThreddsScalarLayer.java
+++ b/tds/src/main/java/thredds/server/wms/ThreddsScalarLayer.java
@@ -179,6 +179,27 @@ class ThreddsScalarLayer extends AbstractScalarLayer implements ThreddsLayer {
 
   }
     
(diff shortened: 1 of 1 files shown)


dopplershift:
This got closed when I deleted the `4.5.4` branch.
Cherry-picked into `master` for the 4.6.1 release.

Thanks for the contribution, sorry it languished so long.

PR #99, opened 2015-02-13
MRMS Support

dopplershift:
Several changes to GRIB for MRMS:
- Support for getting missing and fill values from the Grib table--MRMS specifies a variety of values for "no coverage" vs. "missing"
- Support for the Template 5.41 (PNG) format
- MRMS variable tables


dopplershift:
@JohnLCaron I'd love your signoff on this since I've modified `GribParameter`.

dopplershift:
I'm also curious about adding tests for this. There don't seem to be any tests in the grib package itself. I'd like to test an MRMS file, for the PNG formatting, the variable tables, and the missing values. I would also like to add a test for jira ticket TDS-529, which has a GRIB file with PNG and a missing values bitmap.

For MRMS, I'm also curious, there is one field PrecipFlag that maps discrete values to various meanings: ftp://ftp.nssl.noaa.gov/projects/MRMS/GRIB2_TABLES/UserTable_MRMS_PrecipFlags.csv  Is there anyway to support this translation here?

ethanrd:
Hey Ryan,

CF has a section on flags [1] for just this sort of thing.

Ethan

[1]
http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch03s05.html

dopplershift:
Well that's handy. Now, how to inject that into our GRIB processing. I could add it to the MRMS custom table, but that requires another addition to the GribParameter class...

ethanrd:
Are there no parameters that use flags in the standard tables? I guess
probably aren't too common in model data. I even checked a few NDFD files
and didn't see any thing. I was wondering about cloud cover but it is
stored as percent.

Sounds like an interesting addition.

Ethan

JohnLCaron:
Youve probably figured this out, but most grib tests are in cdm_test module, mostly because they read grib files from cdmUnitTest.

only put tests in GRIB is they are true unit tests or read data thats in the repo.

JohnLCaron:
> For MRMS, I'm also curious, there is one field PrecipFlag that maps discrete values to various meanings: ftp://ftp.nssl.noaa.gov/projects/MRMS/GRIB2_TABLES/UserTable_MRMS_PrecipFlags.csv Is there anyway to support this translation here?

where does this PrecipFlag get encoded in the GRIB record?

JohnLCaron:
Its wrong to add fillvalue, missing value to parameter tables. is there no place in the DRS or Data section where these values are encoded directly into the record? Im not seeing it, it probably should have been in DRS spec 5.41. so once again we have non-standard GRIB hell.

your solution is good enough for now i guess.

the other two pull requests look fine.

JohnLCaron:
you think "no Coverage" means "Fill value" ??

dopplershift:
Some responses @JohnLCaron :
- The sample files I have are small enough, so I'll add tests and data to the grib package (and repo), since that makes the most sense to me.
- The PrecipFlag is a full-fledged variable that depicts precipitation type (convective, stratiform, snow, etc.). The field encodes these types as integers. Does it make sense to include this metadata in our NetCDF representation of this GRIB file? How? Another extension to the internal GRIB table?
- Unless you can find the missing *and* no coverage values in the files I sent you, I don't see any other good way to handle masking out those variables that isn't *more* intrusive
- "No Coverage" means that there is no instrument (radar, etc.) that can provide data in that region--it's outside the area we expect to cover. "Missing" means that area is covered, but the data are removed (noise, QC, etc.). Since the two attributes were available, it seemed good to use them both, but I'm completely open to just supplying one of `missing_value`/`_FillValue` with a list of values if that's going to make more sense to the CF clients of the world.

JohnLCaron:
On Tue, Feb 17, 2015 at 6:52 PM, Ryan May <notifications@github.com> wrote:

> Some responses @JohnLCaron <https://github.com/JohnLCaron> :
>
>    - The sample files I have are small enough, so I'll add tests and data
>    to the grib package (and repo), since that makes the most sense to me.
>
> ok

>
>    - The PrecipFlag is a full-fledged variable that depicts precipitation
>    type (convective, stratiform, snow, etc.). The field encodes these types as
>    integers. Does it make sense to include this metadata in our NetCDF
>    representation of this GRIB file? How? Another extension to the internal
>    GRIB table?
>
> ok i understand. for now, we just expose these as integers.
what youve got really is an enum type. I used these a lot in BUFR, havent
used before in GRIB. we can experiment with those when im back.


>    - Unless you can find the missing *and* no coverage values in the
>    files I sent you, I don't see any other good way to handle masking out
>    those variables that isn't *more* intrusive
>
> Well, one could just remove them in the data. thats easy to do when those
values are encoded in the DRS or data section. When in an external table
(!) your approach seems fine for the moment. probably review when im back.


>
>    - "No Coverage" means that there is no instrument (radar, etc.) that
>    can provide data in that region--it's outside the area we expect to cover.
>    "Missing" means that area is covered, but the data are removed (noise, QC,
>    etc.). Since the two attributes were available, it seemed good to use them
>    both, but I'm completely open to just supplying one of missing_value/
>    _FillValue with a list of values if that's going to make more sense to
>    the CF clients of the world.
>
> yes, ok for now. if we decided to remove in the data, we probably wouldnt
keep the distinction.
I dont really like using _FillValue in this way. Would prefer some explicit
missing_value variant attributes. But I dont think we have them, so
_FillValue is ok.

dopplershift:
Went ahead and merged to 4.6.0 since it seems like all we need are, possibly, a few tweaks.

PR #106, opened 2015-03-17
NetcdfSubset didn't fire basic challenge when dataset is restricted

### LINKED ISSUE #103: The NetCdf Subset Service has 404 problems when restricted datests are accessed
Dear all, 

I'm using TDS version 4.3.22 and I have configured NCSS servlet to serve datasets. The servlet is working, but when I'm tryn to access restricte datasets the servlet reports a 404 HTTP error. 

I think it's a issue with the restricted acces configuration, because if I access to same dataset using the OpenDAP service the TDS challenges to me with the Basic authentication and after login I can access (partially) to the restricted dataset using the ncss. I said partially because only the web form is shown. The css and images files are not loaded and the requests are not working. 

Regards

Antonio

vegasm:
fd is null when a Feature type is not supported and when a dataset is restricted if the user is not logged in. Maybe the final solution is to split the behaviour in the Dataset service  throwing like an Unauthenticated exception that can be catched by the controllers. This solution mentioned requires to modify all the controllers which use FeatureDatasetService in order to catch it.

Fixes #103

### DIFF
--- a/tds/src/main/java/thredds/server/ncss/controller/NcssDatasetInfoController.java
+++ b/tds/src/main/java/thredds/server/ncss/controller/NcssDatasetInfoController.java
@@ -89,7 +89,7 @@ class NcssDatasetInfoController extends AbstractNcssController {
       fd = datasetService.findDatasetByPath(req, res, datasetPath);
 
       if (fd == null)
-        throw new UnsupportedOperationException("Feature Type not supported");
+    	  return;
 
       String strResponse = ncssShowDatasetInfo.showForm(fd, buildDatasetUrl(datasetPath), wantXML, showPointForm);
 


JohnLCaron:
I was just working on this yesterday in 4.6.0. Have you tried that branch?

vegasm:
I haven't tried in 4.6.0. I've tried in 4.5.4 and 4.5.6 because are the versions deployed in the production server. Is that issue fixed in 4.6.0? I can test in that version in any case but it would be great to have a fix in 4.5.6.

JohnLCaron:
4.5 is not a stable release, probably wont backport. 4.6.1 will be stable release soon.

PR #107, opened 2015-03-18
Upgrades to HTML view of catalogs

lesserwhirls:
Added CSS/JavaScript to collapse datasets with level > 0.

Make sure that we fallback to old behavior if javascript is not
  enabled on the client.

Would like to test this on ATM to see how well it does.


dopplershift:
Looks nice and simple. Seems more straightforward to use https://code.jquery.com/jquery-1.11.2.min.js as the source for the jquery though.

lesserwhirls:
Thanks for the feedback Ryan - I've updated the pull request to use the query site.

PR #108, opened 2015-03-18
Support of Uniplot files and different charsets

JSchnabel:
The data analysis tool Uniplot creates files which are oriented on the netcdf3 format but show some differences:
- "CDH" magic, little endian byte order and ISO-8859-1 encoding
Additionally we have customer files of the hdf5 format with text which isn't UTF-8 encoded.
To be able to read such files we decided to extend the code as follows:
- N3header now knows the "CDH" magic
- text encoding may be now configured for netcdf3 and hdf5
When creating a NetcdfFile instance one can now provide a Map with ImportProperties. Currently the only key defined is AbstractIOServiceProvider.PROP_VALUE_CHARSET. N3Header and H5Header try to get the actual file encoding from this map.

### DIFF
--- a/cdm/src/main/java/thredds/catalog/DataFormatType.java
+++ b/cdm/src/main/java/thredds/catalog/DataFormatType.java
@@ -54,6 +54,7 @@ public final class DataFormatType {
     public final static DataFormatType HDF5 = new DataFormatType("HDF5");
     public final static DataFormatType NETCDF = new DataFormatType("netCDF");
     public final static DataFormatType NETCDF4 = new DataFormatType("netCDF-4");
+    public final static DataFormatType NETCDH = new DataFormatType("netCDH");
     public final static DataFormatType NEXRAD2 = new DataFormatType("NEXRAD-2");
     public final static DataFormatType NCML = new DataFormatType("NcML");
     public final static DataFormatType NIDS = new DataFormatType("NEXRAD-3");
@@ -77,7 +78,7 @@ public final class DataFormatType {
       this.name = s;
       members.add(this);
     }
-  
+
     private DataFormatType(String s, boolean fake)
     {
       this.name = s;
@@ -126,10 +127,10 @@ public final class DataFormatType {
     public String toString() { return name; }
 
      /** Override Object.hashCode() to be consistent with this equals. */
-     public int hashCode() { return name.hashCode(); }
+    public int hashCode() { return name.hashCode(); }
 
      /** DataFormatType with same name are equal. */
-     public boolean equals(Object o) {
+    public boolean equals(Object o) {
        if (this == o) return true;
        if (!(o instanceof DataFormatType)) return false;
        return o.hashCode() == this.hashCode();
--- a/cdm/src/main/java/thredds/client/catalog/writer/DataFactory.java
+++ b/cdm/src/main/java/thredds/client/catalog/writer/DataFactory.java
@@ -518,7 +518,8 @@ public class DataFactory {
         // these are the file types we can read
         if ((DataFormatType.BUFR == format) || (DataFormatType.GINI == format) || (DataFormatType.GRIB1 == format)
                 || (DataFormatType.GRIB2 == format) || (DataFormatType.HDF5 == format) || (DataFormatType.NCML == format)
-                || (DataFormatType.NETCDF == format) || (DataFormatType.NEXRAD2 == format) || (DataFormatType.NIDS == format)) {
+                || (DataFormatType.NETCDF == format) || (DataFormatType.NETCDH == format) || (DataFormatType.NEXRAD2 == format) 
+                || (DataFormatType.NIDS == format)) {
           access = tryAccess;
         }
       }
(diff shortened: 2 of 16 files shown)


JohnLCaron:
Hi JSchnabel:

In general, I would prefer a different IOSP for a different file format, to
keep things clear as to what is whose. Im open to the possibilty of
subclassing.

A question I have is how do you know what the text encoding is? Is it in
the file itself? Could you send an example where it makes a difference?

thnks,

John

JSchnabel:
Hi John,

 

A subclass of For the Uniplot CDH format (netcdf3 with "CDH" magic, little endian byte order and ISO 88591-1 encoding) subclassing N3header and maybe N3iosp can be a fine solution. Because here I can recognize the type by parsing the Magic. 

But unfortunately we found no way to automatically detect the encoding of the hdf5 files. So currently the customer has to select the charset in our software. We I think it would be a good way to introduce a customizable encoding at least for n3 and hdf5 even if it is expected to be UTF-8, because there are tools creating files without considering this and just use the local charset to write text.

 

We choosed a solution which changed as little as possible, but of course you have the big picture in mind. So just see it as a proposal. I will try to send you some example files during the next one or two days.

 

Yours sincerely,

 

Jan Schnabel

Senior Software Engineer

  _____  

AMS Gesellschaft für angewandte Mess- und Systemtechnik mbH

Bahnhofstr. 6

D-09111 Chemnitz

 

phone:       +49 (371) 918668-15

mobile:      +49 (176) 55020769

fax:                +49 (371) 918668-99

Mail: jan.schnabel@AMSonline.de

Web: www.AMSonline.de <http://www.amsonline.de/> 

 

CEO: Dr. Ing. Bernhard Sünder

HRB 9528 Amtsgericht Chemnitz

UStIdentNummer: DE 811 512 920

 

AMS Logo aktuell groß BASEG

JohnLCaron:
> But unfortunately we found no way to automatically detect the encoding of
the hdf5 files. So currently the customer has to select the charset in our
software. We I think it would be a good way to introduce a customizable
encoding at least for n3 and hdf5 even if it is expected to be UTF-8,
because there are tools creating files without considering this and just
use the local charset to write text.

Are we talking about the variable/dimension attribute names? or data values
or attribute values?

cwardgar:
This seems like a rather invasive change just to add support for another file format. Have you considered implementing an IOServiceProvider and then passing the charset as the iospMessage argument? [1]

[1] https://github.com/Unidata/thredds/blob/4.6.0/cdm/src/main/java/ucar/nc2/NetcdfFile.java#L423

JSchnabel:
Hi Christian,

 

Usage of the iospMessage sounds good. I’ll try creating an IOSP for the files created by uniplot (with CDH magic). To read n3 and h5 files with the wrong charset I need to enhance the existing IOSPs, because I see no possibility to select the IOSP to use for a given file.

 

Thank you, Jan

JSchnabel:
Hi John,

 

currently I can give you 2 small files, created by Uniplot. “HDF testfile.nc” is a HDF file containing one attribute with umlauts.  “CDF test_Tabelle1.xlsx.nc” is a “standard” NetCDF file containing attributes and units with umlauts. “test.xlsx” is the excel file the “CDF test_Tabelle1.xlsx.nc” file is created from.  For other files I need the customers agreement. That may take a while.

 

I’ll try to create an IOSP for Uniplot files (with CDH magic) derived from N3 implementation. Christian W. suggested to use the IospMessage instead of my new Map argument in NetCDFFile.open() to pass the manual selected charset to the IOSP. What do you think? Must I expect usage-conflicts at the IospMessage  argument?

 

Kind regards, Jan Schnabel

cwardgar:
Hi Jan,

You can specify the IOSP to use with [one of the NetcdfFile open methods](https://github.com/Unidata/thredds/blob/4.6.0/cdm/src/main/java/ucar/nc2/NetcdfFile.java#L487). However, it'd be even better if you [registered your IOSP](https://github.com/Unidata/thredds/blob/4.6.0/cdm/src/main/java/ucar/nc2/NetcdfFile.java#L115) and then used [IOServiceProvider.isValidFile](http://www.unidata.ucar.edu/software/thredds/current/netcdf-java/javadocAll/ucar/nc2/iosp/IOServiceProvider.html#isValidFile-ucar.unidata.io.RandomAccessFile-) to determine if it's Uniplot or not. The presence of the "CDH" magic should be enough information to decide that. You might also consider subclassing N3header as part of your implementation, since Uniplot files and regular NetCDF-3 files are substantially similar.

Cheers,
Christian

JSchnabel:
Hi Christian,

 

for Uniplot NetCDH files I can actually simply create a new IOSP. For the other files (N3 with CDF or H5 with HDF magic) we need the customizable charset because the only difference to standard seems to be the usage of another value charset than UTF8. So I’ll create a new IOSP for Uniplot N3 by subclassing N3iosp and N3header, remove the import properties map and enhance the sendIospMessage() of N3Header and H5Header to accept a Charset argument.

 

Thank you, Jan

dopplershift:
This was closed because we removed the 4.6.x branches and instead use `master`. If you resubmit the PR, please do so against `master`.

JSchnabel:
Thank you, I will …

PR #114, opened 2015-03-27
Update UnidataDD2MI.xsl to the threddsncISO 2.2.5 version

### LINKED ISSUE #113: Update the UnidataDD2MI.xsl
The version of TDS 4.6 released on Mar 26, 2014 still has this old UnidataDD2MI.xml file from 2012:
https://github.com/Unidata/thredds/blob/4.6.0/tds/src/main/webapp/WEB-INF/classes/resources/xsl/nciso/UnidataDD2MI.xsl

This needs to be replaced with this one:
https://github.com/Unidata/threddsIso/blob/master/src/main/resources/xsl/nciso/UnidataDD2MI.xsl

rsignell-usgs:
Fixes #113


cwardgar:
Merged. The long-term solution to this problem is just have THREDDS use the file in threddsIso, instead of having duplicate (and often out-of-sync) versions in both.

ethanrd:
There's lots of stuff that THREDDS uses from threddsIso. So I think the
right solution is to figure out / clean up our chain of builds needed to
build the TDS. Besides threddsIso, that would also include ncWMS.

PR #119, opened 2015-04-08
improvements to VariablePlot

petejan:
now plots when first dimension is not time, and when variable has no units


JohnLCaron:
thanks!

PR #130, opened 2015-04-30
HTTP sessions fix  #129

### LINKED ISSUE #129: Massive session creation in Threads
Hi,

I would like to expose a problem we have getting datasets data from our Thredds instance using netcdf-all. 

We have tools developed in R to execute netcdf-all and to download data with our tool we define the following:

HTTPBasicProvider bp = new HTTPBasicProvider("user", "password");
HTTPSession.setGlobalCredentialsProvider(bp);

Then, we open the dataset:

GridDataset gd = GridDataset.open("URL_TO_DATASET");

and after doing that, we execute GeoGrid.makeSubset and GridDataType.readDataSlice many times which create new requests to Thredds.

The problem is the way requests are developed forces Thredds to create a new session and re-authenticate in each request because the cookie store reference is lost. This behavior floods our Tomcat creating up to 200 sessions per user at the same time to download the data. I checked the code deeply in order to find what was happening:

For example, each time DConnect2.openConnection is called, a new session is generated and the cookies from the last session are lost because you get cookies from SessionClient which is redefined in each new session.

As you know, there are so much code deprecated in the classes I mentioned and in my opinion the client must be replaced by the new CloseableHttpClient and created through its builder. Another thing is the amount of code these clases have makes them difficult to maintain.

vegasm:
#129


DennisHeimbigner:
Manuel, there is a separate branch in github (http43) in which
I am moving httpservices to use Apache httpclient 4.4. This
involves a number of changes (especially handling immutability).
I will work with your changes in the context of that branch.
=Dennis Heimbigner
  Unidata
p.s. for future reference, it is best to not combine unrelated change sets
    in the same pull request. This request could have been submitted
    as two: one for httpservices and one for coordinate axis.

DennisHeimbigner:
I have inserted a variant of you pull request into a branch off of
4.6.2. I did no create an e.g. HTTPSessionCloseable, but rather
made HTTPSession (and several other classes AutoCloseable).
I also modified a number of classes to repeated open/close of HTTPSession.
I have placed what I hope are the relevant jars here:
ftp://ftp.unidata.ucar.edu/pub/netcdf-java/
Can you see if they solve your problems, and report on this
comment thread?

vegasm:
Which is the branch where you are performing the changes? http44 or http43? I would like to test it in my eclipse. Thanks

DennisHeimbigner:
It is the "close" branch. 
=Dennis Heimbigner
 Unidata

Manuel wrote:

DennisHeimbigner:
Have yoou had a chance to try this branch?

vegasm:
I expect to check it tomorrow ;)

vegasm:
Hi Dennis, 

it looks great because it works as expected (it does not create a session per url request) but It creates two sessions and I need to do more testing to figure out if things are going ok. I'll let you know as soon as possible to integrate officially these changes.

Best regards

vegasm:
Hi Dennis, I fully tested your code and maybe we can discuss some aproaches.

As you know, if I want to download data from a dataset using the Opendap service there are two connections:

1)NetcdfDataset#CheckIfDods executes directly a URL using the HTTPFactory:
      HTTPMethod method = HTTPFactory.Get(location + ".dds")) 
      int status = method.execute();
2)DODSNetcdfFile#DODSNetcdfFile(String datasetURL, CancelTask cancelTask) generates in its constructor a DConnect2 instance and then are invoked Dconnect2#getDDS() and Dconnect2#getDAS();

The second one is invoked in the downloading process many times because we get data like this:

vegasm:
Hi Dennis, I fully tested your code and maybe we can discuss some aproaches.

As you know, if I want to download data from a dataset using the Opendap service there are two connections:

1)NetcdfDataset#CheckIfDods executes directly a URL using the HTTPFactory:
      HTTPMethod method = HTTPFactory.Get(location + ".dds")) 
      int status = method.execute();
2)DODSNetcdfFile#DODSNetcdfFile(String datasetURL, CancelTask cancelTask) generates in its constructor a DConnect2 instance and then are invoked Dconnect2#getDDS() and Dconnect2#getDAS();

The second one is invoked in the downloading process many times because we get data like this:
/dodsC/system4/System4_Seasonal_15Members.ncml.dods?tp[238:1:238][0:1:0][28:1:118][64:1:73][0:1:7]

That was causing problems due to the implementation before was considering to create a new session per request. That was solved because now you don't create a new session each time (_session = HTTPFactory.newSession(urlString);) and the closeable implementation is correct. But it persists that a new session is created to check if it's Opendap and another to download the data itself. 

I implemented a class which is in the pull request (HTTPCloseableSession) which allows users to access data without URL restrictions generating only one session. With your current implementation If I get another dataset (different URL) two new sessions are created. And If you check it, HTTPCloseableSession includes the cookie store so it is possible to add a cookie from a local store which allows more possibilities (recycling like browsers)

I think it'd be nicer to use only one session (No URL dependencies) like my implementation does. Could it be possible to include that approach?

best regards

DennisHeimbigner:
It appears you would like to make HTTPSession be a singleton
shared by all method calls. I do not think this is a good idea
because it introduce potential security holes.

Semantically, HTTPSession is intended to provide the shared state for
a group of related method calls. Specifically, each HTTPSession is associated
with a specific server plus, potentially, a single "principal" (user).
This is why, for example, it is possible to set both a global credentials
provider as well as a per-instance credentials provider (with the latter taking
precedence). This capability is lost when a singleton HTTPSession is used.

One more thing. With respect to NetcdfDataset#checkIfDods, it is static,
as is its callers all the way up to NetcdfDataset#acquireDataset.
This means that at the point of call, there is no HTTPSession object,
hence it must be created, used, and destroyed. Trying to be smart and
keeping it around for later use in NetcdfDataset seems more trouble than its
worth since creating an instance of HTTPSession is relatively cheap.

If you have other suggestions for optimization, then please let me know.

=Dennis Heimbigner
  Unidata

Manuel wrote:

DennisHeimbigner:
This is essentially fixed and tested. So close it.
b

vegasm:
Ok. Will it be available in the next release?

dopplershift:
These changes will appear in 4.6.3--no ETA on when that will be released.

DennisHeimbigner:
I merged the code into master last night, so it is available
in the current github master.
=Dennis Heimbigner

Ryan May wrote:

dopplershift:
Right. That wasn't the question, though.

PR #131, opened 2015-04-30
Fix problem reading huge attributes.

msdsoftware:
netCDF-Java was unable to open HDF5 datasets with huge attributes because of exception thrown due to bad BTree address.


JohnLCaron:
looks good - i will run it through our regression tests tonite to see if it breaks anything unexpected.

PR #132, opened 2015-04-30
Exclude 3rd-party code-signing files from ToolsUI

msdsoftware:


dopplershift:
@cwardgar Any thoughts?

msdsoftware:
Without this patch, executing ToolsUI from command line (i.e., java -jar toolsUI-4.6.1-SNAPSHOT.jar) will immediately throw "Exception in thread "main" java.lang.SecurityException: Invalid signature file digest for Manifest main attributes".

cwardgar:
Starting in 4.6.1 — which we hope to release next week — we will be building THREDDS using Gradle instead of Maven. The Gradle build already excludes those files when creating ToolsUI.

Issue #141, opened 2015-06-22
ucar.nc2.util.DiskCache2.canWrite returns true for read-only UNC directories


cofinoa:
The ucar.nc2.util.DiskCache2.canWrite method returns true value when shared windows resources (UNC) paths are read-only.

cofinoa:
This an well known bug for nio implementation:
https://bugs.openjdk.java.net/browse/JDK-8052010

no solution exists

cwardgar:
Hmm, so is there anything to do for this one? I doubt that the old java.io.File methods are any better.

cofinoa:
To avoid issues with API guessing if a path is writable, I think is better to test if a tem file can be created and deleted. 
Please, take a look to the pull-request that I have made.

cwardgar:
Based on you link, it seems like this issue was fixed in 8u12 or earlier (I can't reproduce it in 8u45). Nonetheless, it's worthwhile for people who are running NetCDF-Java on earlier versions.

I built on your example to check for another couple cases: https://github.com/Unidata/thredds/blob/92a026d03ec3bbfc44dd379874b3380bc0409137/cdm/src/main/java/ucar/nc2/util/DiskCache2.java#L261

Thanks for the contribution!

PR #142, opened 2015-06-23
Create a temp file to test writable

cofinoa:
Relates to issue #141 

Because Java NIO has some issues with windows UNC read-only shares, my
proposal it's to create a real temp file on File parent path.


cwardgar:
This PR was the basis for the solution to issue #141.

PR #174, opened 2015-09-09
Fix travis libs broken by `b3e724`.

dopplershift:
Not sure what happened to these files to get them into such a weird state, but I'm pretty sure they were modified in `b3e724` by mistake.


lesserwhirls:
Thanks Ryan!

PR #175, opened 2015-09-09
5.0.0

JohnLCaron:


cwardgar:
Why roll your own Optional? There's one in both JDK8 and Guava.

JohnLCaron:
I needed an Optional which allowed me to put an "error message" into it when it was null.

java8 Optional if final. Didnt know about guava's. Its possible i can subclass it.

PR #180, opened 2015-09-10
Fix failing tests

dopplershift:
The overall purpose here is to get Travis green again. This does a couple things now:
- Remove an assertion caused by sending Long.MAX to joda, which is no longer valid. It also refactors the test to be an actual unit test. There are some commented out tests that do not pass (which printed
errors before).

- Because its down currently, we're now aware that we depend on getting to http://echo.httpkit.com This adds it to the collection of servers usable for `Assume`. It also renames the `ThreddsServer` to `ExternalServer` since we now include something non-THREDDS.

I've opened a separate issue on things that still need work on `TestCalendarDateUnit`, but at least Travis should be clean now.


dopplershift:
YES!

lesserwhirls:
Green is good - death to red!

Issue #182, opened 2015-09-11
5.0 Build Errors and Warnings


lesserwhirls:
@JohnLCaron - I dug into what was going on with Coverity. There was a compiler upgrade, and I went ahead and upgraded it as well as upgraded the latest JDK 8, but here is what I am running into when I try to build:

```
cdm/src/main/java/ucar/nc2/util/EscapeStrings.java:21: error: unmappable character for encoding UTF-8
  "Starting from version 15.0, Google�s excellent Guava libraries have UrlEscapers class to do just that!
                                     ^
cdm/src/main/java/ucar/nc2/util/EscapeStrings.java:25: error: unmappable character for encoding UTF-8
    3) urlPathSegmentEscaper for the names and values of request parameters (things after the �?�).
                                                                                              ^
cdm/src/main/java/ucar/nc2/util/EscapeStrings.java:25: error: unmappable character for encoding UTF-8
    3) urlPathSegmentEscaper for the names and values of request parameters (things after the �?�).
                                                                                                ^
cdm/src/main/java/ucar/nc2/ft/point/remote/RemotePointCollection.java:36: error: package com.sun.istack.internal does not exist
import com.sun.istack.internal.NotNull;
                              ^
cdm/src/main/java/ucar/nc2/ft/point/remote/RemotePointCollection.java:121: error: cannot find symbol
  @NotNull
   ^
  symbol:   class NotNull
  location: class RemotePointCollection
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
2 errors
:cdm:compileJava FAILED

FAILURE: Build failed with an exception.
```

A few non-UTF-8 characters, and a missing dependency for NotNull?

lesserwhirls:
See PR Unidata/thredds#185 for a fix.

lesserwhirls:
Just ran and submitted a coverity build, and the console output looks good (that is, I think the scan will be useful this time!).

PR #185, opened 2015-09-11
Fixes gh issue Unidata/thredds#182

lesserwhirls:
removed non UTF-8 chars
changed RemotePointCollection to use javax NotNull
added "validation-api" compile time dependency to cdm


lesserwhirls:
Once this is merged, I'll re-run coverity so that we can get something useful out of it.

PR #194, opened 2015-09-15
Unicode NetCDF Object Names - non-English letter as first character

randallwhitman:
[Documentation on identifier/object names](http://www.unidata.ucar.edu/software/thredds/current/netcdf-java/CDM/Identifiers.html) states:
"The first character of a name must be alphanumeric, a multi-byte UTF-8 character, or `_` (reserved for special names with meaning to implementations, such as the `_FillValue` attribute)."
By regular expression
  `([a-zA-Z0-9_]|{MUTF8})([^\x00-\x1F\x2F/\x7F]|{MUTF8})*`

However, the regular expression used in `N3iosp.isValidNetcdfObjectName` (called from `NetcdfFileWriter.isValidObjectName`) is more strict than what is documented - first character is *English*-letter/decimal-digit/underbar.
  `[a-zA-Z0-9_][^\\x00-\\x1F\\x2F\\x7F]*`

Proposed regex - changes only the character class for the first character:
  `[\\p{L}0-9_][^\\x00-\\x1F\\x2F\\x7F]*`

### DIFF
--- a/cdm-test/src/test/java/ucar/nc2/TestWriteMisc.java
+++ b/cdm-test/src/test/java/ucar/nc2/TestWriteMisc.java
@@ -38,12 +38,12 @@ public class TestWriteMisc {
     NetcdfFileWriter fileWriter = NetcdfFileWriter.createNew(NetcdfFileWriter.Version.netcdf3, TestLocal.temporaryDataDir + "objnames.nc");
     Dimension[] dim = new Dimension[1];
     int aSize = 8;
-    dim[0] = setDimension(fileWriter, "ñ", "blah", aSize);
+    dim[0] = setDimension(fileWriter, "ñÁÉ", "blah", aSize);
     Variable v = fileWriter.addVariable(null, "áéíóú", DataType.FLOAT, Arrays.asList(dim));
     fileWriter.addVariableAttribute(v, new Attribute("_FillValue", -9999));
     fileWriter.addVariableAttribute(v, new Attribute(CDM.MISSING_VALUE, -9999));
-    System.out.println("Did not throw exception");
-    fileWriter.create();  // throws error missing native C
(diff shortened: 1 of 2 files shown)


cwardgar:
The problem is worse than you say: we're not allowing multi-byte UTF-8 characters **at all**, even though they're permitted to appear anywhere in an identifier. There's a telling comment in N3iosp: https://github.com/Unidata/thredds/blob/v4.6.3/cdm/src/main/java/ucar/nc2/iosp/netcdf3/N3iosp.java#L137

It says "LOOK do we need to implement this ??", followed by a function from the NetCDF-C library that checks names. The comment is 6-and-a-half years old, so maybe we really *don't* need to worry about it.

@JohnLCaron, do you have any thoughts? The C implementation of the UTF-8 stuff seems like a massive pain in the ass, which I imagine is why you didn't want to bother with it if you could help it.

@randallwhitman, I'm not sure that `\\p{L}` is quite enough, as it only matches Unicode characters in the "letter" category. It would seem that any multi-byte UTF-8 character is permissible, though I'd have to dig through the C implementation to see exactly what it's doing. For our purposes, the C code can be considered the authoritative reference.

randallwhitman:
Indeed, that would allow any character other than the CTRL/DEL characters in ASCII range.  I'll defer on how much that is really needed.  I will claim that allowing all Unicode letters (rather than English-only) is a practical valuable step forward.

DennisHeimbigner:
Not sure what the issue is here. the incoming name is UTF-16 already
(not, admittedly utf-24, but probably close enough for our purposes).
So we can do our name verification on that and then convert to utf-8.
Am I missing something?

cwardgar:
I wanted to do the implementation differently, so I created a pull request of my own: #196

Thanks for spotting this issue, Randall!

PR #201, opened 2015-09-18
Update LambertConformal.java

msdsoftware:
I have a sample dataset of WRF temperature data for the contiguous US that uses Lambert conformal conic gridding with a projection origin of latitude = +90 degrees. It cannot be opened by the NJ library because the LambertConformal class throws an exception.

The test in LambertConformal's precalculate() method whether lat0 equals +/- PI_OVER_2 is incorrect, as this is a possibly valid choice. I have altered the test to check instead that abs(lat0) is not greater than PI_OVER_2. Doing so allows NJ to open the dataset, and subsequently there is no mathematical error in any of the projection calculations.

The tests of whether the two standard parallels are valid are correct in trying to test whether those two values equal +/- 90 degrees (and if so reject them) but they did not set whether those values were outside the range (-90,90).


dopplershift:
Thanks for the contribution. Any chance you could add a test in `cdm/src/test/ucar/unidata/geoloc/projection`? It would help greatly to ensure we don't break this in the future; whatever you did to prove to yourself that this is working properly is fine. Unless you verified by looking at plots of WRF, in which case checking a few numbers project properly (or even just validating that the error is no longer thrown).

msdsoftware:
@dopplershift, Compiling my local copy of the library with the suggested changes and linking it into the Panoply app, I have made a plot of the sample WRF data that specified lat0 = +90.

But if future breakage is a concern, I'll take a look at adding a test or two.

dopplershift:
Yeah, the greatest concern is making it so we can change our code in the future and be confident (via tests) we haven't broken other people; your case (which we hadn't ever encountered ourselves) is a perfect example.

msdsoftware:
Added a TestLambertConformal class. Deleted the TestLC class which was redundant (basically TestUtm by another name).

dopplershift:
Test failure unrelated (due to TestDConnect2 failing with remotetest down). I restarted it manually.

dopplershift:
This is great. Thanks for the tests.

Issue #163, opened 2015-08-28
Copy/paste in toolsUI on Mac OSX


lesserwhirls:
On MacOSX, the default key combination for copy/paste is CMD+C/CMD+V. In order to use copy/paste in toolsUI (and likely TdsMonitor), you must use Ctrl+C/Ctrl+V.

msdsoftware:
If you alter the createUI method in ToolsUI to _not_ install Nimbus L&F if it detects the user is on OS X, but instead install no UI at all, then you will get the default OS X L&F. In that case the editing keystrokes will be as expected by the Mac user.

Thus, to get something that looks slightly more Mac-like, that method might instead start

  private static void createGui() {
    if (System.getProperty ("os.name").contains ("Mac"))
    {
        System.setProperty ("apple.laf.useScreenMenuBar", "true");
    }
    else
    {
        try {
          // Switch to Nimbus Look and Feel, if it's available.
          for (UIManager.LookAndFeelInfo info : UIManager.getInstalledLookAndFeels()) {
            if ("Nimbus".equals(info.getName())) {
              UIManager.setLookAndFeel(info.getClassName());
              break;
            }
          }
        } catch (ClassNotFoundException | InstantiationException | IllegalAccessException |
                UnsupportedLookAndFeelException e) {
          log.warn("Found Nimbus Look and Feel, but couldn't install it.", e);
        }
    }

  // etc...

There are plenty of other things I would also alter in the makeMenu method to make the ToolsUI package look more Mac-like when running on OS X, but that's too much to get into here.

lesserwhirls:
Thank you @msdsoftware for the pointers!

We would really like to make toolsUI a "first class" utility. It was originally developed as a helper tool for internal development, and thus the lack of "polish", so to say, but we recognize that many people are using it for their own purposes. I'm personally stuck on whether or not we try to fix things as is in the Swing framework, or if we try to port the most useful parts to JavaFX. Honestly, both would be a lot of work.

I'll give this a stab later today and see what I can come up with. Also, we are very open to suggestions, so please feel free to create an issue with your toolsUI suggestions. We are testing the waters for using github issues as our unified issue/bug/feature request/etc. tracking system, as it's open to all.

Thanks again!

msdsoftware:
[...]

PR #202, opened 2015-09-18
Added some mac related fixes to toolsUI

lesserwhirls:
If on mac osx, use system default look and feel. This enables mac key bindings for copy paste. Fixes Unidata/thredds#163

If on a mac, make sure windows are closed when using quit from the system menu. This will trigger the exit method so that the preferences are written out just like if toolsUI was quit using the window 'x' button. [TDS-671]


lesserwhirls:
Before merging this, can someone test if the gui layout and prefs are saved between sessions of toolsUI on Linux and Windows?

msdsoftware:
Works as desired on the Mac and Windows set-ups I tried. Only thing I didn't like was that the L&F looked odd on Windows until I used the menu item to change it.

BTW: With the shutdown hook connected to the Quit item in the OS X app menu, you could omit the redundant "Exit" menu item in ToolsUI's "System" menu if OS X is detected.

cwardgar:
I resized the ToolsUI window and moved it into the corner. Then I quit and relaunched. The window appeared with the same location and size that it had prior to quitting. So, it looks like prefs are being saved and loaded okay.

@msdsoftware On Windows, toolsUI uses the Nimbus L&F, which looks like [this](http://www.unidata.ucar.edu/software/thredds/current/netcdf-java/reference/ToolsUI/ToolsUI.html). Is that what you got? In that case, that's the intended behavior.

msdsoftware:
I was using Win 7 via Parallels on OS X. The look of the window looked slightly Mac-like, esp. the scroll bars. Maybe that's what Nimbus is supposed to look like, but it didn't match the Win 7 system look.

What I normally do is check the os.name and if it's either OS X or Windows, to set the L&F to
UIManager.setLookAndFeel (UIManager.getSystemLookAndFeelClassName ( ));

Of course, on OS X you'll still get the system L&F without doing this.

PR #215, opened 2015-09-28
Support for RadarServer crawling S3 (via S3fs)

dopplershift:
So here are the changes I've made to allow the RadarServer to crawl/query a collection of radar data on S3 (provided the radarCollections.xml file is properly configured). The support for S3 is using the s3fs library (from the fork here: https://github.com/Upplication/Amazon-S3-FileSystem-NIO2).

As it turns out, by having the S3 support implemented through a java NIO2 FileSystem (and Path, etc.), there are no S3 specific changes in the code. The biggest change was to try to find the appropriate `FileSystem` to use based on the passed in location from the configuration file (other changes to make sure I use s3fs-supported parts of the NIO2 API). The last commit in the PR currently modifies our dependencies to include s3fs and adds the appropriate class to `META-INF/services`, but I'm not sure now this is the right way to go for a runtime dependency that's only needed if we want radarServer to hit S3. I also have the amazon credentials in an `amazon.properties` located in `WEB-INF` on my local repo; it's possible to pass these as environment variables.

So open questions:
- [x] Is there a way to drop in the jar (and add the services file) alongside our normal TDS install (i.e. can we eliminate the need to put s3fs dependency in our codebase) ?
- [x] What's the right way to set up the amazon credentials? Should we have them added to `setenv.sh`?

Once we decide these I'll be sure to create documentation (in asciidoc of course) to allow others to set this up.

### DIFF
--- a/docker/tds-base/.gitignore
+++ b/docker/tds-base/.gitignore
@@ -1,2 +1,3 @@
 *.jar
 *.war
+config.zip
--- a/docker/tds-base/Dockerfile
+++ b/docker/tds-base/Dockerfile
@@ -45,7 +45,7 @@ COPY tomcat-files/bash_logout /usr/local/tomcat/.bash_logout
 #
 # Copy thredds.war
 #
-COPY thredds.war /usr/local/tomcat/webapps/thredds##04.06.03-SNAPSHOT.war
+COPY thredds.war /usr/local/tomcat/webapps/thredds##04.06.04-SNAPSHOT.war
 
 #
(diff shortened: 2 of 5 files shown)


dopplershift:
Ok, looks like you can just drop the s3fs jar (and all its dependencies) into $CATALINA_HOME/lib and they will be found just found.

As far as environment variables go, you can set them just fine in setenv.sh. I'm also going to look into setting them in the environment of our docker container.

PR #219, opened 2015-09-30
force added setenv.sh

lesserwhirls:


lesserwhirls:
This commit changes nothing in the code but adds the setenv.sh file for tds-base docker build. Once the tests go green, I'll merge it in.

PR #222, opened 2015-10-05
skeleton for tds gitbook

lesserwhirls:
skeleton layout of tds gitbook
basic header level formatting changes on some adoc files


JohnLCaron:
would be good to get clear on what the headers / stylesheets look like, so
we can start to add them as we clean up the adocs.

lesserwhirls:
Hi John,

Doug it taking a look at the output to get a good style sheet for it. The
css may be injected using apache, like for the main unidata site. That
means we would not have to link or manage any css stuff for the docs, just
the content.

I'll let you all know when Doug has something for us to check out.

Cheers,

Sean

Issue #191, opened 2015-09-14
Cannot write Variables with ENUM DataTypes to NetCDF-4


cwardgar:
From [KPZ-208087](https://andy.unidata.ucar.edu/esupport/staff/index.php?_m=tickets&_a=viewticket&ticketid=26109&departmentid=15&ticketstatusid=3)

Failure ultimately happens
`at ucar.nc2.jni.netcdf.Nc4Iosp.createVariable(Nc4Iosp.java:2395)`
where we try to call the `nc_def_var` NetCDF-4 C library method. The [docs](https://www.unidata.ucar.edu/software/netcdf/docs/netcdf-c/nc_005fdef_005fvar.html) say:
```
xtype
One of the set of predefined netCDF external data types. The type of this parameter,
nc_type, is defined in the netCDF header file. The valid netCDF external data types
are NC_BYTE, NC_CHAR, NC_SHORT, NC_INT, NC_FLOAT, and NC_DOUBLE. If the
file is a NetCDF-4/HDF5 file, the additional types NC_UBYTE, NC_USHORT, NC_UINT,
NC_INT64, NC_UINT64, and NC_STRING may be used, as well as a user defined type ID. 
```
but we're trying to create a variable with type `NC_ENUM`. Looking through Nc4Iosp as well as its tests, it doesn't look like we have any code at all to handle the enum case.

cwardgar:
I added a failing test `ucar.nc2.jni.netcdf.TestNc4IospWriting.writeEnumType()` that demonstrates this issue: #192

DennisHeimbigner:
I made some changes that make it work upto the point where it invokes
nc_var_puts. So, the problem is probably in the netcdf-c code. So I
am investigating that.

DennisHeimbigner:
I did some additional testing where I duplicated
the testcase in c using the netcdf-c library directly.
It worked ok in that setting.
So, my current hypothesis is that you are using a
version of the netcdf-c library that is out-of-date.
At some point, nc_put_vars was changed to allow
it to work on user defined types (including enums).
Before that, it would fail. 
So, the thing I need to know is what version of the netcdf-c
library you are using. You can find this out as follows:
1. find the ncdump program associated with the library you are using.
2. Execute ncdump with no arguments. At the end of the usage output,
    it should say what library version is being used.

cwardgar:
@DennisHeimbigner I'm on 4.3.2.

[...]

PR #224, opened 2015-10-05
Fix for Github issue #191

DennisHeimbigner:
(https://github.com/Unidata/thredds/issues/191)
Note: this is a fix to master.
Several related fixes needed.
1. The EnumTypedef needed to store the typeid.
2. Needed to cause enum types to be created in the netcdf4 file.
3. The call to write the enum data needed to get the correct typeid.
4. Extra: (issue 223) removed reference to mbari.org as no longer reachable.


cwardgar:
Let's hold off on merging this until we resolve the issue I raised here: https://github.com/Unidata/thredds/issues/191#issuecomment-145695909

dopplershift:
I'm not clear...why was this closed?

DennisHeimbigner:
Closed because I am going to add the bug fix for the issue that Christian found.
I will resubmitt a different pull-request.

Issue #223, opened 2015-10-05
TestFiles depends on unmarked external server


dopplershift:
TestFiles relies upon accessing dods.mbari.org; however, this is not protected by an assume, which means it can cause a test failure.

DennisHeimbigner:
I have been rethinking this.

IMO, we should rethink the use of ExternalServer.java.
The reason is that most of the servers we reference
are used only for testing or are not under the control
of Unidata (like mbari.org).

So, the only way we have to detect server failure in a timely
fashion is through the failure of a testcase under Jenkins
or Travis. As it is, using ExternServer in effect hides the
information that a server has failed by silently not running
some tests.

Thredds.ucar.edu is one exception to this for the simple reason
that it is used by enough other users that we will hear about it
if it fails. This is not true for the servers used in testing only.

Bottom line: I am closing this issue without any modifications.

cwardgar:
Resolved by #236.

PR #228, opened 2015-10-07
* Fix for Github issues #191 and #223

DennisHeimbigner:
* Fix for Github issue #191
Several related fixes needed.
1. The EnumTypedef needed to store the typeid.
2. Needed to cause enum types to be created in the netcdf4 file.
3. The call to write the enum data needed to get the correct typeid.
4. Fix endian problem because getDataAsByteBuffer does not properly
   handle endian order.
   - This required some mods to ucar.ma2.ArrayXXX.java
5. Modify TestNc4IospWriting.java to capture and textually compare
   the internal CDM file with the externally written netcdf file.

* Fix for Github issue #223
  Remove reference to mbari.org since it is no longer accessible.


cwardgar:
Looks good.

cwardgar:
Actually does this address #223? I can't see a code change related to it.

PR #229, opened 2015-10-07
Fixed FlattenedDatasetPointCollection and dependent tests.

cwardgar:
This should get the :cdm tests back to green.


cwardgar:
Rebased on top of John's latest 5.0.0 changes, which are causing a few test failures.

Can someone merge this in? It's my week on support, so I'll review and merge everybody else's.

PR #231, opened 2015-10-09
Update BAMutil.java

msdsoftware:
Use META_MASK when on Mac OS so that any action accelerators get Mac-like keystroke.


dopplershift:
Awesome, thanks for the contribution!

PR #232, opened 2015-10-09
Fix collection update during the catalog reinitialization

kristiansebastian:
Re-initialize the catalog through the "Reinitialize" option (/thredds/admin/debug?catalogs/reinit) doesn't manage correctly the job and trigger scheduler, because the job and triggers are already sheduled.  Example of warn and error displayed in the fc.collection.log files:

```
2015-10-09T11:37:24.531 +0200 WARN  - scheduler failed to add updateJob for UpdateCollection.hf_radar_ibiza-scb_codarssproc001_L1_agg. Another Job exists with that identification.
2015-10-09T11:37:24.532 +0200 ERROR - scheduleJob failed to schedule cron Job
org.quartz.ObjectAlreadyExistsException: Unable to store Trigger with name: 'hf_radar_ibiza-scb_codarssproc001_L1_agg' and group: 'rescan', because one already exists with this identification.
```

In order to fix this issue a new method has been implemented to reinitialize the scheduler when the DataRootHandler.reinit() is called. We haven't modified the TDM because we don't use it, and it is likely that some modification are needed.

Best regards,

Kristian

### DIFF
--- a/cdm/src/main/java/thredds/inventory/CollectionUpdater.java
+++ b/cdm/src/main/java/thredds/inventory/CollectionUpdater.java
@@ -74,15 +74,19 @@ public enum CollectionUpdater {
   }
 
   private CollectionUpdater() {
-    try {
-      scheduler = StdSchedulerFactory.getDefaultScheduler();
-      scheduler.start();
-      // scheduler.getListenerManager().addSchedulerListener(new MySchedListener());
-    } catch (Throwable e) {
-      failed = true;
-      throw new RuntimeException("quartz scheduler failed to initialize", e);
-    }
+	  start();
   }
+  
+  private void start() {
+	    try {
+	        scheduler = StdSchedulerFactory.getDefaultScheduler();
+	        scheduler.start();
+	        // scheduler.getListenerManager().addSchedulerListener(new MySchedListener());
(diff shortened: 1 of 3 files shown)


cwardgar:
Hi Kristian,

Have you tested this out? The (Quartz Javadocs)[http://quartz-scheduler.org/api/2.2.0/org/quartz/Scheduler.html#shutdown%28boolean%29] state that the scheduler cannot be restarted once it's shutdown.

Also, this change is not compatible with 5.0.0, as both `DataRootHandler` classes have been removed.

@JohnLCaron Any thoughts? If it were this easy to re-init collections, I think we'd have done it by now.

kristiansebastian:
Hi @cwardgar,

Yes, I have tested in my local environment and it works (the jobs are created and the collections are updated). Now, I don't know why because as you mention the scheduler shouldn't be restarted after shutdown. It is the first time that I work with the Quartz library, so you may know a better way to solve this issue.

Thanks!

cwardgar:
@kristiansebastian Thanks for your work on this. We've decided that this feature is better suited for version 5.0.0, where the relevant code is simpler and we can more thoroughly test it. I created an issue for it (#264) that you can watch if you'd like to follow our progress.

PR #236, opened 2015-10-12
Use NeedsExternalResource JUnit Category instead of ExternalServer

cwardgar:
NeedsExternalResource is enabled on Travis, disabled elsewhere.


cwardgar:
`TestLogicalReduce.testStationProfile()` is failing even though it only refers to a local file: `8602wh-a.ncml`. But in `8602wh-a.ncml`, we have:
````
location="dods://geoport.whoi.edu/thredds/dodsC/WFAL/8602wh-a.nc
````

Sneaky. I'll try running the tests with no internet connection to find the other sneaky ones.

cwardgar:
Back to green! From my last commit message:

 Finished applying NeedsExternalResource Category to appropriate tests.
* I didn't check for external resource references in :cdm-test or :it.
  Should try to track those down. However, those tests don't run on
  Travis, so it shouldn't be an immediate problem.
* I disconnected from the internet when running tests, so I should have
  caught everything.

lesserwhirls:
Sweet! Thanks @cwardgar!

PR #237, opened 2015-10-15
Removed bufferSize from PointFeature iterators

cwardgar:
A whole lot of code is devoted to propagating `bufferSize` through the point stack, but in practice, we're not actually taking advantage of it. In fact, if you look through the diff, you'll notice that every single time we specify a `bufferSize` value (e.g. `pointFeatureCollection.getPointFeatureIterator(-1)`), we always use `-1`, the default.

So it seems like nobody's using this feature, not even us. Furthermore, the preferred way to iterate over a `PointFeatureCollection` now is to use the for-each construct, wherein you can't specify a `bufferSize` anyway.

I'd really like to get rid of this cruft if I can. @JohnLCaron Is there any reason to keep it around?

### DIFF
--- a/bufr/src/main/java/ucar/nc2/ft/point/bufr/BufrFeatureDatasetFactory.java
+++ b/bufr/src/main/java/ucar/nc2/ft/point/bufr/BufrFeatureDatasetFactory.java
@@ -222,7 +222,7 @@ public class BufrFeatureDatasetFactory implements FeatureDatasetFactory {
         }
 
         @Override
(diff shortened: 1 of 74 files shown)


cwardgar:
For those interested, I tracked the movement of `bufferSize` through the stack, and it's ultimate utilized in exactly one place: [Structure.IteratorRank1.setBufferSize(int)](https://github.com/Unidata/thredds/blob/v4.6.3/cdm/src/main/java/ucar/nc2/Structure.java#L476) (via the `Table`/`NestedTable` machinery).

JohnLCaron:
i think its ok to remove.

PR #240, opened 2015-10-16
Add unit test for profiler rass data

JohnLCaron:


dopplershift:
Failed because it can't get to the file, which I assume we at some point need to include in our collection of test data. You need to add a `@Category(NeedsCdmUnitTest.class)`

JohnLCaron:
done, thanks

PR #261, opened 2015-10-28
S3: Handle case when a "directory" is a valid object.

dopplershift:
On S3, a prefix that we consider a directory can have an object
attached to it--even a 0 byte one. This breaks the crawling logic
deciding whether to treat something as a collection or not.

We consider a file as a valid S3 object, whereas a directory is not. However, on S3, while the keys look hierarchical they're not really; the key '2009/' can map to an object and then we can still have files under '2015/10/20/foobar'. We solve this by filtering '2009/' from the list of objects under the prefix '2009/'; it is doubtful that a "directory" would ever have valid data itself that a user would want to serve. This eliminates a confusing case for when a placeholder key was created in the S3 bucket.


cwardgar:
Superseded by #263.

PR #265, opened 2015-10-29
Asciidoctor changes

lesserwhirls:
adds a build script - run with no args

adds thredds specific css file

adds [[threddsDocs]] to all asciidoctor files, which add an id of threddsDocs to the body tag of all generated html (to help lock down css mods)


lesserwhirls:
Ok, I think this is ready.

PR #271, opened 2015-11-02
Fix calls to setContentLength(). (Fixes #197)

### LINKED ISSUE #197: Unicode catalog issue
Taken from e-support HNE-535275:

"""
I have a catalog in which I get a strange tag at the end of the page
(see attached image).

But, if I replace that o with acute with a plain o, the strange tag is
gone. As you can see in the image I'm using latest version (that problem
may be present in earlier versions anyway).

I think there is a point in which software is not correctly handling
UTF-8 strings. Maybe there is a point in the code in which byte length
(and not character length) is being used to print characters and that
could be adding the extra characters at the end of the page.

Anyway I'm attaching catalog if it helps to reproduce the problem.
"""

dopplershift:
setContentLength() sets the content-length HTTP header, which is defined
as the number of bytes. In all of these changed places, we were passing
in the length of the string, which is the number of characters. This
breaks in the presence UTF-8 characters, which may be encoded as up to 4
bytes; the result is truncated HTML data (observed in #197).

Original report saw stray characters on HTML view; I was not able to see these in chrome, but they were clear (only `</h` of the closing `</html>` tag) when viewing the HTML source. Two items that may warrant discussion:
- I'd like to add a test, but I'm not sure how best to get a useful one that checks the HTML returned when serving up an appropriate catalog. Pointers welcome.
- I hit all of the users of `setContentLength()`, but I'm not sure if that's all of the problem. Anyone know of anything else easily searchable that could be a victim of this?

### DIFF
--- a/tds/src/main/java/thredds/core/ConfigCatalogHtmlWriter.java
+++ b/tds/src/main/java/thredds/core/ConfigCatalogHtmlWriter.java
@@ -38,6 +38,7 @@ import thredds.client.catalog.*;
 import thredds.client.catalog.writer.DatasetHtmlWriter;
 import thredds.server.config.HtmlConfig;
 import thredds.servlet.HtmlWriter;
+import static thredds.servlet.ServletUtil.setResponseContentLength;
 import thredds.util.ContentType;
 import ucar.nc2.units.DateType;
 import ucar.unidata.util.Format;
@@ -87,15 +88,18 @@ public class ConfigCatalogHtmlWriter {
   public int writeCatalog(HttpServletRequest req, HttpServletResponse res, Catalog cat, boolean isLocalCatalog) throws IOException {
     String catHtmlAsString = convertCatalogToHtml(cat, isLocalCatalog);
 
-    res.setContentLength(catHtmlAsString.length());
+    // Once this header is set, we know the encoding, and thus the actual
+    // number of *bytes*, not characters, to encode
     res.setContentType(ContentType.html.getContentHeader());
+    int len = setResponseContentLength(res, catHtmlAsString);
+
     if (!req.getMethod().equals("HEAD")) {
       PrintWriter writer = res.getWriter();
       writer.write(catHtmlAsString);
       writer.flush();
     }
 
-    return catHtmlAsString.length();
+    return len;
   }
 
   /**
--- a/tds/src/main/java/thredds/server/cdmremote/CdmRemoteController.java
+++ b/tds/src/main/java/thredds/server/cdmremote/CdmRemoteController.java
@@ -161,9 +161,8 @@ public class CdmRemoteController extends AbstractController implements LastModif
 
           ncfile.setLocation(datasetPath); // hide where the file is stored
           String cdl = ncfile.toString();
-          res.setContentLength(cdl.length());
+          size = (long) thredds.servlet.ServletUtil.setResponseContentLength(res, cdl);
           pw.write(cdl);
-          size = cdl.length();
           break;
         }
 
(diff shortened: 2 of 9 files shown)


DennisHeimbigner:
Alternately, we could disallow setContentLength();
Is there anywhere that requires it?
=Dennis

Ryan May wrote:

dopplershift:
No idea as to the utility of setting the value. I can say that in a couple places we return this size out of our API (listed as "bytes"), so those still need to be corrected. Not sure what the ramifications are of that value being incorrect.

cwardgar:
For the test, you could check that the returned HTML is well-formed. That, at least, would catch the closing tag being `<\h` instead of `</html>`.

dopplershift:
For the test, I'm more clueless as the right layer to come in and how to set up testing a server response. Can I mock if I really want to test the actual server functionality?

cwardgar:
Do you want to test the behavior of a Controller? If so, yes, you can completely mock that interaction using the `spring-test` stuff. `AnyHtmlControllerTest` is a good example. Also, maybe you just want to test `ConfigCatalogHtmlWriter.writeCatalog()`? In that case, you could probably avoid the `WebAppConfiguration` stuff (since Spring isn't autowiring a controller for you) and just use `MockHttpServletRequest` and `MockHttpServletResponse`. You'll still probably want a real `Catalog` though.

Some related weirdness that I noticed: in 4.6 `ConfigCatalogHtmlWriter.writeCatalog()` is only being called from 2 unused controllers: `LocalCatalogServiceController2` and `RemoteCatalogServiceController2`. In 5.0, it's being called from `CatalogServiceController` and `RemoteCatalogServiceController`. So is it unused in 4.6? Why are we seeing the behavior in #197?

DennisHeimbigner:
Remember that the body of the message may not ascii or even utf8;
it could be binary, in which case the specified content length may be correct.

DennisHeimbigner:
In looking at some of the changes, we probably need to do something like this:
    X.setContentLength(string.getBytes(Charset.forName("UTF-8")).length())
at least for those cases where we have a string.

dopplershift:
Can you point me to a place I've changed where binary data may be present? As far as I can tell all places were either some version of a catalog or a place where it was explicitly converting something to a string already.

As far as this:
'''
In looking at some of the changes, we probably need to do something like this:
X.setContentLength(string.getBytes(Charset.forName("UTF-8")).length())
at least for those cases where we have a string.
'''
How exactly is that different from what I added in `ServletUtils.java`:
```java
   /**
* Set the proper content length for the string
*
* @param response the HttpServletResponse to act upon
* @param s the string that will be returned
* @return the number of bytes
* @throws UnsupportedEncodingException on bad character encoding
*/
public static int setResponseContentLength(HttpServletResponse response, String s) throws UnsupportedEncodingException {
 int length = s.getBytes(response.getCharacterEncoding()).length;
 response.setContentLength(length);
 return length;
}

```

DennisHeimbigner:
I must have been looking at the wrong thing,
because I thought I still saw things of the form
...setContentLength(s.length()).


Ryan May wrote:

JohnLCaron:
Setting reponse.setContentLength() gets the length into the HTTP header, and into threddsServlet.log. For a while I was trying to use it when possible.

If not set, i _think_ tomcat may set length if content is small enough to fit into buffer. otherwise, it cant set it (because the header must go before the content, who designed that?) so does a chunked-encoding thing and streams out the content.

Anyway, probably not very important to make it work (but thanks for getting it right when we do set it!) because the size does get set in tomcat's access.log, obviously written after the response is sent. perhaps i didnt realize that at the time i worked to get that number into threddsServlet.logs. perhaps makes client job easier in some cases if it knows in advance?

dopplershift:
This should be ready to merge now.

PR #284, opened 2015-11-17
moved back to old index.html page for netcdf-java

lesserwhirls:


lesserwhirls:
Note that I have a script that will check for changes and automatically build and push the docs to the web - this is also a good test to see if this is working.

PR #285, opened 2015-11-17
modified index.html again

lesserwhirls:


lesserwhirls:
This is what we are actually using on our webserver for v4.6

PR #289, opened 2015-11-18
Replace use of StringUtils.join from apache commons.

dopplershift:
Use Joiner from Guava instead.


cwardgar:
Is this in response to the security concerns related to Apache Commons? Or just because Guava is the new hotness?

dopplershift:
@JohnLCaron is apparently removing uses of it (I assume security), and my `RadarServerController` was the last offender.

Yours to merge if you deem so acceptable. :)

cwardgar:
Okay, if "org.apache.commons:commons-lang3:3.3.2" isn't being used anywhere anymore, let's make sure to remove it from dependencies.gradle. 5.0.0 is probably the place to do that though.

PR #307, opened 2015-11-28
Fix file leaks in grib tests.

JohnLCaron:
Ryan - have a look.


dopplershift:
Yeah, that looks better. My bad...

PR #326, opened 2015-12-08
Enforce use of shared dimensions in groups

cwardgar:
* Group.addDimension() and Group.addDimensionIfNotExists() methods now throw an exception if the dimension isn't shared.
* NetcdfFileWriter.addDimension() methods no longer have an "isShared" parameter. Such dimensions should always be shared and allowing them to be private is confusing and error-prone.
* Update client code that uses those methods.
* Add notes to UpgradingTo5.adoc

Also, an unrelated but too-small-for-its-own-PR commit:
* If there's an exception when writing NcML <values>, produce less spammy warning. Don't include a stack trace.


cwardgar:
Bump.

cwardgar:
This was screwing up a new PR I was trying to do, so I went ahead and merged it.

PR #338, opened 2015-12-18
Apply default calendar attribute for CF and COARDS

lesserwhirls:
Make sure that datasets following CF and COARDS get the default calendar attribute
applied when the attribute is missing.

Added explicit @Override annotations to Nimbus and CSMC conventions, which
extend the COARDS convention.


cwardgar:
Did you add the Codacy review, Sean? Not sure if you want to require it to be clean, but the issues it raised are reasonable, I think.

lesserwhirls:
I have not yet, but I certainly will take a look and try to get it cleaned
up. Thanks for the reminder Christian!

lesserwhirls:
Hmmm...It didn't run the Codacy check on this one. Not sure what's up there.

dopplershift:
Codacy is just *really* slow.

cwardgar:
Yeah, and some of its violations are really picky (e.g. it doesn't like when you declare multiple variables on the same line). Useful sometimes, but we definitely don't want to require it for every PR.

PR #339, opened 2015-12-22
Binary write

massonjr:
Update writeBinaryData/dumpData function to support non windows OS

"java.io.FileNotFoundException: C:/temp/file.bin (No such file or directory)"  would happen when attempting to write binary data. I added code to check for a non-windows system and change C:/temp/file.bin to /tmp/file.bin


cwardgar:
Thanks for pointing this out. We should probably pop up a `FileManager` dialog asking the user where they'd like to save the file, rather than just sticking it in the temp directory. There are multiple examples of us using that class in ToolsUI, if you'd like to take a shot at it.

cwardgar:
Hi @massonjr,

Can you use a `ucar.nc2.ui.widget.FileManager` instead of a `javax.swing.JFileChooser`? We use `FileManager` throughout ToolsUI and it'd be nice to have a consistent dialog. Here are some examples of its use from `ucar.nc2.ui.NcmlEditor`:

https://github.com/Unidata/thredds/blob/master/ui/src/main/java/ucar/nc2/ui/NcmlEditor.java#L63
https://github.com/Unidata/thredds/blob/master/ui/src/main/java/ucar/nc2/ui/NcmlEditor.java#L144
https://github.com/Unidata/thredds/blob/master/ui/src/main/java/ucar/nc2/ui/NcmlEditor.java#L208

`DatasetViewer` already has a `FileManager` data member ([here](https://github.com/massonjr/thredds/blob/binary_write/ui/src/main/java/ucar/nc2/ui/DatasetViewer.java#L104)), so you can just use that. It looks like you'll need to create a new instance in `NCdumpPane` though.

lesserwhirls:
Fixed by Unidata/thredds#382, so closing now.

PR #342, opened 2015-12-29
DatasetUrls; GaussianLats; CdmRemote, GCsummary

JohnLCaron:
various fixes and improvements from RDA tests.
also merge with unidata master.


cdm/src/test/java/ucar/nc2/TestUnsigned.java
```diff
@@ -113,7 +114,7 @@ public void testUnsignedWrap() throws IOException {
       float b = data.nextFloat();
       if (b < 0) hasSigned = true;
     }
-    assert !hasSigned;
+    Assert.assertTrue(!hasSigned);
```
dopplershift: Not going to hold up for this, but shouldn't that just be `Assert.assertFalse(hasSigned)`?

Issue #345, opened 2015-12-31
asciidoctor links to adoc files


lesserwhirls:
I thought asciidoctor could handle links to adoc files automagically, but it doesn't (my bad). Currently we try to link to adoc files using something like the following:

```
link:rel/path/to/doc.adoc[text to link]
```

However, we should use the following:

```
<<rel/path/to/doc.adoc,text to link>>
```

otherwise, we simply link to the adoc file itself.

Looks like a job for a regex...which means I should open another ticket, since we now have two problems :-)

cwardgar:
I left some comments about this on the pull request instead of here. Oops. https://github.com/Unidata/thredds/pull/346#issuecomment-168367403

Reopening this for now.

lesserwhirls:
Ok, I've made a pr remove the .adoc file extension and contain a # symbol. Let's see how this does. See PR Unidata/thredds#366

cwardgar:
Fixed by #366.

cwardgar:
Looks like we missed some:
````
[cwardgar@MacBook-Pro ~/dev/projects/thredds2] (5.0.0 %>)
$ egrep -r --include '*.adoc' 'link:.*\.adoc' docs/ | wc -l
     117
````
Reopening.

PR #346, opened 2015-12-31
Fixed links to adoc files

lesserwhirls:
Changed links, according to Unidata/thredds#345, using the following:

```
find . -type f -name "*.adoc" | xargs -n 1 perl -pi -e 's/link:(.*)\[(.*)\]/<<\1,\2>>/g'
```


lesserwhirls:
Ok, my regex was a little too handsy, so I updated it from:

'''
find . -type f -name "*.adoc" | xargs -n 1 perl -pi -e 's/link:(.*)\[(.*)\]/<<\1,\2>>/g'
'''

to

'''
find . -type f -name "*.adoc" | xargs -n 1 perl -pi -e 's/link:(.*\.adoc.*)\[(.*)\]/<<\1,\2>>/g'
'''

lesserwhirls:
It's ok, I'll make a new pull request. This one covered all links, including those for html files.

cwardgar:
It's worth noting that this commit broke the links in the pages that GitHub renders from the asciidoc. Here's an example: https://github.com/lesserwhirls/thredds/blob/fc1d51bc357864613571ef4543fe804958934745/docs/website/netcdf-java/CDM/CFdiff.adoc

It's possible to preserve those links (https://github.com/asciidoctor/asciidoctor/issues/417#issuecomment-36426214) but there are some nasty issues, like referencing a doc that lives in a parent directory (https://github.com/asciidoctor/asciidoctor/issues/844#issuecomment-151968001). I made those changes to CFdiff.adoc and was able to get the links working properly: https://github.com/cwardgar/thredds/pull/2/files

Very kludgy; I don't want to make those changes to all of our docs. Apparently this will be easier once GitHub upgrades to AsciiDoctor v1.5.3 (or possibly v1.5.4). We should revisit it then.

cwardgar:
Also, [this comment](https://github.com/asciidoctor/asciidoctor/issues/417#issuecomment-21982031) indicates that xrefs should omit the file extension and always contain a `#` symbol. Are the links in the AsciiDoctor-generated HTML working even though we aren't doing that?

lesserwhirls:
Looks like the ability to link to adoc file in parent directories now works (asciidoctor 1.5.4). I had to upgrade, but this is a very good thing! It works just like linking to any other adoc file:

```
<<my.adoc#, My Ascidoctor file>>
```
and
```
<<../../my.adoc#, My relative Ascidoctor file>>
```
both work (second didn't in 1.5.3).

PR #355, opened 2016-01-06
Finished converting opendap tests to JUnit 4

cwardgar:
dap4 tests were partially done in c8eee04cc971fc229721ef5279075b3910e9e15d. Unfortunately, many test methods were not marked with the `@Test` annotation. So, those tests haven't been recognized by Gradle or run as part of our test suite since October. It's really troubling that a bunch of tests just disappeared and nobody noticed.

Most of the dap4 tests really ought to be parameterized, but I'm not going to do that work. That way, the outcome of each test case could be reported individually, instead of a single method failing if any of the cases fail.

While reviewing the dap4 tests, I noticed that 6 test classes aren't being included:
````
dap4/test/TestFilters.class
dap4/test/TestFrontPage.class
dap4/test/TestH5Iosp.class
dap4/test/TestHyrax.class
dap4/test/TestPrinter.class
dap4/test/TestSerial.class
````
@DennisHeimbigner, is that intended? Are any critical code paths being excluded from testing by not running those classes?

The other opendap tests had already been converted to JUnit 4. `TestEncode` just got missed, I think.


cwardgar:
Okay, so re-enabling those tests resulted in test failures (not surprising). As a result, we can't merge this into master until they're fixed. So, I'm going to push this work to an upstream branch so that someone (@DennisHeimbigner ?) can work on it.

cwardgar:
Continued in #356. No reason to leave this PR open anymore.

PR #356, opened 2016-01-06
Finished converting opendap tests to JUnit 4 (take 2)

cwardgar:
Continuing from #355. Everybody on the project should be able to push to the "convert-dap4-to-junit4" branch. This PR will be automatically updated when that happens.


cwardgar:
`dap4.test.TestServlet` is failing in Travis right now. It's very weird: I can't recreate it using the IntelliJ test runner. I *can* recreate it using the Gradle runner, but not when I run it individually. That is, in
````
./gradlew --info cleanTest :dap4:d4tests:test --tests *TestServlet
````
`TestServlet` passes. But in
````
./gradlew --info cleanTest :dap4:d4tests:test
````
it fails. So that suggests that there's some interaction with the other tests that's causing the problem. I'll dig in some more.

DennisHeimbigner:
I was waiting for test-dev and remotetest to be up again.
=Dennis


Christian W wrote:

cwardgar:
This one was incredibly tricky. I was banging my head against the wall until I noticed–almost by accident–that the failing dataset, `dap4/d4tests/src/test/data/resources/testfiles/test_fill.nc`, was being rendered one way in ToolsUI:
````
netcdf /Users/cwardgar/dev/projects/thredds/dap4/d4tests/src/test/data/resources/testfiles/test_fill.nc {
  variables:
    byte uv8;
      :_Unsigned = "true";

    short v16;

    int uv32;
      :_FillValue = 17U; // int
      :_Unsigned = "true";
}
````
and in a different way in the test code:
````
netcdf /Users/cwardgar/dev/projects/thredds/dap4/d4tests/src/test/data/resources/testfiles/test_fill.nc {
  variables:
    byte uv8;
      :_Unsigned = "true";

    short v16;

    int uv32;
      :_Unsigned = "true";
      :_FillValue = 17; // int
}
````
Notice that in the latter, `uv32:_FillValue = 17` lacks a trailing `U`, meaning that it's *not* unsigned, as it should be.

Why the difference? After a bunch of digging, I found that the former was generated by `ucar.nc2.iosp.hdf5.H5iosp` while the latter was generated by `ucar.nc2.jni.netcdf.Nc4Iosp`. `H5iosp` produces a correct result.

As I mentioned above, the `TestServlet` failures only occur in specific testing scenarios, so it would seem that SOMETIMES `TestServlet` is using `H5iosp` and sometimes it is using `Nc4Iosp`.

As I predicted, interaction with another test was ultimately the culprit. `dap4.test.TestNc4Iosp` *explicitly registers `Nc4Iosp` in preference to all other IOSPs*. So, whenever `TestNc4Iosp` ran before `TestServlet`, the failure would occur. In any other scenario, it would succeed. GAH!

This commit deletes `TestNc4Iosp.<clinit>`, where we were unconditionally registering `Nc4Iosp`.

Issue #170, opened 2015-09-04
Asciidoc Cleanup


lesserwhirls:
Many warnings when converting asciidoc files to html regarding headings / sections out of sequence.

```
asciidoctor: WARNING: PointFeatures.adoc: line 24: section title out of sequence: expected level 2, got level 3
asciidoctor: WARNING: CdmrCoverageParams.adoc: line 1: section title out of sequence: expected level 1, got level 2
asciidoctor: WARNING: CdmrFeaturePoint.adoc: line 53: section title out of sequence: expected level 2, got level 3
asciidoctor: WARNING: CdmrFeaturePoint.adoc: line 97: section title out of sequence: expected level 2, got level 3
asciidoctor: WARNING: NcStreamGrammer.adoc: line 129: list item index: expected 3, got 1
asciidoctor: WARNING: NcStreamGrammer.adoc: line 130: list item index: expected 4, got 2
asciidoctor: WARNING: NcStreamGrammer.adoc: line 137: list item index: expected 4, got 1
asciidoctor: WARNING: NcStreamGrammer.adoc: line 138: list item index: expected 5, got 2
asciidoctor: WARNING: FeatureCollections.adoc: line 501: section title out of sequence: expected level 2, got level 3
asciidoctor: WARNING: Grib.adoc: line 8: section title out of sequence: expected level 2, got level 3
asciidoctor: WARNING: Grib.adoc: line 16: section title out of sequence: expected level 2, got level 3
asciidoctor: WARNING: Grib.adoc: line 21: section title out of sequence: expected level 2, got level 3
asciidoctor: WARNING: Grib.adoc: line 26: section title out of sequence: expected level 2, got level 3
asciidoctor: WARNING: Grib.adoc: line 33: section title out of sequence: expected level 2, got level 3
asciidoctor: WARNING: Grib.adoc: line 41: section title out of sequence: expected level 2, got level 3
asciidoctor: WARNING: Grib.adoc: line 62: section title out of sequence: expected level 2, got level 3
asciidoctor: WARNING: Grib.adoc: line 64: section title out of sequence: expected level 2, got level 3
asciidoctor: WARNING: Grib.adoc: line 66: section title out of sequence: expected level 2, got level 3
asciidoctor: WARNING: Grib.adoc: line 68: section title out of sequence: expected level 2, got level 3
asciidoctor: WARNING: Grib.adoc: line 71: section title out of sequence: expected level 2, got level 3
[...]

PR #366, opened 2016-01-10
Clean up adoc xref links

lesserwhirls:
Remove .adoc file extension in link

end link with # symbol


cwardgar:
The diff looks good. Of course, we won't know for sure until we [clean up the AsciiDoctor warnings](https://github.com/Unidata/thredds/issues/170#issuecomment-170443585) and get a complete build of our documentation.

Issue #365, opened 2016-01-09
Upgrade to Apache httpclient 4.5


DennisHeimbigner:
Currently, thredds (4.6x) uses apache httpclient 4.5 as its library, but it uses it as if it were httpclient 4.2. This leads to a large number of deprecated warnings.

Apache httpclient 4.3 and later have made major changes to the httpclient/core/mime interfaces by moving to an immutable+builder model. It is desirable to upgrade thredds to properly use 4.3 and later and most notably, the most recent httpclient: version 4.5.

Additionally, httpclient 4.5 has made significant changes in authentication handling. This forces a major re-think about how thredds httpservices module handles authentication.

DennisHeimbigner:
Fixed, by pull request 367.
Solution(s) as follows:

Move httpservices to properly use Apache httpclient 4.5 and later;
actually 4.3 and later, but 4.5 appears to be a relatively stable
version.

The big change to Apache httpclient 4.5 is that the httpclient/core/mime
interfaces have moved to an immutable+builder model. This means that,
for example, to create an instance of HttpClient, one must use a client
builder for which various parameters are set. Then when finished, the
.build() method is called to create an immutable HttpClient object that
can be used for, for example, method execution. Similary, one must use a
RequestBuilder in order to construct a Reques object to do GET, HEAD, etc.

Whatever the merits of the immutable model, it is a pain to use at the
client level. So httpservices now hides the immutability and continues
to allow the client to set parameters as needed. It caches the immutable
objects and invalidates them when the client makes parameter changes
such as changing the credentials or user agent.

Additionally, httpclient 4.5 has made significant changes in
authentication handling, most notably:
* Httpclient now makes externally visible the HttpClientContext
  object; this object provides caching for credentials and cookies.
* CredentialsProvider now can contain multiple (AuthScope,Credentials)
  pairs. This is how, for example, one can communicate to a site
  that is password protected thru a proxy that is itself password
  protected: both sets of password credentials are stored in the
  CredentialsProvider.
* There is now built-in support for accepting self-signed certificates
  from a server.
* SSH handling is now much simpler and there is no longer a need for
  CustomSSLProtocolSocketFactory and CustomX509TrustManager.

These changes force a number of internal changes to httpservices.
* The httpservices module must now maintain an HttpClientContext object
  on a per-session object basis and use it for all http calls (GET,
  HEAD, etc). The context allow us to properly cache credentials and
  cookies.
[...]

PR #367, opened 2016-01-11
Fix thredds issue https://github.com/Unidata/thredds/issues/365

DennisHeimbigner:
Move httpservices to properly use Apache httpclient 4.5 and later;
actually 4.3 and later, but 4.5 appears to be a relatively stable
version.

The big change to Apache httpclient 4.5 is that the httpclient/core/mime
interfaces have moved to an immutable+builder model. This means that,
for example, to create an instance of HttpClient, one must use a client
builder for which various parameters are set. Then when finished, the
.build() method is called to create an immutable HttpClient object that
can be used for, for example, method execution. Similary, one must use a
RequestBuilder in order to construct a Reques object to do GET, HEAD, etc.

Whatever the merits of the immutable model, it is a pain to use at the
client level. So httpservices now hides the immutability and continues
to allow the client to set parameters as needed. It caches the immutable
objects and invalidates them when the client makes parameter changes
such as changing the credentials or user agent.

Additionally, httpclient 4.5 has made significant changes in
authentication handling, most notably:
* Httpclient now makes externally visible the HttpClientContext
  object; this object provides caching for credentials and cookies.
* CredentialsProvider now can contain multiple (AuthScope,Credentials)
  pairs. This is how, for example, one can communicate to a site
  that is password protected thru a proxy that is itself password
  protected: both sets of password credentials are stored in the
  CredentialsProvider.
* There is now built-in support for accepting self-signed certificates
  from a server.
* SSH handling is now much simpler and there is no longer a need for
  CustomSSLProtocolSocketFactory and CustomX509TrustManager.

These changes force a number of internal changes to httpservices.
* The httpservices module must now maintain an HttpClientContext object
  on a per-session object basis and use it for all http calls (GET,
  HEAD, etc). The context allow us to properly cache credentials and
  cookies.
* All non-global credentialing and cookies are scoped by an HTTPSession
  instance so that as long as that instance exists, the same credentials
  and cookies are used by all GET, HEAD, etc. method executions.  Of
  course global credentials setting is still supported.
* There is no longer a need to specify the kind of scheme being used
  (e.g. BASIC, DIGEST, etc) because that is implied by the credentials
  provider.

An additional word on HTTPSession as a scope. Apache httpclient
effectively limits the part of the url for which a session
can be defined. Specifically a scope is the combination of host+port.
Previously, httpservices attempted to provide a finer grain scope
that included the path. That did not work, so scope is now
reduced to using host+port, which can be extracted from a URL,
and is best embodied by the apache HttpHost and AuthScope classes.

Other Major Changes:
* Force all creation of HTTPSession and HTTPMethod instances to be
  performed thru the HTTPFactory API.
* Deprecate a number of API methods no longer needed.
* Add methods to avoid the client having to explicitly set request
  headers such as compression. Directly setting headers is still
  allowed, but is undesirable.
* Fix numerous test cases and other classes to use the revised
  httpservices API by getting rid of deprecated method calls.
* Document the new httpservices API
* Modify UrlAuthenticatorDialog.java in ui module so that it acts like
  the proper new kind of CredentialsProvider (see above).
* Move towards a fluent interface by changing a number of set methods to
  return the containing object instead of having a void return value.
* Remove a number of classes no longer needed because of changes in the
  httpclient API (esp. the authentication API).
* Remove misc. junk files.

To Do:
* Verify that TDM now works.
* Get OAUTH2 support working so that URS authentication can be supported.


cwardgar:
Cool! My only note is that you made some changes to the documentation and we have to make sure that those are made against the AsciiDoc on 5.0.0 as well.

DennisHeimbigner:
ok

PR #368, opened 2016-01-12
Tests delete the files that they create

cwardgar:
* All of these tests created files in TestLocal.temporaryDataDir ("target/test/tmp") and relied on the build tool to clean them.
* Now using JUnit's TemporaryFolder @Rule, which guarantees that it will delete temp files after a test finishes.
* ucar.nc2.TestLocal is no longer being used anywhere. Deleted it.
* ucar.nc2.TestAll was being used by one class, TestAggMisc. Transferred needed methods and deleted TestAll.
* Did additional cleanup to the affected test classes:
  - Used try-with-resources where possible
  - Removed unnecessary spamming of STDOUT
  - Converted JUnit 3 to JUnit 4
  - Added assertions to some tests that weren't actually testing anything. There are still a LOT of these lurking.
  - Used JUnit's @FixMethodOrder to address a test order dependency in ucar.nc2.TestWrite

Coverity will catch fire and explode if we ever make it analyze our tests. *shudder*


cwardgar:
Bump

cwardgar:
Rebased.

lesserwhirls:
Looks ok to me - thanks @cwardgar! If you rebase this one, I can merge it in and then we can merge in Unidata/thredds#372 since it will need to be rebased, too.

PR #373, opened 2016-01-17
Apply Fix for thredds issue https://github.com/Unidata/thredds/issues…

DennisHeimbigner:
…/365 to version 5.0.0

Apply (most) of the pull request
https://github.com/Unidata/thredds/pull/367
changes to version 5.0.0. That pull request fixed issue
https://github.com/Unidata/thredds/issues/365, which upgraded support
for Apache httpclient 4.5.

The goal here is to make later merges of master into 5.0.0 simpler by
doing the changes now in 5.0.0. Note that this has only been tested by
Travis since Jenkins is down.


cwardgar:
Good idea. I had to remind myself that this sort of thing — applying *part* of a commit's changeset and then later applying the whole thing in a merge — is safe to do in Git. Not a good idea in Subversion or CVS.

PR #376, opened 2016-01-20
GRIB: Handle missing values for more time interval fields (Fixes #375)

### LINKED ISSUE #375: 5.x errors with National Blend grib2 files
So the 5.x branch throws:
```java.lang.RuntimeException: TimeInterval(2) has different units```
when reading [National_Blend_min_temp.grib2](https://files.zenhub.io/569ed287bb5ef2e60394ae79)
Uploaded using [**ZenHub.io**](http://www.zenhub.io)

(I pulled out only the relevant field that causes problems.) It looks like they didn't encode any kind of interval (a bunch of 255's) to go with the minimum/maximum temps, even though they're supposed to be daily values.

@JohnLCaron want to take a look and see if there's something more sensible lurking somewhere? Right now I can fix the error and get things to read in without error, but then it calls values "Minimum temperature (255_Hour Minimum)" which seems sub-optimal, but could (horribly) be the best we can do.

dopplershift:
Handle 255's that we see in the National Blend of Models files. Mimics other code in these areas that are already handling 255's for some fields.

### DIFF
--- a/grib/src/main/java/ucar/nc2/grib/grib2/table/Grib2Customizer.java
+++ b/grib/src/main/java/ucar/nc2/grib/grib2/table/Grib2Customizer.java
@@ -218,12 +218,12 @@ public class Grib2Customizer implements ucar.nc2.grib.GribTables, TimeUnitConver
     // calculate total "range"
     int range = 0;
     for (Grib2Pds.TimeInterval ti : pdsIntv.getTimeIntervals()) {
-      if ((ti.timeRangeUnit != timeUnitOrg) || (ti.timeIncrementUnit != timeUnitOrg && ti.timeIncrementUnit != 255 && ti.timeIncrement != 0)) {
+      if ((ti.timeRangeUnit != timeUnitOrg && ti.timeRangeUnit != 255) || (ti.timeIncrementUnit != timeUnitOrg && ti.timeIncrementUnit != 255 && ti.timeIncrement != 0)) {
         log.warn("TimeInterval has different units timeUnit org=" + timeUnitOrg + " TimeInterval=" + ti.timeIncrementUnit);
         throw new RuntimeException("TimeInterval(2) has different units");
       }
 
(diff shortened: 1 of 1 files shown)


dopplershift:
I'd like @JohnLCaron to weigh in here.

JohnLCaron:
So its likely that they said "we dont need/understand these time intervals, so lets set to 255"

Im guessing that they just want to indicate that the interval goes from forecast time to "end of overall time interval" and  im sympathetic that easy things should be easy.

1) do we think thats the correct thing to put here? is there a way to check, eg against wgrib2?

2) we can put in an override in NcepLocalTables, if we assume this is used for all NCEP. If should just be used for NECP subcenter 14, we would add a new localTable class for that. safer to override just subcenter 14 in terms of messing other things up.

 we can either make a guess or try to contact them to find out.

JohnLCaron:
If someone could run wgrib2 on a few records, lets see what they say and try to follow that, by adding a new subclass of NcepLocalTables.

JohnLCaron:
Can one of you domain experts look at this file and decide what correct value is? And/or contact them and/or run wgrib2.

dopplershift:
So do we need this or not? I've put an example of a fixed file in #375, but there remains the issue of the files on NOAAPORT until then. Is your fix using a local table enough @JohnLCaron ?

JohnLCaron:
I think we should run wgrib2 on them and see if that agrees with what i did. Also, i think one of you "domain experts" should vet this against your understanding of what they are doing.

dopplershift:
Define "them". New files or old files?

JohnLCaron:
i thought you were talking about the old files.

i havent looked at new ones yet, but i will get to it.

dopplershift:
I can check to see if what you have going makes sense easily enough...

JohnLCaron:
Can close it, see comments in #375.

PR #380, opened 2016-01-23
Grib1GDS BigDecimal implementation to fix floating point errors

### LINKED ISSUE #340: Out of range Lon normalized incorrectly for grib file.
Outputting the stats of the file in question with wgrib I get the following:

```
rec 1:0:date 2015120900 TMPA kpds5=25 kpds6=105 kpds7=2 levels=(0,2) grid=211 2 m above gnd valid 30-30d: bitmap: 55047 undef
  TMPA=Temp. anomaly [K]
  timerange 2 P1 30 P2 30 TimeU 2  nx 360 ny 179 GDS grid 0 num_in_ave 0 missing 0
  center 250 subcenter 1 process 255 Table 2 scan: WE:SN winds(grid) 
  latlon: lat  -89.001000 to 88.998000 by 0.000000  nxny 64440
          long -180.000000 to 179.000000 by 0.000000, (360 x 179) scan 64 mode 8 bdsgrid 1
  min/max data -0.7 3.7  num bits 6  BDS_Ref -7  DecScale 1 BinScale 0
```

but when I read in the same file via GridDataset (`GridDataset gridDataset = GridDataset.open(filename);`) the lon range is wrong.

```
lat -> min: -89.00100708007812 max: 88.99800872802734
lon -> min: 179.99998474121094 max: 539.0
```

I tracked the issue down to [this](https://github.com/Unidata/thredds/blob/5612f5ff64823b8bd36513887236df48748c4290/cdm/src/main/java/ucar/unidata/geoloc/LatLonPointImpl.java#L121)

```
  static public double lonNormal(double lon) {
    if ((lon < -180.) || (lon > 180.)) {
      return Math.IEEEremainder(lon, 360.0);
    } else {
      return lon;
    }
  }
```

Which normalizes the value `-180.00001525878906` to the positive value of  `179.99998474121094`.

I just naively updated the `lonNormal` function to:

```
static public double lonNormal(double lon) {
    if (lon < -180.) {
      return -180.;
    } else if (lon > 180.) {
      return 180.;
    } else {
      return lon;
    }
  }
```

Which is similar to what `latNormal` does, but this breaks a number of tests.  So I think I may be fixing this in the wrong place.

rgaskill:
fixes #340

### DIFF
--- a/cdm/src/main/java/ucar/nc2/util/Misc.java
+++ b/cdm/src/main/java/ucar/nc2/util/Misc.java
@@ -33,11 +33,10 @@
 
 package ucar.nc2.util;
 
-import com.google.common.primitives.Floats;
-
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
+import java.math.BigDecimal;
 import java.util.*;
 
 /**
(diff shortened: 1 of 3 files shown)


cwardgar:
@JohnLCaron Was this issue addressed in #392? Should I close this PR?

PR #382, opened 2016-01-25
Refactored ToolsUI's "write binary data to file" to use FileManager

cwardgar:
Builds on massonjr's commits in #339.


lesserwhirls:
Thanks @massonjr and @cwardgar!

PR #383, opened 2016-01-26
Test cleanup

dopplershift:
Some clean-ups to some tests. Adds names for some parameterized tests, moves some prints to actual logging messages, and gives a much more useful message when we fail a comparison on a single HTTP code (instead of `AssertionError` we actually get a message saying what the codes were.

I was hoping to run this on jenkins to see if I can use this to see why all those tests are failing, but I can't get the integration tests to run on anything but master or 5.0.0 on jenkins.

### DIFF
--- a/httpservices/src/main/java/ucar/httpservices/HTTPSession.java
+++ b/httpservices/src/main/java/ucar/httpservices/HTTPSession.java
@@ -811,7 +811,7 @@ public class HTTPSession implements Closeable
             throws HTTPException
     {
         assert (scope != null);
-        if(actualurl == null)
+        if(actualurl != null)
             this.sessionURI = actualurl;
         else
             this.sessionURI = HTTPAuthUtil.authscopeToURI(scope).toString();
--- a/it/src/test/java/thredds/TestWithLocalServer.java
+++ b/it/src/test/java/thredds/TestWithLocalServer.java
@@ -78,9 +78,11 @@ public class TestWithLocalServer {
       HTTPMethod method = HTTPFactory.Get(session);
       int statusCode = method.execute();
 
-      if (expectCodes == null)
+      if (expectCodes == null) {
         Assert.assertEquals(200, statusCode);
-      else {
+      } else if (expectCodes.length == 1) {
+        Assert.assertEquals(expectCodes[0], statusCode);
+      } else {
         boolean ok = false;
         for (int expectCode : expectCodes)
           if (expectCode == statusCode) ok = true;
--- a/it/src/test/java/thredds/TestWithLocalServer.java
+++ b/it/src/test/java/thredds/TestWithLocalServer.java
@@ -53,6 +53,7 @@ import java.io.InputStream;
  * @since 10/15/13
  */
 public class TestWithLocalServer {
+  static private org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(TestWithLocalServer.class);
   public static String server = "http://localhost:8081/thredds/";
 
   public static String withPath(String path) {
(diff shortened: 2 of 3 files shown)


cwardgar:
@dopplershift You want to run the Gradle task `testAll`, not `test`. `test` will run tests for all subproject **except** `it`, because `it`'s test task is named `integrationTest` instead. I've made the change to your Jenkins config and kicked off a build. I also fixed a couple other configs that were using `test` instead of `testAll`.

cwardgar:
Changes look good, except that `logging.properties` should be in `it/src/test/resources`, not projectRoot. Yes, that means that each subproject will need its own test-logging config file, but I think that's okay — preferable, even. Plus, it avoids cluttering projectRoot.

Also, I don't think there's a good reason why we're using java.util.logging (JUL) instead of [logback](http://logback.qos.ch/) for test logging. Logback is superior in [just about every way](http://stackoverflow.com/questions/11359187/why-not-use-java-util-logging/11360517#11360517). We could go ahead and make the change, in which case you'd write a `logback.groovy` config file instead. Up to you.

As for logging in production, we have some log-parsing code that makes it more difficult to switch to Logback. It's not the easy transition that test logging is.

dopplershift:
Regarding `logging.properties`, I'm ok with nuking the file from this PR. I only included it because I wanted to have some easily available config to make the tests noisy again--the tests work completely fine without any config file. I will say that I originally put the file in `it/src/test/resources`, but the file was still not picked up automatically; given that, it didn't seem useful to put the config file in that deeply nested a spot. The top-level might not be the right spot, but if it's not picked up automatically, it makes the most sense to me to put a single config in a spot that's easy enough to specify on the command line.

I'd be ok with using logback--would the way I've created the loggers in this PR need to change in that event?

dopplershift:
Jenkins changes worked great, thanks @cwardgar. Looks like the problem is that links that should resolve are throwing 404's. Not sure how @DennisHeimbigner 's changes caused them but:
- the change coincides with the start of failures
- the function involved in test was changed
Something to continue to investigate...

cwardgar:
Weird that the config file was not picked up automatically; that may be a peculiarity of JUL. In Logback and others, the config file simply needs to be on the classpath, and `it/src/test/resources` is certainly on `it`'s test classpath.

Logback is fully compatible with SLF4J, so the API calls in the code wouldn't need to change. We'd just need to depend on "logback" instead of "slf4j-jdk14" [here](https://github.com/Unidata/thredds/blob/master/gradle/testing.gradle#L148). Looks like we'd need to add logback to dependencies.gradle too. Super simple change. You'd also have to write a new logging config file, but only if you need something different than the default settings.

dopplershift:
Ok, I nuked `logging.properties` since I don't think it's long for this world.

DennisHeimbigner:
Tracked down one case of 404: TestMetadataService.testXML
It appears to be coming from the server in accessing this url:
http://localhost:8081/thredds/metadata/gribCollection/GFS_CONUS_80km/GFS_CONUS_80km_20120227_0000.grib1?metadata=variableMap&accept=xml
There is a null pointer exception occurring at InvDatasetFcGrib, line 645.
    GribCollectionImmutable.Dataset ds = gc.getDatasetByTypeName(paths.get(0));
The value of gc is null; no idea why.

DennisHeimbigner:
TestOpendapFIlters - this is true client-side error. Since I do not have access to the actual pull request
(I think) I will have to give the fix here:
In file branch 5.0.0 thredds.httpservices/src/main/java/ucar/httpservices/HTTPSession.java, line 813
Change:
         if(actualurl == null)
to
        if(actualurl != null)
Same problem may exist in master, so will do separate pull request if so.

dopplershift:
Cherry-picking that commit fixed up 55 failing tests. I think this can go in, unless there are some other objections?

PR #398, opened 2016-02-01
Add XML callouts to CatalogPrimer.adoc, as a test.

cwardgar:
See [here](http://www.unidata.ucar.edu/software/thredds/v5.0/test/CatalogPrimer.html) for the rendered HTML using AsciiDoctor's default CSS. I'm curious if the callouts will still render properly with the Unidata CSS that we're using.

### DIFF
--- a/docs/website/tds/tutorial/CatalogPrimer.adoc
+++ b/docs/website/tds/tutorial/CatalogPrimer.adoc
@@ -1,6 +1,8 @@
 :source-highlighter: coderay
 [[threddsDocs]]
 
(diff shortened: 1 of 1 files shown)


lesserwhirls:
Give it a minute and we'll see how it looks.

lesserwhirls:
Looks like it does not work...I think we're missing some css in our TDS css file: https://github.com/Unidata/thredds/blob/5.0.0/docs/website/css/master_thredds_adoc.css

lesserwhirls:
ok, I've added the css we need, I think...see Unidata/thredds#399

PR #404, opened 2016-02-03
Cleaned up remaining AsciiDoctor warnings.

cwardgar:
Will Travis be green now? The Shadow knows.


dopplershift:
The stderr thing is what happens when you convert from your familiar `2>&1`, badly.

Issue #214, opened 2015-09-28
Getting 5.0.0 green on Travis


cwardgar:
I thought I could fix all remaining test failures in #211, but I was wrong. In [our latest Travis build result](https://travis-ci.org/Unidata/thredds/builds/82515716), we now have a problem in dap4:
````
dap4.test.TestConstraints > testConstraints STANDARD_ERROR
    java.lang.IllegalArgumentException: setCachedData type=class java.lang.Byte incompatible with variable type=ubyte
    	at ucar.nc2.Variable.setCachedData(Variable.java:1569)
    	at ucar.nc2.Variable.setCachedData(Variable.java:1554)
    	at ucar.nc2.Variable._read(Variable.java:807)
    	at ucar.nc2.Variable.read(Variable.java:681)
    	at ucar.nc2.dataset.VariableDS.reallyRead(VariableDS.java:523)
    	at ucar.nc2.dataset.VariableDS._read(VariableDS.java:507)
    	at ucar.nc2.Variable.read(Variable.java:681)
    	at ucar.nc2.NCdumpW.print(NCdumpW.java:340)
    	at ucar.nc2.NCdumpW.print(NCdumpW.java:242)
    	at dap4.test.TestConstraints.ncdumpdata(TestConstraints.java:248)
    	at dap4.test.TestConstraints.doOneTest(TestConstraints.java:188)
    	at dap4.test.TestConstraints.testConstraints(TestConstraints.java:162)
    	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    	at java.lang.reflect.Method.invoke(Method.java:483)
    	at junit.framework.TestCase.runTest(TestCase.java:176)
    	at junit.framework.TestCase.runBare(TestCase.java:141)
    	at junit.framework.TestResult$1.protect(TestResult.java:122)
    	at junit.framework.TestResult.runProtected(TestResult.java:142)
    	at junit.framework.TestResult.run(TestResult.java:125)
    	at junit.framework.TestCase.run(TestCase.java:129)
    	at junit.framework.TestSuite.runTest(TestSuite.java:252)
    	at junit.framework.TestSuite.run(TestSuite.java:247)
[...]

PR #411, opened 2016-02-05
Remove System.exit(1) calls from TestConstraints

cwardgar:
On Jenkins, this was causing:
````
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':dap4:d4tests:test'.
> Process 'Gradle Test Executor 13' finished with non-zero exit value 1
````
5.0.0 builds have been failing on Jenkins for a while as a result.

As stated in #214, this PR merely fixes the build failure; it does not fix the broken test.


cwardgar:
I need to publish a new 5.0.0-SNAPSHOT and I'm blocking on this, so I'm going to merge my own PR.

Issue #378, opened 2016-01-21
NCSS Point Collections date problems


dopplershift:
When making an NCSS request against a PointFeature collection, when giving it request for newer data (since the server started up), it incorrectly says that the request does not overlap with the actual data date range. The metadata for the collection agrees properly with the data on disk.

PR #413, opened 2016-02-08
Point Collection FeatureCollections update working

lesserwhirls:
Fixed Unidata/thredds#378

The update function would update the local state of the available date/time range
of the dataset in the catalog, but this was not being properly checked in the
request validation. Note that this will not scale well, as the validation bean
checks the date/time range on the CompositePointDataset object, which only  gets
updated on server restart.

### DIFF
--- a/tds/src/main/java/thredds/server/ncss/params/NcssParamsBean.java
+++ b/tds/src/main/java/thredds/server/ncss/params/NcssParamsBean.java
@@ -39,6 +39,7 @@ import thredds.server.ncss.validation.NcssRequestConstraint;
 import thredds.server.ncss.validation.TimeParamsConstraint;
 import thredds.server.ncss.validation.VarParamConstraint;
 import ucar.nc2.ft.FeatureDataset;
+import ucar.nc2.ft.point.collection.UpdateableCollection;
 import ucar.nc2.time.Calendar;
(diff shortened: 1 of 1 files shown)


tds/src/main/java/thredds/server/ncss/params/NcssParamsBean.java
```diff
@@ -405,6 +407,20 @@ public boolean isValidGridRequest() {
...
+      // if this is an updatable collection, check for a new CalendarDateRange (i.e. new
+      // data available).
+      // The issue is that fd does not get updated with thredds.catalog.InvDatasetFcPoint
+      //
+      // This allows requests to be fulfilled, even if the correct CalendarDateRange is
+      // not available in the catalog. Ideally these should be synced.
+      //
+      UpdateableCollection uc = (UpdateableCollection) fd;
+      have = uc.update();
+    } catch (IOException e) {
```
dopplershift: Probably need to log an error on the `IOException` and catch (and swallow) any errors from casting to `UpdatableCollection`.

lesserwhirls: Ok, this should be fixed in the updated PR. Thanks!

PR #422, opened 2016-02-11
Provide core netcdf4 jna not intertwined with CDM

DennisHeimbigner:
Break netcdf4 module into two parts:
1. netcdf4-core that does not depend on cdm
2. netcdf4 that adds the cdm iosp.
Reason: I need a pure (i.e. no CDM dependencies) netcdf-4
for use with dap4 d4ts test server.


cwardgar:
I feel like there's probably a simpler solution to this, possibly powered by Gradle. So you need to exclude :cdm from :dap4:d4ts? Would something like this in `dap4/d4ts/build.gradle` work?
````Gradle
configurations.all {
    exclude group: 'edu.ucar', module: 'cdm'
}
````

DennisHeimbigner:
that solution does not work because Nc4IOSP is in this module and it inherently
depends on CDM. So your solution would not compile.
One could of course construct the netcdf4-core module on the fly (maybe)
but that seems overly complicated compared to the split and god knows how Intellij would handle it.

cwardgar:
True, `:netcdf4` needs `:cdm` on the classpath to compile `Nc4Iosp`. However, that `:cdm` exclusion goes in `:dap4:d4ts`, not `:netcdf4`. As a result, there is no compilation failure (I've tested).

The only failure I can foresee is if you attempt to use `Nc4Iosp` from `:dap4:d4ts`; `:cdm` won't be on the runtime classpath. As long as you avoid that, you should be okay.

PR #426, opened 2016-02-15
Cleaner logging in NetcdfFile

msdsoftware:
Altered numerous calls to System.out.println or System.err.println to instead use logging utility.

### DIFF
--- a/cdm/src/main/java/ucar/nc2/NetcdfFile.java
+++ b/cdm/src/main/java/ucar/nc2/NetcdfFile.java
@@ -110,106 +110,106 @@ public class NetcdfFile implements ucar.nc2.util.cache.FileCacheable, Closeable 
     try {
       registerIOProvider("ucar.nc2.stream.NcStreamIosp");
     } catch (Throwable e) {
-      if (loadWarnings) log.info("Cant load class: " + e);
+      if (loadWarnings) log.info("Cant load class NcStreamIosp: {}", e);
     }
     try {
       registerIOProvider("ucar.nc2.iosp.hdf5.H5iosp");
     } catch (Throwable e) {
-      if (loadWarnings) log.info("Cant load class: " + e);
+      if (loadWarnings) log.info("Cant load class H5iosp: {}", e);
     }
     try {
       registerIOProvider("ucar.nc2.iosp.hdf4.H4iosp");
     } catch (Throwable e) {
-      if (loadWarnings) log.info("Cant load class: " + e);
+      if (loadWarnings) log.info("Cant load class H4iosp: {}", e);
     }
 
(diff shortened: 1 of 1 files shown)


cwardgar:
Beautiful! We've been going through this process of converting System.out calls to SLF4J as well, especially in our tests. Thanks, Robert!

msdsoftware:
Had to do some diagnostics on a file over the weekend and ran into this logging consistency. (This is also why I was messing around with figuring out how to use gradle last a few days ago.)

Another thing I ran into is that depending on the class, the logging object seems to get different names. It was "log" in NetcdfFile, but I've seen it as logger and classLogger elsewhere in the library. Is there any general preference? I invariable call it logger or LOGGER in my own code, depending on how the object is created.

Oh, also, I changed all the println's to log.info here based on seeing it done that way previously, although I very reasonably could have gone for log.debug. Decided it didn't matter as the bulk of the calls were within if-then blocks based on debug settings.

dopplershift:
If the debug flags only exist to control the logging, I'd prefer to remove the debug flag and just use `log.debug`, since you can control that without rebuilding code.

cwardgar:
@msdsoftware I prefer the name `logger` in my own code, but it's not something we're fussy about. Much more important is to get all of that spam off STDOUT, whatever the logger is called.

The vast majority of those debug flags are used exclusively for logging and I agree that they should be eliminated and replaced with `logger.debug()` calls instead. If necessary, there can be multiple debug loggers. An example from [FileCache](https://github.com/Unidata/thredds/blob/master/cdm/src/main/java/ucar/nc2/util/cache/FileCache.java#L92):

````java
Logger log = LoggerFactory.getLogger(FileCache.class);
Logger cacheLog = LoggerFactory.getLogger("cacheLogger");
// Debug-only loggers
Logger printLog = LoggerFactory.getLogger(FileCache.class.getName() + "Print");
Logger cleanupLog = LoggerFactory.getLogger(FileCache.class.getName() + "Cleanup");
````
The last three loggers replace the debug flags:
````java
  static boolean debug = false;
  static boolean debugPrint = false;
  static boolean debugCleanup = false;
````
Also, if we need to perform extra steps in the event of a debug scenario, we can use [Logger.isDebugEnabled()](http://www.slf4j.org/api/org/slf4j/Logger.html#isDebugEnabled()):
````java
if (cacheLog.isDebugEnabled()) {
    // Extra steps here.

    cacheLog.debug("Stuff");
}
````
I'll make an issue about this.

msdsoftware:
NetcdfFile has 4-5 separate debug flags controlling which diagnostics the logger would report. It was actually rather nifty when I figured it out and discovered that it was not an all-or-nothing spigot.

PR #432, opened 2016-02-19
HTTPMethod.close not releasing connection

DennisHeimbigner:
Fix Issue 431.
John Caron reports a test program (TestHttpRda.java) that shows that
making 5 method calls works, but the 6th time, it hangs.
Investigation shows:
1. the number 5 is coming from the max connections parameter.
2. no timeout is occurring because CONN_REQ_TIMEOUT is set to infinity.

Further investigation showed that the connection pool is not properly being told to release the connection. This occurred in the switch to apache httpclient 4.5.
Fixes:
1. properly release the connection.
2. allow for setting the connection request timeout via setConnectionTimeout.
3. Disable (for now) setting max connections because I cannot get it to work;
   the httpclient library seems to be ignoring this parameter.
4. add TestHttpRda as part of test set.

### DIFF
--- a/cdm-test/src/test/java/ucar/nc2/util/net/TestHttpRda.java
+++ b/cdm-test/src/test/java/ucar/nc2/util/net/TestHttpRda.java
@@ -1,0 +1,131 @@
+/*
+ * Copyright 1998-2015 John Caron and University Corporation for Atmospheric Research/Unidata
+ *
+ *  Portions of this software were developed by the Unidata Program at the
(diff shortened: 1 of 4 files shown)


lesserwhirls:
@DennisHeimbigner - how much work would it take to get this on 4.6 (still the master branch)?

DennisHeimbigner:
Should of thought of this myself, sorry. Effort shoudl be essentially 
zero. We might be able to cherry pick the change
onto master. Otherwise, I will just do it.
=Dennis

PR #439, opened 2016-02-22
Improve performance of timeseries reading.

joanpau:
The current implementation of timeseries reading in `ThreddsScalarLayer`
queries the dataset once for each involved time instant with
`CdmUtils.readHorizontalPoints`. The original layer implementation in
ncwms uses the function `CdmUtils.readTimeseries`, which retrieves
all the data from the earliest to the latest time instant at once.
With the modifications in socib/ncWMS@timeseries_fast_read_thredds
the loading can be delegated to that function with a significant
improvement in speed and caching opportunities. In addition,
this is the same approach used to implement `readHorizontalPoints`.
Please note that this will only work with the above mentioned changes to ncwms.

See Unidata/ncWMS#18 and Unidata/thredds#439.

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -189,7 +189,7 @@ libraries["commons-compress"] = "org.apache.commons:commons-compress:1.8.1"
 
 libraries["oro"] = "oro:oro:2.0.8"
 
-libraries["ncwms"] = dependencies.create("uk.ac.rdg.resc:ncwms:1.2.tds.4.6.4") {
+libraries["ncwms"] = dependencies.create("uk.ac.rdg.resc:ncwms:1.2.tds.4.6.5-SNAPSHOT") {
     // This is an external dependency on an old artifact of ours.
     // The "cdm" artifact that we build in this project will take its place.
     // This needs to be a local exclusion; globally excluding "edu.ucar:cdm" would be very bad.
--- a/tds/src/main/java/thredds/server/wms/ThreddsScalarLayer.java
+++ b/tds/src/main/java/thredds/server/wms/ThreddsScalarLayer.java
@@ -176,41 +176,19 @@ class ThreddsScalarLayer extends AbstractScalarLayer implements ThreddsLayer {
       throw new IOException(e);
     }
     return horizontalPoints;
-
   }
     
   @Override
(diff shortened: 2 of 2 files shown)


cwardgar:
Hi @joanpau. Thanks for the PR, but this fails to compile on Travis.

dopplershift:
Here's the build error:
```
:tds:compileJava/home/travis/build/Unidata/thredds/tds/src/main/java/thredds/server/wms/ThreddsScalarLayer.java:190: error: method readTimeseries in class CdmUtils cannot be applied to given types;
      return CdmUtils.readTimeseries(null, grid, horizGrid, tIndices, zIndex, xy);
                     ^
  required: NetcdfDataset,String,HorizontalGrid,List<Integer>,int,HorizontalPosition
  found: <null>,GridDatatype,HorizontalGrid,List<Integer>,int,HorizontalPosition
  reason: actual argument GridDatatype cannot be converted to String by method invocation conversion
```

Also, we're looking at moving our master branch over to 5.0, so if we don't get this merged before then, the PR will automatically close and you'll need to resubmit.

joanpau:
As stated in the PR description, as well as in the commit message,
these PR requires the modification to ncWMS in PR Unidata/ncWMS#18 to be merged first.
It is a trivial addition of the missing `readTimeseries` method reported in the compile error,
using the same delegation approach used for  `readHorizontalPoints`.

lesserwhirls:
Ah, I see. Sorry about that @joanpau! I've merged in the PR on the ncWMS side. Would you mind updating this PR with one simple addition? The version of ncWMS will need to be bumped in order to pick up your change.

In line 192 of the file `gradle/dependencies.gradle`, simply change:

```
libraries["ncwms"] = dependencies.create("uk.ac.rdg.resc:ncwms:1.2.tds.4.6.4") {
```

 to 

```
libraries["ncwms"] = dependencies.create("uk.ac.rdg.resc:ncwms:1.2.tds.4.6.5-SNAPSHOT") {
```

Thanks!

joanpau:
I amended the commit to update the version number of the ncWMS dependency, as requested.

cwardgar:
This was rebased onto master and merged in #481.

cwardgar:
Thanks @joanpau!

PR #452, opened 2016-02-27
Part 1 of the dap4 refactor.

DennisHeimbigner:
1. Move all cdm dependent code into dap4/d4cdm
2. Modify various gradle files to support #1.
3. Misc. other mods to support #1.
4. Temporarily breaks d4tswar.


DennisHeimbigner:
Kill until I can figure out why travis failing but jenkins is not

PR #459, opened 2016-03-01
Part 1 of the dap4 refactor.

DennisHeimbigner:
1. Move all cdm dependent code into dap4/d4cdm
2. Move all  code in d4client into dap4/d4shared
   (will probably undo partly in part2 or later)
3. Move all code in d4tswar into dap4/d4ts
4. Misc. other mods to support above
5. Modify various gradle files to support above.


DennisHeimbigner:
Still not right. withdrawn

DennisHeimbigner:
Oops, I did get it right; re-open

dap4/d4shared/build.gradle
```diff
@@ -2,6 +2,9 @@ dependencies {
     compile project(':dap4:d4core')
     compile project(':httpservices')
 
+    compile libraries["jcommander"]
+    compile libraries["slf4j-api"]
+    runtime libraries["slf4j-jdk14"]
```
cwardgar: This causes a problem for TDS, as well as for other applications that use the `:dap4:d4shared` module (directly or transitively):
```
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/Users/cwardgar/dev/projects/thredds2/build/artifacts/tds/exploded/tds-5.0.0-SNAPSHOT.war/WEB-INF/lib/log4j-slf4j-impl-2.4.1.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/Users/cwardgar/dev/projects/thredds2/build/artifacts/tds/exploded/tds-5.0.0-SNAPSHOT.war/WEB-INF/lib/slf4j-jdk14-1.7.13.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [org.apache.logging.slf4j.Log4jLoggerFactory]
```
Best practice is to exclude SLF4J bindings in library modules. They should only appear in applications at the "leaves" of the dependency tree, such as TDS and ToolsUI.

DennisHeimbigner: ok, will fix.

DennisHeimbigner: BTW, what is to be done when I have a module that is both a leaf and will
be used in another module.

cwardgar:
Okay, this looks good.

To answer your question, it's a good idea to have your application program(s) in their own module, and for that module to not be used anywhere else. We do that with `:ui`. It looks like you used to do this too before the reorganization.

On the other hand, we have several [CLI programs](http://www.unidata.ucar.edu/software/thredds/current/netcdf-java/reference/manPages.html) in `:cdm`, and of course `:cdm` doesn't have a SLF4J binding. So, if someone is using those programs via netcdfAll.jar, we instruct them to [add a logger of their choice](https://www.unidata.ucar.edu/software/thredds/current/netcdf-java/reference/BuildDependencies.html#_logging). Up to you what you think is best for DAP4.

PR #463, opened 2016-03-02
[WIP] CDMRemote fixes on 5.0

dopplershift:
This is an on-going collections of things I've found while implementing support for CDMRemote from TDS 5.0 in Siphon. Current list:
- `MAGIC_DATA2` mis-defined in docs
- Vlen dimensions write out `-1` to unsigned length field. Why write a length at all?
- Fix writing and caching of opaque data (Need to ensure cached `ByteBuffer`s aren't exhausted.)

This is not yet complete or ready for merge, just want to get feedback from @JohnLCaron.


dopplershift:
I'm especially concerned about the opaque fix; I'm wondering if we shouldn't be returning something other than the stateful `ByteBuffers` when reading in opaque data--or at least not caching them.

dopplershift:
I read that and came away thinking `duplicate()` is the right solution, because it eliminates touching mutable state. The "copy" is another copy of a `ByteBuffer` object, still referencing the original source data--so no copies of data, just a cheap instance that I think we can live with.

The reason `mark()` works is because that place is the only place in the chain that was actually reading data from the `ByteBuffer`--nowhere else was mucking with `bb`'s internal position.

I would really prefer to treat the `ByteBuffer` as a read-only object and avoid touching any state inside the original.

JohnLCaron:
Using duplicate() is reasonable, as long as it doesnt copy the data itself.
I didnt follow all branches in the JVM to verify that.

Underlying problem is the mutability of ByteBuffer. One could look for an
immutable substitute.
We could use a wrapper class to keep ByteBuffer API from leaking, then redo
when theres a clear alternative.

but probably not worth extensive rewrite at this point.

DennisHeimbigner:
Is this ready for merge yet?

dopplershift:
Should be. I haven't found any more issues and am moving on to CDM Remote Feature.

PR #343, opened 2015-12-30
Ncwms2

lesserwhirls:
See Unidata/thredds#234 for more info.

PR #473, opened 2016-03-08
Nuke ncIdv module and attempt to fix FileCache synchronization

cwardgar:
This attempts to fix the failure reported [here](https://github.com/Unidata/thredds/pull/343#issuecomment-192353389). There were definitely some questionable accesses of shared mutable data that I found, but I can't be sure whether this commit will fix the problem, as I can't reproduce it reliably. I guess we'll see.


cwardgar:
This one is ready to go.

Issue #472, opened 2016-03-08
PointFeatureCollection broken by updating file


dopplershift:
On our systems, the METAR data are written into daily files. After requesting data from today's file, subsequent requests get an error once the file is updated with more data:
```
Illegal Range for dimension 0: last requested 106738 > max 98823
```

This can be demostrated using Siphon:
```python
from datetime import datetime
from siphon.catalog import TDSCatalog
from siphon.ncss import NCSS

cat = TDSCatalog('http://atm.ucar.edu/thredds/catalog/nws/metar/ncdecoded/catalog.xml')
ds = list(cat.datasets.values())[0]
ncss = NCSS(ds.access_urls['NetcdfSubset'])
print(ncss.variables)

query = ncss.query()
query.lonlat_box(east=-80, west=-90, south=25, north=35).accept('netcdf4')
query.variables('air_temperature', 'dew_point_temperature').time(datetime(2016,3,8,12))
nc = ncss.get_data(query)
```

My guess is that something is caching part of the file, but the request for the current length of the "recNum" unlimited dimension goes through to the actual file. We probably never encountered this before since collections were never properly updating (or at least we couldn't query them).

dopplershift:
It seems to fix itself on the hour(?), which is odd, since the `FeatureCollection` tag tells it to check every 15 minutes thredds/atm.

dopplershift:
So it looks like problematic sequence (through **so many** nested calls) is:

1. Load netcdf file header information for feature collection--this includes variables sizes
2. File on disk updates
3. Get request
4. Get station information, which necessitates reading information from disk
5. Station information includes the offsets for this linked-list-in-NetCDF.
6. Offsets from data on disk are greater than our (cached) size of variables
7. ~~Profit!~~ Exception

dopplershift:
Options that I can think of:
- Abandon any support for this linked-list thingy
- Somehow make updating the collection pull in the station information. This would end up joining some code that's pretty far apart as far as I can tell
- Somehow make sure the NetcdfFile instances for the collection are not cached.
- ????
We're in this unique hell because not all of the header/metadata information is read in one "atomic" operation and we're simultaneously having the netcdf files, which are not message-based, updated on disk.

@JohnLCaron any thoughts?

dopplershift:
[...]

PR #479, opened 2016-03-09
AbstractIOServiceProvider.getLastModified() returns correct values when called after close()

cwardgar:
* Get last-modified time from underlying `File`, not the `RandomAccessFile`.
* Clarified `FileCacheable.getLastModified()`: zero is returned when time can't be determined for any reason.
* Reverted code in `FileCache` that considered files with `lastModified == 0` to be unchanged.

This is a better way of fixing the cache misses than my previous solution. I'll merge the commit into 5.0.0 once this PR is accepted.

### DIFF
--- a/cdm/src/main/java/ucar/nc2/iosp/AbstractIOServiceProvider.java
+++ b/cdm/src/main/java/ucar/nc2/iosp/AbstractIOServiceProvider.java
@@ -33,14 +33,18 @@
 
 package ucar.nc2.iosp;
 
-import ucar.ma2.*;
+import ucar.ma2.Array;
+import ucar.ma2.InvalidRangeException;
+import ucar.ma2.Section;
+import ucar.ma2.StructureDataIterator;
+import ucar.nc2.NetcdfFile;
 import ucar.nc2.ParsedSectionSpec;
 import ucar.nc2.Structure;
-import ucar.nc2.NetcdfFile;
 import ucar.nc2.util.CancelTask;
 import ucar.unidata.io.RandomAccessFile;
 import ucar.unidata.util.Format;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.OutputStream;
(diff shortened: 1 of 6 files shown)


DennisHeimbigner:
Christian- How does this play with HttpRandomAccessFile?

cwardgar:
`HTTPRandomAccessFile` inherits from `RandomAccessFile.getLastModified()`:
```java
  @Override
  public long getLastModified() {
    File file = new File(getLocation());
    return file.lastModified();
  }
```
Seems a little suspect. However, this commit doesn't impact `HTTPRandomAccessFile` at all.

DennisHeimbigner:
Looks to me like the HttpRandomAccess.getLastModified is wrong. Not sure how to fix, though.
Perhaps it should be converted to throw UnsupportedOperationException and see what fails.

dopplershift:
Maybe just return 0?

cwardgar:
I suspect that it's always just returning zero, because `new File("http://some.url.com").exists() == false`. `File.getLastModified()` returns zero for non-existent files.

Is there a good way to determine whether remote files have changed?

DennisHeimbigner:
Probably the only way is to have it encoded in an Http header.
There is such a header, but it would require mods to the server code.
14.29 Last-Modified
The Last-Modified entity-header field indicates the date and time at which the origin server believes the variant was last modified.
       Last-Modified  = "Last-Modified" ":" HTTP-date
An example of its use is
       Last-Modified: Tue, 15 Nov 1994 12:45:26 GMT

DennisHeimbigner:
Anyway, lets take the easy way out and have HttpRandomAccessFile.getLastModified
always and specifically return 0.

cwardgar:
Okay, I'll add that to this PR.

cwardgar:
Okay, this is ready to go, unless @JohnLCaron or someone else wants to comment. It looks like it'll fix #472 as well.

Issue #492, opened 2016-03-18
.OIS data not being read properly


alanpaulkwan:
Here is an example dataset.

https://www.dropbox.com/s/ri24oz6h1rxl7on/F12199501290001.OIS?dl=0

Here is a link to the issue: http://stackoverflow.com/questions/35929150/using-ucar-iosp-to-read-a-dataset/36069839#36069839

It complains when I run the following line that it was not a valid CDM file, even though one of the hallmark examples of this IOSP is its use on the .OIS data. I happen to need the .OIS data so would really appreciate if any suggestions for fixing it! 

NetcdfFile ncfile1 = NetcdfFile.open(s);

dopplershift:
I poked at the file, and yours seems to have (at least) one header line that the netCDF-java code isn't expecting and doesn't ignore. I'll take a look when I get a chance to see what happens if I try adding the line. If you're game for trying to run this down, the file `DMSPHeader.java` is what you're looking for.

alanpaulkwan:
Thanks! I have no real idea how to diagnose Java - haven't used it for 10
years.

I'll make an independent effort, but I imagine even if I figured out which
line to delete is causing the problem I wouldn't know how to fix this.

dopplershift:
#493 has code that at least lets the file be opened without errors. No idea if there are any other lurking problems.

PR #493, opened 2016-03-18
Fix DMSP files

dopplershift:
Gotten at least one new file in #492. In lieu of having an actual format document, I've just made the IOSP a bit more permissive. I need to add this file to our unit test directory.


alanpaulkwan:
The dda.x file has the schema, but that format is like from the early 1990s, my hipster CS friend tells me. I'm not sure what to make of it. 

Is there an easy way to get the compiled .jar?

lesserwhirls:
Thanks @dopplershift!

@alanpaulkwan - here is a snapshot jar that contains the fix:

https://artifacts.unidata.ucar.edu/content/repositories/unidata-snapshots/edu/ucar/netcdfAll/4.6.5-SNAPSHOT/netcdfAll-4.6.5-20160319.183320-4.jar

We will be having an official release of v4.6.5 next week, but this should get you up and running for now.

PR #502, opened 2016-03-24
Purge a bunch of stuff.

dopplershift:
The 5.x version of this stuff is still available from "theAttic" tag.


cwardgar:
So this just nukes everything in `archive/src/docs/web/`?

dopplershift:
This nukes:
- `archive/`
- `wmotables/`
- `cdmvalidator/`

PR #514, opened 2016-04-04
Release v4.6.5

cwardgar:
Begin work on 4.6.6-SNAPSHOT

I created a tag for this and pushed it upstream to cwardgar/thredds, but I'm not sure if it'll get included in this PR. Might have to create another tag on Unidata/thredds.


dopplershift:
Pretty sure tags don't ever get moved from PR's--so you'll need to push to unidata. Somebody with admin rights *ahem* @lesserwhirls *ahem* can also make a new tag/release.

dopplershift:
I stand corrected: any of us can make a new tag/release. (Which makes more sense, actually.)

PR #521, opened 2016-04-08
Revert "Link to artifacts on FTP server, not Nexus"

lesserwhirls:
Reverts Unidata/thredds#519


lesserwhirls:
Oops...I didn't mean to merge that one.

Yes, if we leave links to the FTP server, we don't need to change them. But, we moved to artifacts as a one stop shop for all downloads.

cwardgar:
I don't really care, except that it adds another step to 4.6.x releases--a painful one because there ALWAYS seem to be permission issues when updating the website.

You shouldn't revert my changes to release.txt though.

lesserwhirls:
Dang, good point. The permissions stuff sucks bad. I'll kill this one.

PR #529, opened 2016-04-13
Add -PpublishSources flag to publish sources jars that are otherwise …

bencaradocdavies:
…not published for SNAPSHOTS

Sources jars enable cross-project debugging of SNAPSHOTS. For example, local modifications to NetCDF-Java master are published with ```-PpublishSources``` as ```4.6.6-SNAPSHOT```, then when debugging GeoServer built with ```-Dnetcdf.version=4.6.6-SNAPSHOT``` in Eclipse, attached sources of NetCDF-Java are available.

I am happy to port to ```5.0.0``` if you like.


cwardgar:
Hi Ben,

There was a disk space issue on our Nexus server that prevented us from publishing these jars for all SNAPSHOTS, but that's been resolved. So, I see no reason not to just publish the source and javadoc jars for all releases. I went ahead and made those changes in another PR: #532, which includes your commit. Thanks for the contribution!

PR #530, opened 2016-04-13
Fallback to FSL/NCEP local table superclasses for GRIB2 level and sta…

bencaradocdavies:
…tistic names

[NOAA Rapid Refresh (RAPv3)](http://rapidrefresh.noaa.gov/) native grid GRIB2 files contain names that are defined in FSL/NCEP local table superclasses yet are inexplicably overridden to return null. While this works for the first null, subsequent nulls cause duplicate NetCDF variable names and failure to read the GRIB2 file. This fix enables fallback to definitions in superclasses.

I am happy to port this to 5.0.0 as well.

### DIFF
--- a/grib/src/main/java/ucar/nc2/grib/grib2/table/FslLocalTables.java
+++ b/grib/src/main/java/ucar/nc2/grib/grib2/table/FslLocalTables.java
@@ -139,24 +139,22 @@ public class FslLocalTables extends NcepLocalTables {
 
   @Override
   public String getLevelNameShort(int id) {
-    if (id < 192) return super.getLevelNameShort(id);
     switch (id) {
       case 200:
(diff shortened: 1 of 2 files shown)


cwardgar:
This seems reasonable; those methods simply return `"UnknownLevelType-" + id` and `"UnknownStatType-" + id`, in instances where they used to return `null`. Also, [Jenkins test run](https://jenkins-aws.unidata.ucar.edu/job/cwardgar/17/testReport/) looks good. Any issue that you can see, @lesserwhirls?

lesserwhirls:
Sounds good to me - thank you @bencaradocdavies! I'm not sure how easily this translates to the 5.0 branch, but if you would not mind making another PR for that, that would be great!

PR #532, opened 2016-04-14
Publish sources and javadoc jars for all releases

cwardgar:
There was a disk space issue on our Nexus server that prevented us from publishing these jars for all SNAPSHOTS, but that's been resolved.

EDIT: This follows from PR #529, opened by @bencaradocdavies.


bencaradocdavies:
@cwardgar this looks good to me. It is a simpler solution.

lesserwhirls:
rebase bump @cwardgar...I wish there was a good way around this.

dopplershift:
Only way around it is to let the admin (you) override for things you're not worried about breaking on latest master.

lesserwhirls:
Ah...yeah, for this change to gradle for publish src jars, I'm not worried.

lesserwhirls:
But, to be consistant, I suppose it's best just to jump through the hoops for now.

PR #533, opened 2016-04-14
Fixed multiple issues regarding very large, filtered HDF5 data chunks

cwardgar:
* Ensure that no more than MAX_ARRAY_LEN bytes are allocated for decompressing a chunk. Avoids "Negative initial size" overflow error and fixes [KXL-349288](https://andy.unidata.ucar.edu/esupport/staff/index.php?_m=tickets&_a=viewticket&ticketid=26872).
* Detect when chunk is too big for Java (but valid according to HDF5 spec) and throw an exception with a useful message.
* Catch OutOfMemoryErrors resulting from large chunks and rethrow with a useful message.


lesserwhirls:
Rebase bump @cwardgar

PR #543, opened 2016-05-03
Moved testing utility classes out of :cdm and into their own subproject

cwardgar:
* As a result, :cdm is no longer producing a testArtifacts jar. All of the classes in its "test" SourceSet are actually tests now!
* The former setup was a holdover from the Maven build. It was confusing and error-prone.
* All testing utility classes have been moved to the 'ucar.unidata.util.test' package. Import statements in test classes were updated to match.
* Nuked TestDiskCacheUtils: it wasn't being used.

This change was spurred by IntelliJ IDEA 2016.1.1, which can't make sense of the testArtifacts jar that we were producing.
Sure, IntelliJ is technically wrong because Gradle could still build the project from the command line, but that doesn't mean that our old setup wasn't stupid.


dopplershift:
Fixes it for me here as well. Thanks @cwardgar !

Do you need to do a separate PR for 5.x?

cwardgar:
I think it'd be easier to just merge master into 5.0.0 and fix what breaks. There are new classes in 5.0.0 that don't refer to the new package, so there'll definitely be breakage.

PR #547, opened 2016-05-08
Return 403 for not authorized dataset.

cofinoa:
For not authorized datasets TDS authorizer should return 403 (Forbidden) instead 401 (Unauthorized). 401 should be sent for unauthenticated user

### DIFF
--- a/tds/src/main/java/thredds/servlet/restrict/TomcatAuthorizer.java
+++ b/tds/src/main/java/thredds/servlet/restrict/TomcatAuthorizer.java
@@ -116,7 +116,7 @@ public class TomcatAuthorizer implements Authorizer {
       }
     }
 
-    res.sendError(HttpServletResponse.SC_UNAUTHORIZED, "Not authorized to access this dataset.");
+    res.sendError(HttpServletResponse.SC_FORBIDDEN, "Not authorized to access this dataset.");
   }
 
 }


dopplershift:
@DennisHeimbigner any thoughts?

DennisHeimbigner:
Looking at this:
https://en.wikipedia.org/wiki/HTTP_403#Difference_from_status_.22401_Unauthorized.22
I can see a difference, but I do not understand why you think this difference is relevant to thredds.
As far as I know, we always require authentication by a user to access data sets, so either they
provide the proper info or not. So 401 would seem correct.
Please elaborate by giving e.g. examples or a more detailed explanation.

lesserwhirls:
Does this distinguish the case when a user can authenticate successfully
but does not have acces to a particular protected dataset?

DennisHeimbigner:
As far as I know no such cases exist. Not clear it we even support such a case.

lesserwhirls:
I thought we could support users being authenticated for some
protected datasets but not other protected datasets.

DennisHeimbigner:
Yes, but as far as I know, they must provide credentials to access a given dataset.
Can you outline a sequence of events that would access a specific dataset without
being asked to provide credentials?

DennisHeimbigner:
I did some more searching on this.
I actually have no particular objection to making this change; we might
have to change some test that assume a 401.
Note also that this applies to Spring authentication, not to apache level or tomcat
level auth (despite the name TomcatAutorizor) that we support.
So depending on how the thredds server
is set up, the user might sometimes see 401 and sometimes see 403.

I found this rather concise characterization (http://stackoverflow.com/questions/3297048/403-forbidden-vs-401-unauthorized-http-responses)
> In summary, a 401 Unauthorized response should be used for missing or bad authentication,
> and a 403 Forbidden response should be used afterwards, when the user is authenticated but
> isn’t authorized to perform the requested operation on the given resource.
So, I am trying to figure out how (with thredds as it stands) this second case can occur.

cofinoa:
This applies when TDS is restricting access to datasets
http://www.unidata.ucar.edu/software/thredds/v4.6/tds/reference/RestrictedAccess.html

It could happen that the user has put the correct credentials, but he doesn't has the required role to access dataset.

DennisHeimbigner:
I am putting this PR on hold until such time as I can see if the proposed example
can actually occur. In any case, 401 is the proper response in 99% of cases, so 
this change needs to be role testing specific, not general.

lesserwhirls:
@ethanrd - any comments on this PR?

DennisHeimbigner:
There are two issues here, I think.
One is the generating 403's instead of 401.
I currently do not believe we should do that.
There are possible failure cases where we should generate 403,
but frankly, I do not think that finding these is a high priority.
There is a separate issue about DODS url rewriting, but I have not
investigated this in detail.

cofinoa:
@DennisHeimbigner, 
The authentication challenge it's already solved before arriving to this point. 

That means that user is not authorized (403). This the change a propose. 

But you mean we need to do a more defensive approach? 

It may this code would be more robust:
```
if (req.getUserPrincipal() != null){
  res.sendError(HttpServletResponse.SC_FORBIDDEN, "Not authorized to access this dataset.");
}else{
  res.sendError(HttpServletResponse.SC_UNAUTHORIZED, "Not authenticated to access this dataset.");
}
```

see: https://tomcat.apache.org/tomcat-7.0-doc/servletapi/javax/servlet/http/HttpServletRequest.html#getUserPrincipal%28%29

ethanrd:
The HTTP 1.1 description of a 401 includes: "If the request already included Authorization credentials, then the 401 response indicates that authorization has been refused for those credentials".

I would interpret that as meaning that a 401 Unauthorized response is appropriate when a user has authenticated but their credentials do not have permission to access the requested URL (e.g., the user does not have the appropriate role). That goes against the wikipedia article @DennisHeimbigner referenced above but, I think, goes along with (at least some of) the stackoverflow Q&A also referenced above.

On the other hand, a 403 Forbidden response does not see inappropriate. If for instance the server for some reason doesn't want the user to try again with other credentials.

I would lean towards a 401 in both the unauthenticated case and the authenticated but not authorized case. That would allow client software to allow users to try a different set of credentials in case they have separate credentials for different roles.

cofinoa:
This pull request was made because with the current approach, if a client receive a 401 response there is no way to know if this 401 code is because the credentials provided by the client are OK (user authenticated) or user has provided wrong credentials (user wrong authenticated). 

It happen to me and this was the initial reason because I have proposed this PR. More over, currently I'm solving this intercepting the 401 response code in the error page configuration and checking if the user has authenticated correctly, if so, then the tomcat response is changed to 403. 

This hack raised the issue #551, with the client algorithm written in the netcdf-java to disambiguate dods and http URLs end-points

DennisHeimbigner:
super-ceded by pr 754

PR #548, opened 2016-05-09
Build failed when executing gradle 2.13 (latest release)

vegasm:
Incremented version of the gradle shadow plugin to 1.2.3, which contains a fix for compability with gradle 2.11 and higher

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -38,7 +38,7 @@ ext {
 
 libraries["gretty"] = "org.akhikhl.gretty:gretty:1.2.0"
 
-libraries["shadow"] = "com.github.jengelman.gradle.plugins:shadow:1.2.2"
+libraries["shadow"] = "com.github.jengelman.gradle.plugins:shadow:1.2.3"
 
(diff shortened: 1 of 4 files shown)


cwardgar:
I'm curious: why aren't you using the [wrapper](https://github.com/Unidata/thredds/blob/5.0.0/gradlew) to build? It's meant to prevent precisely these kind of build failures, where developers are using different system-installed versions of Gradle.

vegasm:
I was not able to import the project in IntelliJ using the latest gradle version. After fixing that, everything worked fine. Is there any option to use the wrapper in Intellij? I was using eclipse and it is new for me.

cwardgar:
Yes, there's a way to use the wrapper in IntelliJ: 
screen shot 2016-05-09 at 3 13 03 pm

I don't know for sure about Eclipse, but I'd be surprised if it weren't possible there as well.

cwardgar:
This is a non-issue when using the wrapper, so I'm going to go ahead and close it. The shadow plugin version will get incremented when we update to a newer Gradle. I think I've already done that in a branch, actually.

PR #550, opened 2016-05-10
BUFR updates

dopplershift:
- Update local tables for EUMETSAT (esupport RTS-672326) from ECMWF BufrDC
- Which necessitates fixing parsing of ECMWF tables--the new sequence table has 100 entries, which runs together with first number in the line, breaking parsing using whitespace
- Go ahead and update WMO tables


cwardgar:
Did you get the impression that there are a lot of old tables and other cruft in bufr that we don't need anymore? For example, do we still need `BUFRCREX_22_0_1_CodeFlag_en.xml` now that we have `BUFRCREX_25_0_0_CodeFlag_en.xml`?

There are also a lot of binary files in `bufr/src/main/sources` that are clogging the repo. I suppose we can put those in git-lfs eventually, but I suspect that it's not necessary to keep most of them around anyway.

dopplershift:
The binary files we can probably do without. If we agree, I'll remove the zip file of the WMO tables I added here (there's nothing we can do about the old ones from an overall repository size).

As far as removing the old tables, I'm not sure. For the local tables, I'd be hesitant, just because I'm not sure what policy places follow on keeping backwards compatibility. We might be able to reduce the WMO tables.

cwardgar:
Okay, we concluded in the meeting today that we have to keep the old tables.

The zip file is less than 1 MB, so I'm not particularly irked by it. Once we have git-lfs working though, I'd like to move these kind of files to it. As for decreasing the overall repository size, we could try [BFG Repo-Cleaner](https://rtyley.github.io/bfg-repo-cleaner/).

PR #553, opened 2016-05-20
godiva2: Replace broken WMS background layers

cskarby:
Some WMS background layers were not responding. I have replaced some and
disabled other layers. All the replacement services reports no fees and no
restrictions via their GetCapabilites responses as of 2016-05-20.

Replaced:
  * http://wms-basemaps.appspot.com (seems to be out of service - http/404)
    * Blue Marble + bathymetry - replaced with http://neowms.sci.gsfc.nasa.gov/wms/wms
    * North polar stereographic - falling back to http://nsidc.org/cgi-bin/atlas_north which godiva2 used earlier
    * South polar stereographic - falling back to http://nsidc.org/cgi-bin/atlas_south which godiva2 used earlier

Disabled:
  * http://iceds.ge.ucl.ac.uk/cgi-bin/icedswms (seems to be out of service - unknown host)
    * SRTM DEM - disabled background layer
  * http://labs.metacarta.com/wms-c/Basic.py (suspected to be out of service - requests times out)
    * OpenLayers WMS - disabled background layer
    * Human Footprint - disabled background layer


cwardgar:
@lesserwhirls What do you think of this?

lesserwhirls:
I think this is a very reasonable thing to do, and good cleanup in general. I'm not sure if the folks at Reading would like us to push this back upstream or not (if they are fully on edal-java, then probably not). Thanks for the PR @cskarby!

Issue #518, opened 2016-04-06
HDF5 DataBTree Issue


dblodgett-usgs:
This issue supercedes #501. 

I have found that I can reproduce the stack trace below with a reproducible workflow.

1. Request data from an ncml union aggregation [1] of OPeNDAP source data and a local netcdf file.
2. Request data from a data source that has 2-D coordinate axes. [2], [3].

After some period of time (tens of minutes), this issue goes away for 2-d coordinate axis datasets that _have not_ been accessed producing the error. If the offending aggregation is accessed then the working dataset is accessed, it brakes like the others. That is, the databtree error is being cached in some way.

[1] http://cida-test.er.usgs.gov/thredds/dodsC/stageiv/ncml/stageivRT.ncml.html
    Grib collection: http://cida-test.er.usgs.gov/thredds/catalog/stageiv/01hRT/catalog.html
    .ncml union can be downloaded here: http://cida-test.er.usgs.gov/thredds/catalog/stageiv/ncml/catalog.html
[2] http://cida-test.er.usgs.gov/thredds/dodsC/UpperDeschutes.html
    If the data behind this is needed, it is visible here: http://cida-test.er.usgs.gov/thredds/catalog/stageiv/symlinked_files/deschutes/catalog.html
[3] http://cida-test.er.usgs.gov/thredds/dodsC/stageiv/symlinked_files/stageiv_archive.nc.html
    This is a single NetCDF4 File here: http://cida-test.er.usgs.gov/thredds/catalog/stageiv/symlinked_files/catalog.html?dataset=examples/symlinked_files/stageiv_archive.nc

```
2016-04-06T13:35:44.315 -0500 [    408330][      44] ERROR - thredds.server.opendap.OpendapServlet - path2= /ncep_stageiv.ascii
java.lang.NullPointerException
    at ucar.unidata.io.RandomAccessFile.read_(RandomAccessFile.java:760) ~[cdm-4.6.5.jar:4.6.5]
    at ucar.unidata.io.RandomAccessFile.readBuffer(RandomAccessFile.java:520) ~[cdm-4.6.5.jar:4.6.5]
    at ucar.unidata.io.RandomAccessFile.seek(RandomAccessFile.java:508) ~[cdm-4.6.5.jar:4.6.5]
    at ucar.nc2.iosp.hdf5.DataBTree$Node.<init>(DataBTree.java:170) ~[cdm-4.6.5.jar:4.6.5]
[...]

PR #558, opened 2016-06-08
Added "Netcdf4Clibrary.useForReading" property to threddsConfig.xml

cwardgar:
* Also added it to nj22Config.xml.
* Added HTML documentation for the property.
* Added property to the default threddsConfig.xml.
* Improved several log messages.
* Nuked threddsConfig.xml.txt. We now just link to the default threddsConfig.xml on GitHub.


cwardgar:
This enables a workaround for the problem identified in #518. I will follow with a PR on 5.0.0 that has AsciiDoc changes that mirror the HTML changes I made in this.

dblodgett-usgs:
👍 If there's a built war some place I can use to test, I'm ready to try it out on our production-scale data.

lesserwhirls:
Thanks @cwardgar!

@dblodgett-usgs - you can grab the latest snapshot war here:

https://artifacts.unidata.ucar.edu/content/repositories/unidata-snapshots/edu/ucar/tds/4.6.6-SNAPSHOT/tds-4.6.6-20160608.175934-6.war

Just remember to rename is to thredds##4.6.6-SNAPSHOT.war before deploying.

dblodgett-usgs:
Maybe I'm doing something wrong. Is that context path critical? I'm getting a lot of ` java.io.IOException: java.io.IOException: 2: No such file or directory ` then way down the stack...

```
Caused by: java.io.IOException: 2: No such file or directory
	at ucar.nc2.jni.netcdf.Nc4Iosp._open(Nc4Iosp.java:317)
```
For any ncml aggregation I try that uses netcdf4 files.

cwardgar:
Strange, I cannot reproduce the problem. `barf.sh` from #518 even works for me.

I wonder if netcdf is installed properly. Do the command-line tools such as `ncdump` work for you? Did you build it from source or install it via a package manager? @lesserwhirls Didn't we decide that the netcdf libs that are available from package managers are unsuitable for NetCDF-Java and that it's necessary to build from source? I forget why.

lesserwhirls:
You need to build a threads safe version of HDF5, and make sure netCDF-C
links to that.

For more info, see the section "A note on HDF5" here:

http://www.unidata.ucar.edu/software/thredds/current/netcdf-java/reference/netcdf4Clibrary.html

dblodgett-usgs:
I'm checking with our SA who built it originally. I would bet it's not thread safe. It does appear to work with single files at a time.

dblodgett-usgs:
We got it built with threads-safe enabled and I still have the issue. 

Turns out it is isolated to aggregations that declare the members of the aggregation by path rather than using a scan. We use that pattern a lot. 

ie.
This does not work:
```
<netcdf xmlns="http://www.unidata.ucar.edu/namespaces/netcdf/ncml-2.2">
        <aggregation type="joinExisting" dimName="time">
                <netcdf location="prism_1890-1899.nc" />
                <netcdf location="prism_1900-1909.nc" />
                <netcdf location="prism_1910-1919.nc" />
...
        </aggregation>
</netcdf>
```
this does:
```
<netcdf xmlns="http://www.unidata.ucar.edu/namespaces/netcdf/ncml-2.2">
        <aggregation type="joinExisting" dimName="time">
                <scan location="./" suffix=".nc" subdirs="false"/>
        </aggregation>
</netcdf>
```

I'd really rather not convert everything to scans...

cwardgar:
Okay, I think I have an idea about this. It's probably some stupid issue wrt passing relative paths to the C library. I think I can fix it by just passing absolute paths.

cwardgar:
Fixed in #560. [Here](http://artifacts.unidata.ucar.edu/content/repositories/unidata-snapshots/edu/ucar/tds/4.6.6-SNAPSHOT/tds-4.6.6-20160609.222655-8.war) is a new build of TDS.

dblodgett-usgs:
That appears to have done it.

cwardgar:
Awesome. Any hiccups at all with the latest build?

dblodgett-usgs:
Haven't had enough chance to push on it. Aggregations seem slow to unresponsive for really big ones, but that could be something else... Need to investigate more. 

On June 10, 2016 1:41:09 PM EDT, Christian W <notifications@github.com> wrote:
>Awesome. Any hiccups at all with the latest build?
>
>---
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/Unidata/thredds/pull/558#issuecomment-225247779

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

dblodgett-usgs:
I don't see in the docs, does THREDDS use of NetCDF4 require the parallel option to be turned on?  

I'm tracking down two issues... one where the server completely hangs on normal catalog.html type requests while under load, and another where some aggregations are producing a null pointer that is swallowed so the in-flight request just never comes back. Will do some more looking and get back to you.

cwardgar:
No, THREDDS does not require that libnetcdf be compiled with Parallel I/O Support. I'm actually not sure whether THREDDS would even benefit from that, but I know we don't use it on our production machines.

It looks like you have some remaining issues to troubleshoot, but they seem unrelated to `Netcdf4Clibrary.useForReading`. So, I'm gonna go ahead and cut a v4.6.6 release that includes that feature tomorrow.

dblodgett-usgs:
OK, yeah. I think the issues I'm seeing may be related to a difference in performance between the C library and the native HDF library. I'm continueing to bang on it. 

I'm getting really poor aggregation performance, and the caches don't seem to be getting used. It's rescanning big aggregations every time I come back to them. 

I think I'll just have to rewrite these data into monolothic joined-together files. If it turns out that performance of the C approach is worse, I don't think we'll want to use it, unfortunately. Will keep you posted with findings.

PR #559, opened 2016-06-09
Synchronized Nc4prototypes

DennisHeimbigner:
Modified Nc4Iosp:
1. jna provides a means to force C calls to be synchronized.
   I added this for the Nc4prototypes interface.
   There may be a (small, I hop) performance hit for the calls,
   but improved safety in multi-threaded environments.
2. Remove the default list of places to look for the libraries.
   Now this must be specified either by -Djna.library.path
   or the corresponding env variable or they must be on the classpath.


cwardgar:
Why not delete those lines instead of commenting them out? Commented code is a code smell. I mean, I know we (mostly John) do this already, but I can't think of a time that commented code was actually useful. Plus we have version control.

DennisHeimbigner:
Because I forgot. I left them in in case it turned out that either jenkins
or travis depended on them in which case, I would have tried something else.
Now that I know that jenkins and travis seem to work ok, I can elide them.

lesserwhirls:
@DennisHeimbigner - would you mind updating the doc for loading the c lib?

docs/website/netcdf-java/reference/netcdf4Clibrary.adoc

PR #578, opened 2016-06-28
-DNetcdf4Clibrary.useForReading command line flag.

DennisHeimbigner:
Added support for a -DNetcdf4Clibrary.useForReading
command line property. This allows for testing without
having to modify other config files.

### DIFF
--- a/cdm/src/main/java/ucar/nc2/NetcdfFile.java
+++ b/cdm/src/main/java/ucar/nc2/NetcdfFile.java
@@ -114,6 +114,8 @@ public class NetcdfFile implements ucar.nc2.util.cache.FileCacheable, Closeable 
     }
     try {
       registerIOProvider("ucar.nc2.iosp.hdf5.H5iosp");
+      if(System.getProperty("Netcdf4Clibrary.useForReading") != null)
+	registerIOProvider("ucar.nc2.jni.netcdf.Nc4Iosp"); // Make sure it precedes hdf5 IOSP
     } catch (Throwable e) {
       if (loadWarnings) log.info("Cant load class H5iosp: {}", e);
     }
--- a/cdm/src/main/java/ucar/nc2/util/xml/RuntimeConfigParser.java
+++ b/cdm/src/main/java/ucar/nc2/util/xml/RuntimeConfigParser.java
@@ -291,6 +291,8 @@ public class RuntimeConfigParser {
         }
 
         boolean useForReading = Boolean.parseBoolean(elem.getChildText("useForReading"));
+	if(!useForReading && System.getProperty("Netcdf4Clibrary.useForReading") != null)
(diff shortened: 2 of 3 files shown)


cwardgar:
I think there's a better place for this addition: in `NetcdfFile.<clinit>`, right before we load `H5iosp`.

https://github.com/Unidata/thredds/blob/master/cdm/src/main/java/ucar/nc2/NetcdfFile.java#L115

DennisHeimbigner:
Ok, but is there any problem is the Nc4Iosp is registered twice?

cwardgar:
No, `NetcdfFile` will just attempt to open files twice using `Nc4Iosp`. Of course, for that to happen, a user must enable `Netcdf4Clibrary.useForReading` both on the command-line and in some config file. Seems unlikely.

Still, you can add a flag somewhere and some logic that checks for that condition if you want.

cdm/src/main/java/ucar/nc2/NetcdfFile.java
```diff
@@ -114,6 +114,8 @@
     }
     try {
       registerIOProvider("ucar.nc2.iosp.hdf5.H5iosp");
+      if(System.getProperty("Netcdf4Clibrary.useForReading") != null)
+	registerIOProvider("ucar.nc2.jni.netcdf.Nc4Iosp"); // Make sure it precedes hdf5 IOSP
```
cwardgar: A couple notes:
* For parity with the rest of the IOSP registrations, this should be in its own try-catch block, not `H5iosp`'s.
* If you follow the logic, this will register `Nc4Iosp` *after* `N5iosp`. I understand why you were confused: we eventually delegate to [registerIOProvider(Class iospClass, boolean last)](https://github.com/Unidata/thredds/blob/v4.6.6/cdm/src/main/java/ucar/nc2/NetcdfFile.java#L307) with `last == false`, but pay attention to the `userLoads` flag.

cwardgar:
In addition to my inline note, this PR needlessly modifies every line in `RuntimeConfigParser` and `CdmInit`. This is gonna muddy the history in git. Any time we try to find the last time that a line in one of those files was modified, we'll get f8c80dd179fef997bc126985efb490681dd66a07, even though it seems to only be a whitespace diff.

Maybe you'll need to submit a new PR to correct the problem? I'm not sure if there's any trick in git for that. Also, you may want to investigate why your text editor did something weird there.

DennisHeimbigner:
That is not how I read it. The registerIosp with last == false uses add(0,...)
which inserts at the front, and the searcher codes startsing looking for a match
at position 0. But, if you think otherwise, then change it and see if it works.

DennisHeimbigner:
Never mind,  I see that userloads doesnt get set to true at the point I expected.

PR #585, opened 2016-07-10
Compile-time deps have 'compile' scope in generated Maven POMs, not 'runtime'

cwardgar:
This addresses a known bug in the maven-publish plugin:
https://discuss.gradle.org/t/maven-publish-plugin-generated-pom-making-dependency-scope-runtime/

I easily spent 90% of my time devoted to this commit on figuring out how to test it. I needed the output of the `generatePomFileFor${name}Publication` task, which required a full, nested Gradle build within the test. Luckily, this is possible with GradleRunner, but I needed to make some changes in order to use it:

* Upgraded our Gradle version from 2.9 to 2.14 (latest) so that I could get the java-gradle-plugin.
* Updated Gretty to a forked version of the original, 1.3.0. The old Gretty was incompatible with Gradle 2.14.
  - The new Gretty also fixes a bug where 'slf4j-nop' was being added to TDS's runtime classpath even though 'log4j-slf4j-impl' was present. We no longer need to explicitly exclude that dep now.
* Updated the shadow plugin to 1.2.3, again for compatibility with Gradle 2.14

GradleRunner opens up a lot of opportunities to improve testing of our build code in the future.

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -1,4 +1,5 @@
 apply plugin: 'groovy'
+apply plugin: 'java-gradle-plugin'  // Adds gradleApi() compile dep and gradleTestKit() testCompile dep.
 
 // We'd normally apply the dependencies with something like:
 //     apply from: "$rootDir/gradle/dependencies.gradle"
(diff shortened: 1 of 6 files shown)


lesserwhirls:
Hi @cwardgar - thanks for all of these beautiful PRs! I've been in the land of no internet for the past week, but I'm back now. Let's start with this one and then work our way through to get these merged in. Would you mind rebasing?

cwardgar:
Done. Just waiting for Travis now.

lesserwhirls:
Thanks @cwardgar!

Issue #581, opened 2016-07-05
Array support for Unlimited dimension


LupusUmbrae:
Hello,

I'd like to reduce the number of calls i have to [NetcdfFileWriter.write()](https://github.com/Unidata/thredds/blob/5.0.0/cdm/src/main/java/ucar/nc2/NetcdfFileWriter.java#L1038) as i'm seeing the [Nc4Iosp.writeData()](https://github.com/Unidata/thredds/blob/5.0.0/netcdf4/src/main/java/ucar/nc2/jni/netcdf/Nc4Iosp.java#L2789) taking a lot of my CPU time. It's averaging about >3000 nan-seconds, which isn't bad by itself but it looks like i'll have hundreds of thousands/millions of calls to writing.

I started looking at buffering on the Java side, by collecting all the data and converting them into an [Array](https://github.com/Unidata/thredds/blob/5.0.0/cdm/src/main/java/ucar/ma2/Array.java), however i hit a problem when i had variables with an Unlimited dimension, i could not see any way to create an Array that would support that.

Have i missed something in Array, or it's subclasses that allows me to store multiple rows of data for a variable that has an unlimited dimension (example NcML below).

```
 <dimension name="rows" length="45000"/>
 <dimension name="orphan_pixels" length="1" isUnlimited="true" />

<variable name="x_orphan_i${v}" shape="rows orphan_pixels" type="int">
    <attribute name="long_name" value="Geolocated x (across track) coordinate of detector FOV centre" />
    <attribute name="units" value="m" />
    <attribute name="scale_factor" type="double" value="0.01" />
    <attribute name="add_offset" type="double" value="0" />
    <attribute name="_FillValue" type="int" value="-1000000" />
  </variable>
```

Any suggestion of how to go forward would be appreciated,
Robin

cwardgar:
Hi Robin,

I took your NcML sample and tried to write it out as a NetCDF-4 file using ToolsUI. I got an error:
```
Exception in thread "AWT-EventQueue-0" java.lang.AssertionError
	at ucar.nc2.jni.netcdf.Nc4Iosp.writeData(Nc4Iosp.java:2852)
	at ucar.nc2.jni.netcdf.Nc4Iosp.writeData(Nc4Iosp.java:2789)
	at ucar.nc2.NetcdfFileWriter.write(NetcdfFileWriter.java:954)
	at ucar.nc2.NetcdfFileWriter.write(NetcdfFileWriter.java:938)
[...]

PR #587, opened 2016-07-11
Array.get1DJavaArray(Class) returns correct value for constant arrays

cwardgar:
* Fixes bug in https://github.com/Unidata/thredds/issues/581


lesserwhirls:
Ok @cwardgar - now let's do this one.

cwardgar:
Rebased.

PR #589, opened 2016-07-12
Add test for writing variable with multiple unlimited dims

cwardgar:
* Inspired by https://github.com/Unidata/thredds/issues/581. Turns out we didn't have a good test of this feature yet.


cwardgar:
I've determined that this test fails on netcdf-4.3.3.1 (what Travis is using) but succeeds on netcdf-4.4.1, which explains why I didn't catch it while testing locally. I'll update this PR to include new netcdf binaries for Travis.

cwardgar:
`TestNc4IospWriting.expandUnlimitedDimensions()` currently [fails on Jenkins](https://jenkins-aws.unidata.ucar.edu/job/cwardgar/24/testReport/junit/ucar.nc2.jni.netcdf/TestNc4IospWriting/expandUnlimitedDimensions/) because it requires netcdf-c 4.4.0+ and we've got an older version on there.

@dopplershift I logged into jenkins-aws but could find no trace of libnetcdf (test output claimed that it was in `/opt/netcdf/lib`). I assume that the shared library is in the snapshot, along with cdmUnitTest/?

lesserwhirls:
Next! Almost there!

cwardgar:
Rebased. Now we're cooking. I'll see if I can upgrade libnetcdf on Jenkins tonight so that the test will pass there as well.

lesserwhirls:
w00t!

PR #592, opened 2016-07-17
Remove unused non-API import

bencaradocdavies:
This unused import is detected by Eclipse Neon as being a non-API type (```sun.*``` classes are not part of the Java API). Eclipse reports:
```
Access restriction: The type 'Debug' is not API (restriction on required library '/usr/lib/jvm/java-8-openjdk-amd64/jre/lib/jsse.jar')
```

The import is unused and can be removed.

Workaround for Eclipse Neon users is to change Window / Preferences / Java / Compiler / Errors/Warnings / Deprecated and restricted API / Forbidden reference (access rule): from Error to Warning.

### DIFF
--- a/grib/src/main/java/ucar/nc2/grib/collection/GribCollectionBuilderFromIndex.java
+++ b/grib/src/main/java/ucar/nc2/grib/collection/GribCollectionBuilderFromIndex.java
@@ -36,7 +36,6 @@
 package ucar.nc2.grib.collection;
 
 import com.google.protobuf.ExtensionRegistry;
-import sun.security.ssl.Debug;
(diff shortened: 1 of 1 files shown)


cwardgar:
Weird, we weren't even using it.

cwardgar:
Thanks!

bencaradocdavies:
Yes, it is weird. It was added in this commit: https://github.com/bencaradocdavies/thredds/commit/3951ee18776985d4534a9298ade9719d5ddfd00e#diff-cfd0424cb0b22e474e2395a950406a00

Auto-import run amok, I say.  :-)

PR #599, opened 2016-07-26
Fix GradleRunner failures on Windows

cwardgar:
Ah, Windows paths. How I've missed you.


lesserwhirls:
fire

Sorry, I could not resist...

Issue #576, opened 2016-06-27
RDA time coordinates incorrectly mapped to GRIB records


JohnLCaron:
Report on ds628.0_fcst_surf from RDA user:

Hi Dave, hope life is treating you well and you got through the el nino winter alright. DOwn here in Perth, it's starting off  quite wet with a very stormy may.

I have been using  the RDA data quite a lot in my work, in particular the JRA55 data to drive storm surge models. We are finding it does quite well for Australia and it's particularly good in the way that tropical  cyclones are in the correct location, which is not the case for ERA-I.

For the earlier cases I just did the data subset requests through the web brower as i didn't realise that  I had access to the thredds/opendap server. We are planning on doing a multi-year run and I was happy to find the server and the ncss functionality and that I can download with wget. Very useful!

However, I downloaded a year of data  from the 3-hourly  0.5 deg instantaneous 2d  and Best datasets (see ncss request below). But we are having trouble deciphering what has happened to the time dimension organisation of the files.

If I load the time variable (best) and convert I get nicely ordered times (2010-12-31 21:00z to 2012-01-01 00:00z)

But the corresponding data appears to be mixed up in the time dimension so when i plot each timestep or  the mslp for example, the storms move back and forth in time - ie. it appears to be in the order    6hr  3hr  12h 9hr r etc.

As a workaround I have tried to reorder the data so that the it looks nice and doesn't shift back and forth as you move forward in time.  But then I have an issue of the second (and last)  timesteps being NaN's .

I think i can resolve this by eliminating those and cutting the first and last of the time variable which leaves me with data (i think) corresponding to 2011-01-01 00:00z to 2011-12-31 21:00z.

I have tested for a know storm event, but  I am wary of messing this up and getting the wrong times.

Am I missing something?  Do you know what's going on here?

I also tried with the other (non-best) version with has an extra dimension  Pressure_reduced_to_MSL_msl(reftime, time, lat, lon)  and got  got the same results.

Hopefully this isn't too confusing and my question makes sense.  I was very confused!

Thanks for your time.

Cheers
yasha

ncss request:

[...]

PR #603, opened 2016-07-30
5.0.0

JohnLCaron:
1) refactor GRIB, coverage for bug

https://github.com/Unidata/thredds/issues/576

2) add tests that coverage, gridDataset coordinates match the gbx coordinate. found a number of other issues. now much better!

The ncx4 indices should be regenerated everywhere after this is checked in.


cwardgar:
Looks good. `ucar.nc2.iosp.hdf5.TestN4problems.testTiling2` is failing on Jenkins because `UpperDeschutes_t4p10_swemelt.nc` from #518 isn't present in the test data suite. However, that file is 480 MB, so I'll try to subset it first.

PR #610, opened 2016-08-11
Godiva bug

lesserwhirls:
There are actually two things here:

1. fixed the godiva issue where nothing was being displayed in the godiva client
2. fixed two compile errors related to NCSS params (stations -> stn), but I'm not sure why that one cropped up (happened after I rebased).

@cwardgar - can you check over those two NCSS related changes to see if I screwed something up?

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -204,15 +204,17 @@ libraries["oro"] = "oro:oro:2.0.6"
 
 libraries["thymeleaf-spring4"] = "org.thymeleaf:thymeleaf-spring4:2.1.4.RELEASE"
 
-versions["edal"] = "1.1.2"
+versions["edal"] = "1.2.1"
 
 libraries["edal-common"] = "uk.ac.rdg.resc:edal-common:${versions["edal"]}"
 
 libraries["edal-cdm"] = dependencies.create("uk.ac.rdg.resc:edal-cdm:${versions["edal"]}") {
     // These are dependencies that are included in edal-cdm, but are provided by
     // us already, so we will exclude these from the edal-cdm artifact.
-    exclude group: 'edu.ucar', module: 'grib'
+    exclude group: 'edu.ucar', module: 'cdm'
+    exclude group: 'edu.ucar', module: 'udunits'
     exclude group: 'edu.ucar', module: 'netcdf4'
+    exclude group: 'edu.ucar', module: 'grib'
     exclude group: 'edu.ucar', module: 'opendap'
     exclude group: 'com.sleepycat', module: 'je'
     exclude group: 'oro', module: 'oro'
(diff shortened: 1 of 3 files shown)


cwardgar:
Those conflicts you saw are related to my changes [here](https://github.com/Unidata/thredds/commit/19914c64fff4c6485d0423664f1225a200d74ee0#diff-b895bc3120f8facbdffea7f51094a6b7): "stn" became "station". I did that for consistency with the other fields and methods of `SubsetParams`. For example, we have `SubsetParams.variable` instead of `SubsetParams.var` and `getVariables()` instead of `getVar()`. Can you add a commit to use the new names?

cwardgar:
Actually, I'm not sure why you had conflicts in the first place, since nothing in 65c4f776bd5a2834d0d3040f74796fe889ac275a is related to my changes. Can you just rebase that commit onto the latest 5.0.0?

lesserwhirls:
Hi @cwardgar - Not sure how, but I cannot get 5.0.0 to build:

```java
tds:compileJava
/Users/lesserwhirls/dev/unidata/thredds/50/tds/src/main/java/thredds/server/ncss/params/NcssPointParamsBean.java:84: error: cannot find symbol
      subset.set(SubsetParams.stations, stns);
                             ^
  symbol:   variable stations
  location: class SubsetParams
/Users/lesserwhirls/dev/unidata/thredds/50/tds/src/main/java/thredds/server/ncss/view/dsg/station/AbstractStationSubsetWriter.java:205: error: cannot find symbol
    if (ncssParams.getStations() != null) {
                  ^
  symbol:   method getStations()
  location: variable ncssParams of type SubsetParams
/Users/lesserwhirls/dev/unidata/thredds/50/tds/src/main/java/thredds/server/ncss/view/dsg/station/AbstractStationSubsetWriter.java:206: error: cannot find symbol
      List<String> stnNames = ncssParams.getStations();
                                        ^
  symbol:   method getStations()
  location: variable ncssParams of type SubsetParams
```

It looks like the stns -> stations change is not on 5.0.0. Not sure how travis passed on the last PR I approved for merging into 5.0? So confused at the moment.

lesserwhirls:
Ok, wait...now it builds (?). Has to be something borked on my system. Let me dig in again.

lesserwhirls:
Ok, it looks like things break when I update to edal-java version 1.2.1. I wonder if it pulls in something older based on an old snapshot? I've updated this PR to include a cleaner copy of sutff, without the changes to the ncss code. I'll investigate more.

lesserwhirls:
Ok, just needed to add some excludes to the edal-cdm dep. Let's see how this goes.

cwardgar:
Ah, that makes sense: an old version of cdm was being pulled in. Looks good now though.

Issue #612, opened 2016-08-15
Debug output of H5header is always printed to console


marpet:
In the H5header class the field debugOut is used to print debug information.
At some locations the messages are printed even if debug information is not wanted.
For example:
[H5header.java#L1325](https://github.com/Unidata/thredds/blob/4c3d30702108253c7f8e4480970a7608f611d6ed/cdm/src/main/java/ucar/nc2/iosp/hdf5/H5header.java#L1325)
[H5header.java#L1333](https://github.com/Unidata/thredds/blob/4c3d30702108253c7f8e4480970a7608f611d6ed/cdm/src/main/java/ucar/nc2/iosp/hdf5/H5header.java#L1333)

This leads to unwanted output on the console and can make our users feel insecure about the correctness of the software.

msdsoftware:
Last year when I made a bug fix in `H5header`, there was some discussion of cleaning up the logging there and elsewhere. Looks like no one actually got on it.

lesserwhirls:
There are quiet a bit of places in the codebase that print out debug messages, and we've been trying to move those over to using a logger when we come across them, but most of those are likely on the master branch. We'll take a look at the H5Header class specifically to make this happen.

If you run across any particularly verbose debugging statements like this, let us know here and we'll prioritize those to knock them out as quickly as we can. If you could label any of those issues with Area: Logging to make them easier for us to to find.

msdsoftware:
Turns out the prior discussion was not when I posted an H5header bug fix but when I posted #pull request #426 last winter.

cwardgar:
Fixed in #617.

marpet:
Thanks for the quick reaction

PR #617, opened 2016-08-23
Clean up logging a bit in H5header.java

msdsoftware:


msdsoftware:
Mostly changing println calls to instead use log.debug; some changes of printf calls to log.debug.

If this pull is accepted, then you can close issue #612.

cwardgar:
Eventually I'd like to eliminate the boolean flags and instead use [SLF4J Markers](http://logback.qos.ch/manual/filters.html#sampleTurboFilterConfig), but this is strictly better than before, where everything was being dumped to STDOUT. Thanks @msdsoftware!

PR #619, opened 2016-08-24
Get threddsIso services running in TDS

lesserwhirls:
Add threddsIso deps to gradle build
    Add jaxen runtime dep


lesserwhirls:
Ok @cwardgar - I made the changes we discussed this morning. Thanks!

PR #623, opened 2016-09-01
Dap4.part2.dmh

DennisHeimbigner:
This is the second part of the major DAP4 rewrite.  It also
includes ancillary changes for testing and to support test
server (d4ts and dts). 
The remotetest server d4ts server is down and I do not
have access to fix, to some tests like TestConstraints
are failing for now. But it appears to be passing travis.


cwardgar:
b6b82eb76af10fa685c75480e313bd745d034c58 modifies every line in `ArraySequence` and `FeatureDatasetFactoryManager`, despite your intended changes being only a few lines. Same thing may be happening in `CdmNode` and others as well. Such whole-file changes make `git blame` useless, which is a pretty big deal. I'd really like to avoid them.

The problem is almost certainly related to git's [core.autocrlf](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#coreautocrlf-7kTgIgFwI4) setting. You're on Windows, so yours is probably set to `true` (check by doing `git config core.autocrlf`). However, [this post](http://stackoverflow.com/questions/3206843/how-line-ending-conversions-work-with-git-core-autocrlf-between-different-operat/14039909#14039909) suggests that anything but `false` is just a big headache.

Can your editor deal with LF line endings (IntelliJ can)? If so, I suggest setting that option to `false` so we don't have this problem in the future. As for `ArraySequence` and `FeatureDatasetFactoryManager`, I'm not sure there's a way to revert their every-line modification. @dopplershift Do you know of a way?

cwardgar:
Alternatively, the `core.autocrlf=false` setting could be enforced [per-repository](https://help.github.com/articles/dealing-with-line-endings/#per-repository-settings) with a .gitattributes file:
````
# Unsetting the text attribute on a path tells Git not to attempt any end-of-line conversion upon
# checkin or checkout.
* -text
````
Probably worth doing. We could also override the global setting and force line endings for certain files:
````
*.bat text eol=crlf
*.sh text eol=lf
````

DennisHeimbigner:
I get it set set during git install. I note that it is set to 'input' which
from that post seems ok. In any case, and assuming they have not changed.
it should be possible to check them out of 5.0.0.
NOTE: what I would prefer is that we convert all of our files to use LF only (not CRLF);
even on windows.

cwardgar:
The purist in me agrees: all line-endings in the repo (and in the world, really) should be LF, not CRLF. However, from a practical perspective, our editors all seem to be able to handle both, so doing the conversion doesn't seem like it's worth the cost of breaking `git blame`. And we have a lot of CRLF files because John developed on Windows. To that end, I've added a .gitattributes file in #624 which tells git not to do any EOL conversion except for *.sh and *.bat files.

If you'd remove those two Java files from the commit, I'd appreciate it. Might want to diff them against the old versions to see if there were actually any substantive changes though. You'll probably have to do some EOL manipulation to make the diff useful.

Interestingly, on my Mac, I had `core.autocrlf` set to `input` as well (since changed to `false`), but I never had a problem with git converting CRLF files to LF. Must be a Windows thing.

DennisHeimbigner:
It occurs to me that this is not the only source of problems.
Since we all use different formatting options in Intellij,
accidentally reformatting a file will also cause problems for
git blame.

But I am curious; does not the -w option to git blame solve
at least the crlf problem?

=Dennis

cwardgar:
You're right, the -w option might do the trick for `git blame`. However, I mostly look at that info in the form of [IntelliJ annotations](https://www.jetbrains.com/help/idea/2016.2/working-with-annotations.html). I'm sure it uses `git blame` underneath, but I don't know of a way to make it pass the -w option.

Different formatting options are a problem, yeah. We could all standardize on one format and then add some automation to enforce it, but I doubt we could agree. Instead, I just try to match the existing format of whatever file I'm working on. (Why only two spaces for indents, John? We have wide-screen high-resolution monitors, you know.)

PR #625, opened 2016-09-06
bump gradle shadow plugin, error 403 issue

vegasm:


DennisHeimbigner:
Please do not put multiple independent changes sets
into a single pull request. Separate it into three
requests because I will probably reject the 403 change.

cwardgar:
In this case, the version bump of the shadow plugin had already been done in a previous commit, so the 403 stuff is the only outstanding change in this PR. But yeah, that's good advice in general.

DennisHeimbigner:
Let me review the 403 issue again and decide.

DennisHeimbigner:
As I remembered, the opinions on this are all over the map.
In this specific case (TomcatAuthorizer), the situation is this:
1. the user authenticated to Apache+Tomcat ok by providing
    e.g. a proper user+pwd.
2. TomcatAuthorizer is denying access because that user
    does not have the necessary (tomcat) role to gain access.
The argument for 403 is:
The server understood the request, but is refusing to fulfill it.Authorization will not help and the request SHOULD NOT be repeated.

The two clearest discussions I found were:
https://recalll.co/app/?q=403%20Forbidden%20vs%20401%20Unauthorized%20HTTP%20responses
and
http://stackoverflow.com/questions/3297048/403-forbidden-vs-401-unauthorized-http-responses

DennisHeimbigner:
As I remembered, the opinions on this are all over the map.
In this specific case (TomcatAuthorizer), the situation is this:
1. the user authenticated to Apache+Tomcat ok by providing
    e.g. a proper user+pwd.
2. TomcatAuthorizer is denying access because that user
    does not have the necessary (tomcat) role to gain access.
The argument for 403 is (from RFC) "The server understood the request, but is refusing to fulfill it.Authorization will not help and the request SHOULD NOT be repeated."

The argument for 401 is that the user could retry using a different user+pwd that
has the proper role. The argument agains 401 is that "The origin server MUST send a WWW-Authenticate header field (Section 4.4) containing at least one challenge applicable to the target resource."

The underlying problem is that we are using HTTP level codes for application authentication
and authorization.

I think the WWW-Authenticate header issue probably means that we should use 403. But
I could be persuaded otherwise. Also from a client's point of view, 401 is more informative
because it indicates the problem is authentication. 403's definition is a bit nebulous.

The two clearest discussions I found were:
https://recalll.co/app/?q=403%20Forbidden%20vs%20401%20Unauthorized%20HTTP%20responses
and
http://stackoverflow.com/questions/3297048/403-forbidden-vs-401-unauthorized-http-responses

DennisHeimbigner:
Super-ceded by pr 754

PR #636, opened 2016-09-16
Reorganized the Gradle build logic

cwardgar:
There's now a clear separation between the script plugins that should only be applied to the root project, and those that can be applied to any project (including root). As a result, we're no longer doing configuration injection anywhere, which is a powerful technique but often difficult to follow. Now, for any given subproject, it's easy to see the scripts that are being applied to it, since no configuration is being injected "behind your back".

* Declared inputs and outputs for :coveralls Task. This allows it to participate in Gradle incremental build machinery. Before, it was never considered UP-TO-DATE. I needed to upgrade "coveralls-gradle-plugin" to 2.6.3 in order to get the task output.
* Subproject JacocoReports now only report on the source code in their subproject, instead of the entire codebase. :rootJacocoReport still reports on the entire codebase.
* Fixed issue where :cdm-test:test and :cdm-test:testIndexCreation could never be UP-TO-DATE.


cwardgar:
This is a pretty big change. I'll kick off a Jenkins build to make sure I got it right.

lesserwhirls:
How did the Jenkins build go?

cwardgar:
Success! This is ready to be merged.

lesserwhirls:
Thanks @cwardgar!

PR #640, opened 2016-09-21
Update dependencies and upload embedded TDS logs

cwardgar:
* Updated Gradle from 2.14 to 3.1
* Updated thredds dependencies where possible. These updates were almost entirely limited to minor and patch versions. Nonetheless, attempting to update some deps caused breakage—notably `spring-security` and `jgoodies-forms`. Those breakages have been noted in dependencies.gradle.
* Replaced usages of `jcip-annotations` with `jsr305` and then removed `jcip-annotations` from the project. They provide the same annotations; there's no reason to have both. See http://stackoverflow.com/questions/6322910/

This is another big changeset, so I did a Jenkins run ([cwardgar-35](https://jenkins-aws.unidata.ucar.edu/job/cwardgar/35/testReport/)). Same number of test failures as before the change ([cwardgar-31](https://jenkins-aws.unidata.ucar.edu/job/cwardgar/31/testReport/)). Should be okay to merge.


cwardgar:
Blargh. These Travis failures have been happening more often and I have no idea why. [This](https://github.com/travis-ci/travis-ci/issues/5582) is one possibility. A big problem is that the failure occurs after startup of the embedded TDS and those logs are effectively hidden from us. I'm gonna see if I can figure out a way to upload them to S3.

cwardgar:
Upload the logs generated by the 'it' module's embedded TDS to S3.
* Can only occur on branches of the Unidata repo.
* Improved the printing of artifact S3 URLs.
* Reorganize Travis logic a bit and remove cruft.
* Fix TestAdminDebug. Until now, I don't think it was doing anything remotely useful.
* Added a log4j2-test.xml to "it". It can be used to change specific logger levels while debugging.

cwardgar:
This is ready to merge. I'm not sure if I fixed the original cause of the hang in `:it:integrationTest`, but if it happens again, at least we can look at the logs now.

lesserwhirls:
It's Log!

cwardgar:
All kids love log!

Issue #638, opened 2016-09-19
IospHelper cast error, Integer to Short


msdsoftware:
Trying to work with an HDF-4 dataset that has a variable described in CDL as

```
    short 10-day_averaged_land_surface_temperature(height=5000, width=8000);
      :_Unsigned = "true";
      :Unit = "Kelvins";
      :valid_range = 220S, 350S; // short
      :_FillValue = 65535; // int
      :Gains = 0.01f; // float
      :Offsets = 0S; // short
      :long_name = "10-day averaged land surface temperature";
```

When trying to retrieve the data via `Variable.read(int[], int[])`, I am getting a class cast error:

```
java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.Short
	at ucar.nc2.iosp.IospHelper.makePrimitiveArray(IospHelper.java:678)
	at ucar.nc2.iosp.IospHelper.readDataFill(IospHelper.java:78)
	at ucar.nc2.iosp.hdf4.H4iosp.readData(H4iosp.java:104)
	at ucar.nc2.NetcdfFile.readData(NetcdfFile.java:2020)
	at ucar.nc2.Variable.reallyRead(Variable.java:913)
	at ucar.nc2.Variable._read(Variable.java:898)
	at ucar.nc2.Variable.read(Variable.java:709)
	at ucar.nc2.dataset.VariableDS.reallyRead(VariableDS.java:557)
	at ucar.nc2.dataset.VariableDS._read(VariableDS.java:537)
	at ucar.nc2.Variable.read(Variable.java:709)
	at ucar.nc2.Variable.read(Variable.java:655)
	at somewhere in my code.
```

The line where the exception occurs is attempting to cast the fillValue.

msdsoftware:
Fleshing this out a bit, the problem is not necessarily in IospHelper but possibly instead in the HDF-4 handler. Instead, it seems that the data are extracted as a short DataType, although the attribute _Unsigned = "true" suggests the data values would be better taken as int. But being valued as 65535, the fill value attribute is extracted as an int, and when IospHelper tries to save ac copy of it, it tries to cast it to a short in order to match the DataType of the data array. Since an integer can't be cast to a short, the exception occurs.

JohnLCaron:
Can you send an example file?

msdsoftware:
[...]

PR #642, opened 2016-09-22
fix issue #638

JohnLCaron:


msdsoftware:
Noting that this was performed on the 5.0.0 branch.

Any chance the like can be done on the master branch also, or is that too much a hairball?

lesserwhirls:
I think this can be done on 4.6. Let me make a PR real quick.

lesserwhirls:
Ok, it seemed pretty simple. PR Unidata/thredds#643

lesserwhirls:
@JohnLCaron - thank you so much for this PR!

msdsoftware:
This seems to have worked. Building a netcdfAll SNAPSHOT and using that in Panoply, I can now read the problem data array. Not yet able to plot it correctly yet, but I think that's a separate issue involving the valid_range attribute. Thx.

PR #655, opened 2016-10-17
Fix Google Analytics tracking of TDS page views

cwardgar:
Fixes eSupport ticket ABX-989330.


cwardgar:
This feature broke years ago (probably 2013) and nobody noticed. It would be nice to have a functional test for it so that we're not blindsided again in the future, but it would take a lot of effort to write one. [This](https://developers.google.com/analytics/devguides/reporting/core/v4/) seems like it would be the place to start.

PR #665, opened 2016-10-26
TDS global error handler logs exceptions to threddsServlet.log

cwardgar:
* Previously, those exceptions were just being swallowed.
* Removed custom error page JSPs. We're not using them any more now that we have TdsErrorHandling in 5.0.0.


lesserwhirls:
Yay for not swallowing exceptions! 👍

PR #666, opened 2016-10-26
Fixed broken link.

sudouser2010:
The original link was a dead link so I routed it to the intended page.


cwardgar:
That link is an AsciiDoc cross reference. It actually works, but not on GitHub. Also, we're transitioning our docs to Kramdown, so that syntax isn't long for this world anyway.

PR #679, opened 2016-11-12
Fixed dependency issues discovered by 'gradle-lint-plugin'

cwardgar:
* Most issues were related to dependencies that were needed at compile time (directly), but weren't marked as such. Instead, those deps were being pulled in transitively (usually through 'cdm').
* Some classes in cdm-test were inadvertently using an internal method from jcommander when they were intended to use a same-named method from Guava. I modified those usages and was able to remove jcommander as a dep from cdm-test as a result.
* Eliminated use of 'httpcore' from legacy.
* Eliminated two cases of multiple jars on the classpath containing the same class. One was "com.eclipsesource.restfuse" (now removed from project) and the other was the GWT deps (they belong to their own Configuration now; no reason for them to be in "provided").
* Added 'legacy' to ncIdv at Yuan's request.

### DIFF
--- a/bufr/build.gradle
+++ b/bufr/build.gradle
@@ -14,4 +14,5 @@ dependencies {
     compile libraries["jcommander"]
     compile libraries["protobuf-java"]
     compile libraries["slf4j-api"]
+    compile libraries["guava"]
 }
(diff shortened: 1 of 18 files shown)


lesserwhirls:
Looks good!

PR #681, opened 2016-11-15
Removed "52n-oxf-xmlbeans", which is dragging in GPLed dependencies

cwardgar:
* It turns out that we were using very few of the classes from that dep.
* The license checker show NO outstanding uses of GPLed code!


DennisHeimbigner:
Excellent.

lesserwhirls:
👍

PR #685, opened 2016-11-22
Add GeoCSV output option for NCSS

lesserwhirls:
Adds GeoCSV output option for NCSS GridAsPoint, Point, and Station requests.


lesserwhirls:
Ok, added to existing test (although, admittedly, not much of a test) and did a bit more code cleanup.

cwardgar:
A smoke test is better than no test.

PR #687, opened 2016-11-25
Wrong units on "HighResolution: Base Velocity"

### LINKED ISSUE #674: Confirm Units of Knots for NIDS Base Velocity Data
According to a recent e-support (`ticketid` `27672`) there is ambiguity on whether the NIDS data, base velocity field is in knots or meters per second. @yuanho Can we have a confirmation of the units of knots for the NIDS data base velocity field in 65eda31? (We have the usual problem of knots versus meters per second being only a factor of 2 off from each other.) According to @mike-dixon, the data are probably in meters per second. I have yet to find any NEXRAD documentation that details anything about units pertaining to NIDS data.

julienchastang:
closes #674

### DIFF
--- a/cdm/src/main/java/ucar/nc2/iosp/nids/Nidsheader.java
+++ b/cdm/src/main/java/ucar/nc2/iosp/nids/Nidsheader.java
@@ -2519,7 +2519,7 @@ class Nidsheader{
             ctilt = pname_lookup(99, pLevel);
 
             ctitle = "HighResolution: Base Velocity";
-            cunit = "KT";
+            cunit = "m/s";
             cname = "BaseVelocityDV";
             summary = ctilt + " is a radial image of base velocity field and its range 124 nm";
 


lesserwhirls:
Thanks @julienchastang!

dopplershift:
@julienchastang Did you check products 93, 154, and 182 as well? Those are supposed to have 'm/s' too.

julienchastang:
@dopplershift See latest PR (though don't merge yet).

julienchastang:
I believe 93 and 154 are not addressed in the `Nidsheader.java`. @dopplershift Can we have a unit check on products 55 and 56 (i.e., Storm Relative Mean Velocity) currently listed as knots?

dopplershift:
55 and 56 are legacy products; I can't find anything in the ICD that actually specifies units. I think knots is probably correct.

93 missing doesn't surprise me, but I'm surprised 154 wouldn't be addressed in there since that's a high res, current velocity product.

julienchastang:
As far as I can tell, 154 is not addressed in the `NidsHeader.java`. I don't why this is. Feel free to look as well. Really, all I have been doing is working backwards by searching for everything that has units of `KT`.

dopplershift:
If I give you a file for 154, do you have cycles to throw it into the IDV/netcdf-java?

julienchastang:
Sure, but can we turn 154 into a separate issue so that we can merge #689?

dopplershift:
Seems I don't have any 154 products either--they're probably not shipped across NOAAPORT. I wouldn't worry about it for now.

julienchastang:
Alrighty, then #689 is ready to merge?

PR #689, opened 2016-11-27
NIDS Product 182 Units from KT to m/s

### LINKED ISSUE #674: Confirm Units of Knots for NIDS Base Velocity Data

julienchastang:
Closes #674. Related to #687 #688. Let's code review early next week before merging. There may be other unit issues, as well, which we may or may not want to piggyback on top of this PR.

### DIFF
--- a/cdm/src/main/java/ucar/nc2/iosp/nids/Nidsheader.java
+++ b/cdm/src/main/java/ucar/nc2/iosp/nids/Nidsheader.java
@@ -2756,11 +2756,11 @@ class Nidsheader{
                 lon_min = longitude + t2;
                 lon_max = longitude - t2;
                 summary = ctilt + " is a radial image of base velocity" + (prod_elevation/10 + 1) +  " and  range 32 nm";
-                cunit = "KT";
+                cunit = "m/s";
             }
             else {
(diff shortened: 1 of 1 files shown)


lesserwhirls:
Thanks @julienchastang 👍

julienchastang:
Excellent. Thank you. @cwardgar can we have a new 4.6 branch ncidv.jar? That will let me close a long standing e-support inquiry. Thanks again.

lesserwhirls:
Here is the new [jar](https://artifacts.unidata.ucar.edu/service/local/repo_groups/public/content/edu/ucar/ncIdv/4.6.7-SNAPSHOT/ncIdv-4.6.7-20161129.235044-12.jar), @julienchastang

PR #701, opened 2016-12-05
Removed all use of the dtswar located at remotetest.unidata.ucar.edu

cwardgar:
* Instead, we have an embedded dtswar instance running at localhost:8082/dts. It runs during `:opendap:test` and `:httpservices:test`. This PR is the first step in removing our dependency on `remotetest.unidata.ucar.edu`. In a broader sense, the goal is to not have **any** external resource dependencies in our test suite whatsoever.

* We now incorporate a forked version of the Gretty plugin: ([Bintray](https://bintray.com/cwardgar/maven/gretty-fork), [GitHub](https://github.com/cwardgar/gretty)). [The original](https://github.com/akhikhl/gretty) isn't actively maintained anymore so there wasn't much chance that I could get my needed bug fixes into mainline.

* Moved several tests out of `:cdm-test` and into `:httpservices`. Those tests required a dtswar endpoint and I didn't want to run an instance in `:cdm-test:test`. So instead, we're doing it in `:httpservices:test`. Besides, those tests logically belong in `:httpservices` anyway (they test HTTP stuff). Previously, `:httpservices` had no tests whatsoever.

* Removed `ftpdir` Gradle property, as we no longer use it.

* Fixed and/or improved several opendap tests.
  * Parameterized `TestDapParser` using "JUnitParams". That allowed me to nuke `TestDAS`, `TestDDS`, and `TestErr`.
  * Some tests in ucar.nc2.util.net were swallowing exceptions. Bad!
  * Removed `NeedsExternalResource` annotation from the tests that no longer rely on the external dtswar. They'll run on Travis now as a result.

This was a pretty big changeset, so I ran a build on [Travis](https://travis-ci.org/cwardgar/thredds/builds/181254790) and [Jenkins](https://jenkins-aws.unidata.ucar.edu/job/cwardgar/42/). Looks good.

@DennisHeimbigner I mostly modified your code here, so please have a look and let me know if I did anything dumb.


cwardgar:
@DennisHeimbigner Any thoughts on this?

DennisHeimbigner:
> Moved several tests out of :cdm-test and
> into :httpservices. Those tests required a dtswar endpoint and I
> didn't want to run an instance in :cdm-test:test. So instead,
> we're doing it in :httpservices:test. Besides, those tests
> logically belong in :httpservices anyway (they test HTTP stuff).

This is probably incorrect. All dtswar stuff should be in opendap.
Httpservices module deliberately has no tests. So, the tests you moved
should robably be moved again to somewhere in the opendap module.

dopplershift:
> Httpservices module deliberately has no tests. So, the tests you moved
Why?

DennisHeimbigner:
Because it is used in non-thredds environments that I have. So I chose to keep
the tests elsewhere.

cwardgar:
Classes I moved from `:cdm-test` to `:httpservices`:
* `Login`: not a test class, but used by `TestTomcatAuth`
* `TestFormBuilder`: will apparently work with any live URL. Currently using `TestDir.remoteTestServer`.
* `TestHTTPMethod`: needs `TestDir.dap2TestServer`
* `TestHTTPSession`: needs `TestDir.dap2TestServer`
* `TestHang`: needs `TestDir.dap2TestServer`
* `TestSSH`: needs TDS at `TestDir.remoteTestServer`
* `TestState`: needs `TestDir.dap2TestServer`
* `TestThreading`: needs `TestDir.dap2TestServer`
* `TestTomcatAuth`: needs TDS at `TestDir.remoteTestServer`
* `TestURIParse`: Appears to just parse URL strings. However, there's an ignored test that makes an HTTP connection to `http://localhost:8081/thredds/cdmremote/scanCdmUnitTests/formats/hdf5/grid_1_3d_xyz_aug.h5?req=data&var=HDFEOS_INFORMATION/StructMetadata\\.0`. WTF?

All of the moved files appear to test classes in `:httpservices`. At the least, they all depend on classes in that module.

>> Httpservices module deliberately has no tests.

> Because it is used in non-thredds environments that I have. So I chose to keep
the tests elsewhere.

I don't understand this statement. Or maybe I just don't understand why it matters. Can you elaborate? Where would you move the above classes? Note that I plan to get rid of `:cdm-test` eventually, as the `NeedsCdmUnitTest` annotation obviates the need for a separate module for tests that rely on that data.

DennisHeimbigner:
There are two different points:
1. dtswar specific tests should be in opendap module.
2. httpservices specific test should stay in cdm-test.
I would prefer strongly that the httpservices module
be kept test free.
=Dennis

DennisHeimbigner:
Never mind. Put the httpservices tests into httpservices and I will
deal with the consequences.

DennisHeimbigner:
I see part of the problem. The httpservices tests require
access to a server that can provide data with certain properties.
I used e.g. the dtsserver and the remotetest thredds servers
because they were convenient and I could control the
datasets being served. Mostly, the tests do not need these
servers specifically if we had another server to provide data
with the required properties. Not sure how to fix this. Ideally,
we need some kind of minimal real server to talk to and from
which we can get usable test data.

cwardgar:
Hmm, okay. Hopefully the change doesn't cause you too much grief. One last clarification: `TestHTTPMethod`, `TestHTTPSession`, `TestHang`, `TestState` and `TestThreading` depend on `TestDir.dap2TestServer`. However, it's not clear to me that all of them (any of them?) actually require the features and/or data specifically provided by dtswar. That is, they would work just as well with any live HTTP endpoint. Therefore, I think the best **logical** place for those tests is still `:httpservices`, not `:opendap`, even though they technically depend on dtswar. Do you agree?

In the future, perhaps we can deploy a different, smaller webapp in `:httpservices:test`, but dtswar is probably fine for now.

cwardgar:
Oh, I just saw your newest comment. Looks like we were thinking the same thing. So do you think those tests should be left in `:httpservices`?

DennisHeimbigner:
I analyzed the tests in hredds/cdm-test/src/test/java/ucar/nc2/util/net.
My notes are at the end. Summary, with the exception of TestStream,
the tests either require no server, or require a server that just serves up
specific files, and that also supports byte-ranges.
Two tests: TestSSH and TestTomcatAuth require the ability to set
certificates, roles, and users for a tomcat server.
=Dennis

notes:
=============================
TestFormBuilder.java
- refs remotetestserver as post target, but ignores any failures

TestHang.java
- Test that a single HTTPSession with many method invocations
  does not hang properly releases the connections.

TestHTTPMethod.java
- test stream access to a reply

TestHTTPSession.java
- looks at the request before final sending to see
  if various headers were properly set

TestMisc.java
- tests byte range
- Test that a large number of open/close does not lose connections.

TestSSH.java
- Originally could not be tested in jenkins or travis
  because it needed a tomcat with proper certificates.
  Might be possible now.

TestState.java
- Test the proper state transitions of HTTPSession
  and HTTPMethod and HTTPMethodStream.

TestStream.java
- Caron built this. It uses CatalogBuilder, so it needs
  to be somewhere else than in httpservices module.

TestThreading.java
- Test interaction of multi-threading with httpservices.
  e.g. max connections

TestTomcatAuth.java
- Originally could not be tested in jenkins or travis
  because it needed a tomcat basic auth enabled
  and with specific roles and users defined.
  Also needs a catalog with specific entries to support
  the thredds tomcat auth code.
  See notes at front of the file for details.

TestURIParse.java
- Has a specific test for some Caron detected problem,
  otherwise does not require a server. The Caron test
  can probably now be eliminated.

cwardgar:
Okay, sounds like this PR is ready to be merged.

cwardgar:
Bump.

lesserwhirls:
I can't rebase this one through the web gui - would you mind @cwardgar? Thanks!

cwardgar:
Okay, ready to merge.

PR #704, opened 2016-12-07
add default services, catalogScan, and FMRC examples to starup catalogs

lesserwhirls:


tds/src/main/webapp/WEB-INF/altContent/startup/catalogs/exampleFmrc.xml
```diff
@@ -0,0 +1,18 @@
...
+           http://www.unidata.ucar.edu/schemas/thredds/InvCatalog.1.0.8.xsd">
+
+    <dataset name="Example FMRC">
+    <featureCollection name="GFS 0.25 degree" featureType="FMRC" harvest="true" path="GFS/0p25">
+    <metadata inherited="true">
+        <dataType>Grid</dataType>
+        <dataFormat>netCDF</dataFormat>
+        <documentation type="summary">Example GFS 0.25 degree output from the NetCDF Subset Service</documentation>
+    </metadata>
+        <collection spec="${tds.content.root.path}/thredds/public/testdata/GFS_Global_0p25deg_#yyyyMMdd_HHmm#\.grib2\.nc4$" />
```
cwardgar: Would the user manually have to edit the `spec` attribute here? That is, the catalog wouldn't work out of the box? I think we can fix that. I know for sure that we have specs starting with `${cdmUnitTest}` in our test catalogs that are undergoing textual substitution.

I'm curious whether `spec="${content}/testdata/...` would work here. See [this](https://github.com/Unidata/thredds/blob/5.0.0/tds/src/main/java/thredds/core/DataRootManager.java#L104).

lesserwhirls: I was hoping that CatalogScan would pick up the path change in the spec and thus updating the catalog would show off CatalogScan...however, the change does not get picked up, so that's lame. Maybe CatalogScan does not work like I though. Maybe it only recognizes the addition of a new catalog or the removal of a catalog from the CatalogScan directory, not simple modifications.

dopplershift: It may be straightforward to add tracking for modified files though--depending on how it finds out about additions/removals.

Issue #705, opened 2016-12-07
Empty FMRC produces odd error message


lesserwhirls:
When the FMRC does not find any files, the behavior is odd. First, a general skeleton catalog is produces (links to best, 2D time, files, forecast model run), but when you try to access catalogs/files under those, TDS produces the error message:

FileNotFound: /Users/sarms/dev/unidata/content/5.0/thredds/GFS/0p25/runs/catalog.xml

when really the issue is that the collection is empty.

cwardgar:
I believe this is what we were seeing in #664.

PR #706, opened 2016-12-07
add informative message when FMRC is empty

lesserwhirls:
fixes Unidata/thredds#705

### DIFF
--- a/tds/src/main/java/thredds/featurecollection/InvDatasetFcFmrc.java
+++ b/tds/src/main/java/thredds/featurecollection/InvDatasetFcFmrc.java
@@ -1,34 +1,5 @@
 /*
  * Copyright 1998-2015 John Caron and University Corporation for Atmospheric Research/Unidata
- *
- *  Portions of this software were developed by the Unidata Program at the
- *  University Corporation for Atmospheric Research.
- *
- *  Access and use of this software shall impose the following obligations
- *  and understandings on the user. The user is granted the right, without
- *  any fee or cost, to use, copy, modify, alter, enhance and distribute
- *  this software, and any derivative works thereof, and its supporting
- *  documentation for any purpose whatsoever, provided that this entire
- *  notice appears in all copies of the software, derivative works and
- *  supporting documentation.  Further, UCAR requests that the user credit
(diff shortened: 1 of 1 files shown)


lesserwhirls:
I don't like the result of this though. The runtime error message collides with the TDS IOException and it reads a bit...funny.

cwardgar:
It collides with IOException? I see it being wrapped and rethrown as an IOException in `FeatureCollectionCache.makeFeatureCollection()`, if that's what you mean. Definitely some weirdness happening there.

Also, if possible, the exception would be more useful if it included the scan location and regex.

lesserwhirls:
I think that's ok, but because of `TdsErrorHandling.handle(IOException ex)`, we get the text "IOException sending File" added to the message, so this is what we get in the browser:

`IOException sending File java.lang.RuntimeException: The FMRC GFS_0.25_degreecannot find any files in the collection.`

It just reads funny.

I'll look at adding the addition info as well.

lesserwhirls:
Ok, I've logged the specific info rather than displaying file system info in the browser message. The message looks like:

`[2016-12-07T17:53:24.155-0500] WARN  thredds.featurecollection.InvDatasetFcFmrc: The FMRC GFS_0.25_degree defined in /thredds/catalog/catalogs/exampleFmrc.xml cannot find any files matching the spec /Users/sarms/dev/unidata/content/5.0/thredds/public/testdata/GFS_Global_0p25deg_#yyyyMMdd_HHmm#\.grib2\.nc4$.`

cwardgar:
That message looks good to me. Your diff went crazy though. I wonder why that happens? I've seen it with commits from Dennis as well. Some boring whitespace issue, most likely.

If you know what caused it and can fix it easily, I'd appreciate it, so that we can keep our history nice. If not, no sweat; I'll just merge.

lesserwhirls:
Yeah, I can fix that. I had IntelliJ reformat the code without thinking,
which I do on all new files I create. I'll update here in a bit.

lesserwhirls:
much better :-)

cwardgar:
Lovely.

PR #711, opened 2016-12-09
Allow debug flags to be set at TDS startup

lesserwhirls:
Set debug flags using -Dtds.debug.flags="flag1 flag2 ... flagN"
(example: -Dtds.debug.flags="H5iosp/read H5iosp/filePos")

### DIFF
--- a/cdm/src/main/java/ucar/nc2/util/DebugFlagsImpl.java
+++ b/cdm/src/main/java/ucar/nc2/util/DebugFlagsImpl.java
@@ -71,4 +71,8 @@ public class DebugFlagsImpl implements DebugFlags {
     map.put(flagName, value);
   }
 
+  public String getSetFlags() {
+    return map.keySet().toString();
+  }
+
 }
--- a/cdm/src/main/java/ucar/nc2/util/DebugFlags.java
+++ b/cdm/src/main/java/ucar/nc2/util/DebugFlags.java
@@ -53,4 +53,10 @@ public interface DebugFlags {
    */
   void set(String flagName, boolean value);
 
+  /**
+   *  Return the string representing the current debug flag(s) set. Flags can be either
+   *  or false - they just need to be set.
+   */
+  String getSetFlags();
+
 }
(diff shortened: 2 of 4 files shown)


lesserwhirls:
Let's try this @cwardgar

lesserwhirls:
I think this means we could remove all the manual set debug flag calls from our tests and clean-up the output by default...if a test fails, then we could turn on specific debug flags using the new property to get more info.

dopplershift:
Are things smart enough that you could have the debug dumped to a log and then just cat that log when the test fails? Just wondering if there's a way to eliminate double-running of test.

DennisHeimbigner:
If I read this correctly, the classes for which this would work
is hardwired into TdsInit. Correct? Is there any way to use
reflection to generalize this?

cdm/src/main/java/ucar/nc2/util/DebugFlagsImpl.java
```diff
@@ -71,4 +71,9 @@ public void set(String flagName, boolean value) {
     map.put(flagName, value);
   }
 
+  public String getSetFlags() {
+    String flags = map.keySet().toString();
+    return (flags.contains("${tds.debugflags}")) ? "" : flags;
```
cwardgar: What does this do?

lesserwhirls: It's cruft from before I set the default property value in TdsContext. I'll update to remove it.

lesserwhirls:
@DennisHeimbigner - I think it's possible to use reflection, but this was pretty straightforward. I think we'd still need some class info in `tdsInit` though. I think it would be a good maintenance thing to do, and might require some interface changes to make sure all calls were consistent. At this point, I'm just happy that end users will be able to turn on debug flags on their systems to report info back to us, rather than sending huge files back or running a custom built war with flags turned on.

lesserwhirls:
@dopplershift - unfortunately, things are not that smart, although I am sure gradle would be able to handle the logic.

cwardgar:
In addition to reflection, another way to generalize debug flag setting is to Springify `cdm`. Then, Spring could inject a `DebugFlagsImpl` instance into any object that needed it. But long term, I think the best way to handle this problem is to just write the debug messages to dedicated debug loggers. Then those messages could be enabled/disabled by editing the logger configuration file (log4j2.xml or similar).

Anyway, this looks like a good straightforward approach for now.

PR #714, opened 2016-12-15
Upgrade a few dependencies

lesserwhirls:
migrate to Thymeleaf 3.x


lesserwhirls:
Running on Jenkins, too. Let's hold off until we see how that run goes.

cwardgar:
Can you add some comments that explain which old versions you're overriding and which first level dependencies are dragging them in?

lesserwhirls:
Ok, this one is updated.

cwardgar:
Reorganized a bit. Travis doesn't want to run right now though.

PR #723, opened 2016-12-28
Close in define mode (5.0.0)

cwardgar:
Same as #722, but for the 5.0.0 branch. This PR includes tests that aren't being added to 4.6 (the Gradle builds have diverged too much).


lesserwhirls:
👍 Thanks @cwardgar!

Issue #695, opened 2016-12-01
readDataSlice(int rt, int e, int t, int z, int y, int x) in the geogrid.java doesnot check the unsigned attribute


yuanho:
you can find the dataset here:

The data is at ftp ftp.ssec.wisc.edu in
pub/incoming/A2015145.210[0-3].nc.

lesserwhirls:
Hi @yuanho - it looks like this didn't make it in 4.6.7. However, I'll get a fix in today, but I'll need an example file. The data file is no longer available on the ssec ftp server - do you have a copy?

dopplershift:
FYI, you can zip/gzip netcdf files and attach them to issues--that way they're always available in the matching issue.

lesserwhirls:
test data and ncml agg can be found in [data.zip](https://github.com/Unidata/thredds/files/686292/data.zip)

PR #726, opened 2017-01-05
fixes unsigned byte aggregations

lesserwhirls:
fixes Unidata/thredds#695


lesserwhirls:
Will port to 5.0 if approved.

cwardgar:
Looks good. I wonder if this bug even exists on 5.0.0.

PR #727, opened 2017-01-05
fix ncep grib2 link

lesserwhirls:


cwardgar:
Tough PR to review, but I assume you're just using v17.0.0 of the NCEP tables instead of v13.0.0?

Also, this needs a rebase.

cwardgar:
Never mind, I was able to "Update branch" with the latest changes from master. I guess that's a new feature we have now because of the "Allow edits from maintainers." option. Anyway, it resulted in a [merge commit in your repo](https://github.com/lesserwhirls/thredds/commit/378c4ab6c5d2cf446a802ba09da757ffbde9318b). We can talk about whether that's a git strategy we want to adopt going forward, but it certainly was nice to not have to wait for you to rebase.

cwardgar:
Interestingly, the merge commit doesn't show up in the [Unidata/master history](https://github.com/Unidata/thredds/commits/master). Compare with the [lesserwhirls/ng2 history](https://github.com/lesserwhirls/thredds/commits/ng2). Also the "fix ncep grib2 link" and "fixes unsigned byte aggregations" commits are in different orders. I wonder what kind of git magic is happening there.

dopplershift:
That's so weird; I don't think it's git magic so much as github, but I could be mistaken.

lesserwhirls:
Ah git and github...sometimes they just seem to know what we need done and they do it, in it's somewhat mysterious ways *laughs*

lesserwhirls:
and yes, I was simply moving tables around with one small code change (update the path to the grib tables).

PR #728, opened 2017-01-05
Grib fixes applied to 5.0

lesserwhirls:
Cherry-pick of commits related to GRIB1/2 tables from NCEP and WMO, as made on 4.6.


lesserwhirls:
includes a cherry-pick of Unidata/thredds#727, so that one needs to be approved before this is merged.

PR #729, opened 2017-01-06
Add geocsv to GridAsPoint form

lesserwhirls:


lesserwhirls:
Again...why are the controls for GridAsPoint output format choice, from the TDS, in the CDM module? No idea...but bonus points for the comment:

> // LOOK this is wrong - should be using SupportedOperation class or something

    ¯\_(ツ)_/¯

cwardgar:
Baffling.

PR #734, opened 2017-01-09
Update release procedure. Mostly git stuff.

cwardgar:
Ugh, we're up to 32 steps.

### DIFF
--- a/docs/internal/release.md
+++ b/docs/internal/release.md
@@ -9,22 +9,28 @@
    - The instructions for the two are very similar.
 
 2. Ensure that there are no uncommitted changes.
+   - `git checkout master`
+   - `git status`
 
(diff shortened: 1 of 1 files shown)


lesserwhirls:
Nice work. I wonder how much we could automate (I know you have some notes at the end about that). Does gradle have a hamster plugin?

cwardgar:
In theory, everything can be automated, including deployment to production. But that's probably not realistic. If we could just get all of our artifact publishing automated, that would be huge though.

dopplershift:
I love animating all the things--and having package publishing automated has been great in my experience. It really makes cutting a release no big deal.

Automated deployment to production, at least from a code repo, is a colossal mistake IMO. Deploying to a test server would be awesome, though.

PR #739, opened 2017-01-14
Catch NoSuchMethodError in CoverageDatasetFactory

lesserwhirls:
When trying to load a netCDF file as via FeatureDatasetFactoryManager.open(...), the openGrib method thows a NoSuchMethodError. openGrib is called in the open method as a test to see if the dataset is a Grib file, and because the NoSuchMethodError wasn't being caught, the whole thing stops and the FeatureDatasetFactoryManager returns a null dataset with an error message in the Formatter. Adding this error to the catch allows FeatureDatasetFactoryManager to do the right thing and see that the file isn't a Grib file.

### DIFF
--- a/cdm/src/main/java/ucar/nc2/ft2/coverage/CoverageDatasetFactory.java
+++ b/cdm/src/main/java/ucar/nc2/ft2/coverage/CoverageDatasetFactory.java
@@ -1,40 +1,8 @@
 /*
- * Copyright 1998-2015 John Caron and University Corporation for Atmospheric Research/Unidata
- *
- *  Portions of this software were developed by the Unidata Program at the
- *  University Corporation for Atmospheric Research.
- *
- *  Access and use of this software shall impose the following obligations
- *  and understandings on the user. The user is granted the right, without
- *  any fee or cost, to use, copy, modify, alter, enhance and distribute
- *  this software, and any derivative works thereof, and its supporting
- *  documentation for any purpose whatsoever, provided that this entire
- *  notice appears in all copies of the software, derivative works and
- *  supporting documentation.  Further, UCAR requests that the user credit
- *  UCAR/Unidata in any publications that result from the use of this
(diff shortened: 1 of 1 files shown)


lesserwhirls:
Not done yet...ugh...

lesserwhirls:
Ok, ready for review.

cdm/src/main/java/ucar/nc2/ft2/coverage/CoverageDatasetFactory.java
```diff
@@ -126,17 +95,36 @@ static public FeatureDatasetCoverage open(String endpoint) throws IOException {
...
+        boolean notGribBuried = e.getClass().equals(InvocationTargetException.class) &&
+                                e.getCause() != null &&
+                                e.getCause().getClass().equals(noGrib);
 
-    } catch (IllegalAccessException | IllegalArgumentException | ClassNotFoundException | NoSuchMethodException e) {
-      //e.printStackTrace();
-      return Optional.empty(NO_GRIB_CLASS);
+        if (notGribTopLevel || notGribBuried) {
+          return Optional.empty(NO_GRIB_CLASS);
+        } else if (e.getClass().equals(InvocationTargetException.class) && e.getCause() != null && e.getCause().getClass().equals(noGrib)) {
```
cwardgar: Isn't this unnecessary now that you have `notGribBuried`?

cwardgar: I'm referring to the `else if` statement. That wasn't as clear as I thought it would be in the review that GitHub created.

lesserwhirls: Yep, that was for sure a mistake. Fixed now. Thanks @cwardgar!

lesserwhirls:
Yeah, it felt kludgy to refactor it a bit but not address the original intention.

PR #740, opened 2017-01-14
add signature to NetcdfDataset

lesserwhirls:


lesserwhirls:
Sorry about the odd diff. The javadoc is same as an older signature, which probably screwed the diff up. Basically creates a new signature for `NetcdfDataset.acquireDataset` that uses the default `enhanceMode`.

lesserwhirls:
Ok, diff is better now...not perfect, but much better.

cdm/src/main/java/ucar/nc2/dataset/NetcdfDataset.java
```diff
@@ -463,6 +463,21 @@ static private CoordSysBuilderIF enhance(NetcdfDataset ds, Set<Enhance> mode, Ca
   }
 
   /**
+   * Same as openDataset, but file is acquired through the File Cache, with defaultEnhanceMode,
+   * without the need of settinging the enhanceMode via the signature.
```
cwardgar: Typo.

cwardgar:
Nevermind, fixed it myself in the GitHub UI. Slick!

PR #741, opened 2017-01-18
misc. changes from dap4 work

DennisHeimbigner:
As a result of the dap4 work, it was necessary/desirable
to make small mods to other, non-dap4, code. This changeset
encapsulates them.
1. add nc_inq_var to Nc4prototypes.java
2. fix naming bug for structs in Nc4Iosp.java
3. Add annotation map to CDMNode
5. Add ability to suppress certain special/reserved attributes
   (primarily Attribute.java, but also the writeCDL methods elsewhere).
6. Fix the parsing of string dimension lists in Dimension.java
7. Add another constructor for Structure; will be used by dap4 code
8. Add another constructor for Variable; will be used by dap4 code
9. Add another option to NcDumpW: -datasetname to force
   the use of a particular dataset name; convenient for avoiding
   spurious text differences in comparing two cdl files.
10. Add a new IOprovider registration function: registerIOproviderPreferred
    to NetcdfFile.java. This function allows one to insert
    an IOSP in the registration list just in front of another IOSP.
    Primarily used to support overriding HDF5 IOSP with Netcdf4 IOSP.
11. Add a backslash escape function to EscapeStrings.java.
12. Add "dap4" as a possible entry in a url fragment. This parallels
    the existing "dap2" case (DatasetUrl.java).

### DIFF
--- a/cdm/src/main/java/ucar/nc2/Attribute.java
+++ b/cdm/src/main/java/ucar/nc2/Attribute.java
@@ -54,7 +54,14 @@ import java.util.Map;
  * @author caron
  */
 
-public class Attribute extends CDMNode {
+public class Attribute extends CDMNode
+{
+
+  static final String SPECIALPREFIX = "_";
+  static final String[] SUPPRESS = new String[]{
+          "_NCProperties", "_DAP4_Little_Endian", "_edu.ucar"
+  };
+
 
   /**
    * Turn a list into a map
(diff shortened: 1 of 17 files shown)


cwardgar:
Looks good. I think `registerIOProviderPreferred` is still a good idea, but keep in mind that #558 exists, and its sole purpose is to provide a way to read NetCDF-4 data with `Nc4Iosp` instead of `H5iosp`.

PR #743, opened 2017-01-19
Merge master into 5.0.0

cwardgar:
Gnarly.

PR #744, opened 2017-01-22
Merge master into 5.0.0 (take 2)

cwardgar:
This supersedes #743, which I botched. I think I got it right this time.


cwardgar:
Yar, same error. I'm starting to suspect that this is some GitHub weirdness, because I can rebase the branch onto `5.0.0` from the command line.

Maybe these kind of PRs require a merge commit instead of a rebase? Can you re-enable that option and give it a shot, @lesserwhirls ?

lesserwhirls:
Looks like that worked. Thanks @cwardgar!

PR #733, opened 2017-01-09
Tests no longer target the TDS at remotetest.unidata.ucar.edu

cwardgar:
* Instead, they target the embedded TDS that's launched in `:it:test`. Some test classes needed to be moved into `:it` to make that possible. Had to move around some datasets and tweak some paths as well to keep tests passing.
* `TestSSH`, `TestTomcatAuth`, and `TestByteRange` were improved to allow execution within Jenkins and Travis. Previously, they had to be run manually from IntelliJ.
* Removed `NeedsExternalResource` annotation from every test I could. More pruning to come in the future.
* Added `NeedsCdmUnitTest` to a couple of classes that (sneakily) require it.
* `:it:appRun` no longer requires `:it:processTestResources` to be run first.

With this PR, we no longer have any dependence on remotetest.unidata.ucar.edu that I'm aware of. @DennisHeimbigner I'd like you to take a look at my changes to `TestSSH`, `TestTomcatAuth`, and `TestByteRange` and tell me if they're okay.

PR #745, opened 2017-01-24
Fix test failures stemming from #733

cwardgar:
In #733, I added back some tests to our suite that had been excluded. They belonged to the `thredds.tds` package in `:it`. There were two classes with failures:
* Fixed `TestRestrictDataset` by giving it the proper credentials, and adding the necessary "tiggeData" role to the embedded Tomcat's tomcat-users.xml.
* `TestWaveModel` couldn't be fixed because the files it needs are not present in `cdmUnitTest/`. So I `@Ignore`-ed it. What happened here?

Other changes:
* Added `HTTPSession.clearGlobalCredentials()`, which our tests must call any time they set global credentials. Otherwise, they're ruining the global state for other tests.
* Refactored `TestAdminDebug`. A recent commit caused it to be unnecessarily complex and `testOpenHtmlFail()` wasn't even doing what it was supposed to be doing.
* `TestTomcatAuth` unsets global credentials after it's finished. Also, converted console output to logger.

I gave this a go [on Jenkins](https://jenkins-aws.unidata.ucar.edu/job/cwardgar/43/testReport/) and it eliminated the new failures we've been seeing [on 5.0.0](https://jenkins-aws.unidata.ucar.edu/job/thredds5.0/265/testReport/).

@DennisHeimbigner I'm mostly modifying your code again here. Give it a look, please.

### DIFF
--- a/cdm-test/src/test/java/ucar/nc2/iosp/hdf5/TestN4reading.java
+++ b/cdm-test/src/test/java/ucar/nc2/iosp/hdf5/TestN4reading.java
@@ -56,7 +56,7 @@ public class TestN4reading {
 
   @Test
   public void testGodivaFindsDataHole() throws IOException, InvalidRangeException {
-    // this pattern of reads from godiva is finding a data hole - missing data where therre shouldnt be any
+    // this pattern of reads from godiva is finding a data hole - missing data where there shouldn't be any
     Section[] sections = {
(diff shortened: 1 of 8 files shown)


cwardgar:
@DennisHeimbigner Conflict resolved. This one is ready when you are.

PR #746, opened 2017-01-30
Upgraded log4j and slf4j to latest version

lesserwhirls:
Fixes odd errors encountered on motherlode

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -88,7 +88,7 @@ libraries["52n-xml-sampling-v20"] = "org.n52.sensorweb:52n-xml-sampling-v20:${ve
 
 ////////////////////////////////////////// Logging //////////////////////////////////////////
 
-versions["slf4j"] = "1.7.7"
+versions["slf4j"] = "1.7.22"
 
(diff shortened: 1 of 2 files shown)


lesserwhirls:
Thanks @cwardgar!

cwardgar:
Sure. I'm gonna see if I can get Codecov working as well. We've been using [Coveralls](https://coveralls.io/github/Unidata/thredds), but it gets stuck showing spurious builds for some reason.

dopplershift:
💯  to codecov. Had few problems with it for the Python stuff--and only when doing some advanced filtering.

cwardgar:
My concern is that if we upload the coverage at the end of a Travis run (which seems to be what Codecov and the GitHub PR integration expect), it'll only show partial coverage, since not all of our tests run on Travis. We could do it on Jenkins, of course, but I don't think we'll get the feedback in our PRs.

dopplershift:
Not sure how that's a codecov-specific problem...

cwardgar:
It's not really. We'll just have to disable the codecov bot.

dopplershift:
Add a `codecov.yml` file to the repo with:
```yml
comment: off
```

PR #748, opened 2017-01-31
Add new link for the FMRC poster

lesserwhirls:


lesserwhirls:
Very useful .pdf file for the docs. It went missing when John's home directory was archived.

lesserwhirls:
Need to update just a few more links...opps

lesserwhirls:
Ok, that's got it. Should be good to go.

cwardgar:
@lesserwhirls Oops, meant to "Rebase and Merge" this, not "Create a merge commit". Is there a way that you can make the former the default, while still leaving the latter enabled?

lesserwhirls:
It does not look like it. I disabled "Create a merge commit" for now, and will enable it as needed...hopefully not often *laughs*

PR #751, opened 2017-02-02
Refactored Gretty configuration

cwardgar:
* Eliminated workarounds for Gretty bugs that used to live in it/build.gradle. Instead, I've fixed those bugs directly in v1.0.3 of [my Gretty fork](https://github.com/cwardgar/gretty/commits/master).
* The remaining config has largely been centralized into a new script plugin: gradle/any/gretty.gradle.
* For projects (targets) that overlay other projects (sources) onto themselves for integration testing, their Gretty config is taken from the sources. For example, ':it' takes its Gretty config from ':tds'. The goal was to minimize duplicated code.
* Configured :tds, :dtswar, and :d4ts in such a way that they can be run outside of a testing environment. Simply do something like: `./gradlew :tds:appRun`
  - To that end, moved authorization resources from ':it' to ':tds' so that the TDS can be run from the command-line with HTTPS and admin support.
* Upgraded from Gradle v3.1 to v3.3.
* Fixed Gradle deprecation warnings:
  - Used `Task.doLast` instead of `Task.leftShift`
  - Used `TaskInputs.file` instead of `TaskInputs.source`
* Moved some System-properties-related logic out of [any/root]/testing.gradle and into properties.gradle.
* System properties set on the Gradle process are propagated to all Test JREs and Gretty embedded webapps.


lesserwhirls:
Not having a ton of gradle experience and pretty much zero expertise, I say this looks good. If it builds and runs the tests, and they pass, then hey, how bad can it be?

🥇 👍 💯

cwardgar:
Actually, I intend to write a wiki page for Gretty that should give you and Dennis more expertise. Not that it'll make it much easier to review this PR though.

lesserwhirls:
👍

PR #552, opened 2016-05-16
take into account 401 and 403 HTTP codes. Fix Unidata/thredds#551

cofinoa:
The 403 HTTP response status code is considered as IOException like 401 HTTP response. 
Also the exception message include the return status code.

PR #754, opened 2017-02-07
Convert SC_UNAUTHORIZED (401)   return codes to SC_FORBIDDEN (403)

DennisHeimbigner:
re: issues 551
supercedes pr: 552

1. Convert all generation of SC_UNAUTHORIZED (401)
   return codes to SC_FORBIDDEN (403).
2. Fix tests to accept SC_FORBIDDEN as well as SC_UNAUTHORIZED

I think issue 551 is still outstanding.

### DIFF
--- a/it/src/test/java/thredds/tds/TestRestrictDataset.java
+++ b/it/src/test/java/thredds/tds/TestRestrictDataset.java
@@ -91,7 +91,7 @@ public class TestRestrictDataset {
   @Test
   public void testFailNoAuth() {
     String endpoint = TestWithLocalServer.withPath(path);
-    logger.info("testRestriction req = '%s'", endpoint);
+    logger.info(String.format("testRestriction req = '%s'", endpoint));
 
     try (HTTPSession session = HTTPFactory.newSession(endpoint)) {
       HTTPMethod method = HTTPFactory.Get(session);
@@ -110,7 +110,7 @@ public class TestRestrictDataset {
   @Test
   public void testFailBadUser() {
     String endpoint = TestWithLocalServer.withPath(path);
-    logger.info("testRestriction req = '%s'", endpoint);
+    logger.info(String.format("testRestriction req = '%s'", endpoint));
 
     try (HTTPSession session = HTTPFactory.newSession(endpoint)) {
       session.setCredentials(new UsernamePasswordCredentials("baadss", "changeme"));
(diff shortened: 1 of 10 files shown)


cwardgar:
Many of the `printf` statements are wrong, e.g.
```java
System.out.printf("Should return HttpStatus.SC_UNAUTHORIZED|HttpStatus.SC_FORBIDDEN err=%s%n", e.getMessage());
```
should be:
```java
System.out.printf("Should return %s or %s; err = %s%n", HttpStatus.SC_UNAUTHORIZED, HttpStatus.SC_FORBIDDEN, e.getMessage());
```

Even better would be to convert those STDOUT messages to logger messages.

DennisHeimbigner:
No, those printf's are correct. We only have a set of possible accepted
return values so we can only list the set since no element of
the set was returned.

You are correct about the logger; let me fix that.
=Dennis

cwardgar:
So in the log, you want to see:
```
Should return HttpStatus.SC_UNAUTHORIZED or HttpStatus.SC_FORBIDDEN; err = <MESSAGE>
```
instead of:
```
Should return 401 or 403; err = <MESSAGE>
```
It's fine if you do; just making sure.

Also, slf4j uses a different formatting anchor than printf. See https://www.slf4j.org/api/org/slf4j/helpers/MessageFormatter.html

it/src/test/java/thredds/tds/TestRestrictDataset.java
```diff
@@ -184,12 +175,12 @@ public void testSuccess() {
...
     String endpoint = TestWithLocalServer.withPath(path);
-    System.out.printf("testRestriction req = '%s'%n", endpoint);
+    logger.info("testRestriction req = '%s'", endpoint);
     try {
       try (HTTPMethod method = HTTPFactory.Get(endpoint)) {
         int statusCode = method.execute();
-        if (statusCode != 401 && statusCode != 403) {
-          System.out.printf("statuscode=%d expected 401 or 403%n", statusCode);
+        if (statusCode != HttpStatus.SC_UNAUTHORIZED && statusCode != HttpStatus.SC_FORBIDDEN) {
+          logger.error("statuscode=%d expected HttpStatus.SC_UNAUTHORIZED or HttpStatus.SC_FORBIDDEN; actual=%d", statusCode);
```
cwardgar: Format contains two anchors but only one value is given.

DennisHeimbigner:
That will teach me to do a blanket replace.
=Dennis

cwardgar:
Looks good. Does this address all three of #547, #552, and #625?

DennisHeimbigner:
We need to apply 552, but to same base as this (5.0.0).

PR #757, opened 2017-02-17
Projection fixes

barronh:
I have modified the WRFConvention.java and M3IOConvention.java files to better represent these projections. 

commit 79c8f0e6dae96bb85c3b22b2f911482dd8eaa0a6
    Fixing x/y Coordinate variables
    x/y coordinates for rasters should be cell centered; edges of cells are described by the x_bounds or y_bounds as described in Section 7 of the metdot files for which this treatment is not right, but they are the exception to the rule and a smaller number of files.
    Without this fix, visualization softwares IDV and Panoply will display rasters spatially shifted half a grid cell to the southwest.

commit a5fef4eea91544c899c8818f20dcc30ff961e81e

    Improving WRF Projections
    1. Adding earth radius 6370 consistent with WRF documentation. Addresses issue 753
    2. Updating reference latitude for Mercator projections to improve treatment of nested grids. Addresses issue 752


lesserwhirls:
Thank you so much for this PR!

PR #767, opened 2017-03-04
safely combine coordinate attrs in OPeNDAP

lesserwhirls:
When dealing with GOES-16 data, I found that the `_CoordinateAxis` attribute, was getting duplicate entries added to it when reading the dataset through OPeNDAP. For example:

`_CoordinateAxes = "band_id band_wavelength t y x y x ";`

when it should be

`_CoordinateAxes = "band_id band_wavelength t y x";`

This PR fixes the duplication issue by safely combining the attributes listed by the CF attribute `coordinates` and the attribute `_CoordinatesAxis`.

This also happens in 5.0, so if this is approved, I will make a new PR for 5.

### DIFF
--- a/opendap/src/main/java/ucar/nc2/dods/DODSNetcdfFile.java
+++ b/opendap/src/main/java/ucar/nc2/dods/DODSNetcdfFile.java
@@ -32,6 +32,8 @@
  */
 package ucar.nc2.dods;
 
+import com.google.common.base.Joiner;
+
 import net.jcip.annotations.NotThreadSafe;
 import ucar.nc2.constants.CF;
 import ucar.nc2.util.EscapeStrings;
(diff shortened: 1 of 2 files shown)


lesserwhirls:
When dealing with GOES-16 data, I found that the `_CoordinateAxis` attribute, was getting duplicate entries added to it when reading the dataset through OPeNDAP. For example:

`_CoordinateAxes = "band_id band_wavelength t y x y x ";`

when it should be

`_CoordinateAxes = "band_id band_wavelength t y x";`

This PR fixes the duplication issue by safely combining the attributes listed by the CF attribute `coordinates` and the attribute `_CoordinatesAxis`.

This also happens in 5.0, so if this is approved, I will make a new PR for 5.

lesserwhirls:
Just pushed changes for Codacy happiness. Should be ready for review when it passes.

DennisHeimbigner:
You could consider replacing the guava join with the second form of java.lang.String#join function here:
https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#join-java.lang.CharSequence-java.lang.Iterable-

lesserwhirls:
@DennisHeimbigner - I plan on using the String join function for 5.0, but 4.6.x still supports Java 7. I think the String join method is Java 8 only.

Issue #772, opened 2017-03-07
Time unit  "days since 19500101T000000Z"


jowilk:
I have NetCDF files with a time unit defined as "days since 19500101T000000Z" and although it seems to be valid for the udunits2 library, the NetCDF Java library does not support it (using ucar.nc2.units.DateUnit) .

I can not change the original NetCDF files, would it be possible to support this type of unit in the java library and also in the Thredds WMS server?

Thank you

Jonathan

drf5n:
Look into using NCML in the catalog to modify attributes per http://www.unidata.ucar.edu/software/thredds/v4.3/tds/tutorial/NcML.htm 

Maybe a chunk in the catalog element like this would work:

      ...
      <netcdf xmlns="http://www.unidata.ucar.edu/namespaces/netcdf/ncml-2.2" 
              location="/machine/tds/workshop/ncml/example1.nc">
        <variable name="time">
           <attribute name="units" value="days since 1950-01-01 00:00:00 -00:00"/>
         </variable >
       </netcdf>
    ...

jowilk:
Thank you for your reply. I can actually use the NCML with Thredds but I also need to be able to exploit the files directly from the java library (and there without using the NCML).

Would you be able to tell me whether this form of time unit is correct?

In any case this solution works well with Thredds and I can always handle particular cases in my java code.

lesserwhirls:
Greeings @jowilk - while my first thought was "wow, that's an ugly time format!", you are correct that udunits-2 recognized the string:

~~~bash
sarms:$ ./udunits2 
You have: 2 days since 19500101T000000Z  
You want: 
    172800 s
~~~

I'll try to get this working in our next release. Thanks for the report!

jowilk:
Yes this is not the most readable format but it seems it's also part of ISO8601.
Thank you for the interest in this request.

drf5n:
I use that format in filenames--it's ISO8601 compliant, shell compatible, no special characters, fixed length, sorts sanely, and is pretty easy to parse.

cwardgar:
This was fixed by #773, right?

lesserwhirls:
@drf5n Maybe it's a bug in `udunts2` when using a number in front of the time unit, because this works:

~~~bash
You have: days since 19500101T000000Z
You want: days since 19490101T000000Z
    1 days since 19500101T000000Z = 366 (days since 19490101T000000Z)
[...]

PR #773, opened 2017-03-09
Add support for basic iso format

lesserwhirls:
Add support for the basic iso format, which is like the iso string we all know
and love, but without any date or time delimiters (that is, no '-' or ':'
characters).

Fixes Unidata/thredds#772

### DIFF
--- a/cdm/src/main/java/ucar/nc2/time/CalendarDateFormatter.java
+++ b/cdm/src/main/java/ucar/nc2/time/CalendarDateFormatter.java
@@ -205,7 +205,7 @@ public class CalendarDateFormatter {
   }
 
   //                                                   1                  2            3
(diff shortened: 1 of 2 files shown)


cwardgar:
In the test, can you replace those prints to stdout with a logger?

lesserwhirls:
OMG why didn't I do that - burn `System.out.printf`, :fire:. Is there a specific logger we are using for tests?

cwardgar:
Something like:
```java
private static final Logger logger = LoggerFactory.getLogger(TestCalendarDateFormatter.class);
```
is good enough for now.

lesserwhirls:
Loggers, it be, and a rebase on master, ta boot!

lesserwhirls:
oops...got the log messages backwards on `claimBad`...should be good now.

Issue #766, opened 2017-03-04
GOES-16 issues


lesserwhirls:
There are two issues with the new GOES-16 data and netCDF-Java that I have found.

1) the attribute `cdm_data_type` is set to `Image`
2) the data variables have scalar `coordinates` variables listed in the `coordinates` attribute

The first issue is a problem because, well, netCDF-Java does not have `cdm_data_type` of `Image`. That is, there isn't a Scientific Feature Type for Image that's documented, and I'm not sure that was every fully thought out or implemented anywhere (see this [comment](https://github.com/Unidata/thredds/issues/660#issuecomment-255258401) from @ethanrd). When reading the data from disk using `ToolsUI` `FeatureType -> Grids`, everything appears to be ok. So, while I say this is a "problem", it does not appear that netCDF-Java cares all that much and just says "ah, heck, you can be a grid." So forgiving, it is.

The second issue isn't actually a problem wth the data itself, as scalar coordinates in general are ok. The problem is that when reading through `OPeNDAP`, a  `_CoordinateAxis` attribute is added containing the values of the `coordinates` attribute. On its own, this is no big deal I suppose. However, there is a special case in `ucar.nc2.dataset.CoordSysBuilder` that ends up identifying the GOES-16 data variables as coordinate variables because of the scalar coordinates listed in `_CoordinateAxis`...and of course this special case was put in to handle `grib` data. 🤢 Note that this exists in both 4.6.x and 5.0.

Here is the coresponding comment in the code from @JohnLCaron:

~~~java
      /* WTF? JC 7/15/2010
      /* Old GRIB code basically has a bug in it for lat/lon coordinates
         String latLonCoordSys;
            :_CoordinateAxes = "time lat lon";
         this is a Coordinate System Variable getting misidentified as a data variable, because its doesnt qualify as a CoordinateSystem
           see http://www.unidata.ucar.edu/software/netcdf-java/reference/CoordinateAttributes.html
         but it must be a CoordinateSystem because the  _CoordinateAxes dont fit the variable.

         Problem:
         For some data, some axes are being constructed with anon dimensions, so that isCoordinateAxisForVariable is failing and
         so the data value is being tagged as a CoordinateSystem. Could insist that all variable fail ??
         */
~~~

[...]

PR #775, opened 2017-03-10
Remove special case code from CoordSysBuilder

lesserwhirls:
Update copyright block

Fixes Unidata/thredds#766 on the 5.0.0 branch


lesserwhirls:
There is more that can be nuked from here since we are on 5.0.0...suggestions?

cwardgar:
I added a commit that removed some cruft. I took a careful look at the commented-out code that I nuked and as far as I can tell, it's all either no-longer valid or not useful should it be re-enabled.

cwardgar:
I originally thought that a lot more code could be removed, but it turns out that I was misreading the Jenkins coverage report. Still, any removal of cruft is a win, as it reduces our cognitive load when trying to understand the class.

PR #777, opened 2017-03-12
Make access to the netcdf-c library be synchronized 

DennisHeimbigner:
Upgrade to net.java.dev.jna:jna:4.2.2
so that we can use Native.synchronizedLibrary()
to force synchronized (single-thread) access to
 the netcdf-c library.

This is effectively a back-port of what already was done for 5.0.0


lesserwhirls:
Thanks @DennisHeimbigner!

cwardgar:
So does this obviate the need to build hdf5 with `--enable-threadsafe` for THREDDS? That would be big, as it would allow users to grab netcdf from a package manager, instead of needing to build it themselves.

lesserwhirls:
That's the hope...also the hope that this fixes the issues on motherlode (although it does not help the issues building the dang thing, but nothing in Java that will help that!).

DennisHeimbigner:
It is my belief that it does make it unnessary to
build thread safety into HDF5 (or netcdf-c, for that matter).
But as I said to Sean, I have no good way to test this (yet).
=Dennis

Issue #17, opened 2012-11-30
nc4_64_dll.zip does not contain deps


toelen:
Hi,

the webpage http://www.unidata.ucar.edu/software/netcdf-java/reference/netcdf4Clibrary.html links to  nc4_64_dll.zip and mentions the deps directory. This directory is not present int he zip file however, and searching for the correct version of all dlls is quite tricky. Is it possible to put the deps folder in the zip, or at least put a link on there to the correct downloads?

http://www.unidata.ucar.edu/software/netcdf/win_netcdf/nc4_64_dll.zip

Thanks

Issue #537, opened 2016-04-18
Parameter-provider method is always executed, even if its test class is excluded


cwardgar:
JUnit always evaluates a `Parameterized` test class's `@Parameters` method, even when its actual tests will be skipped due to a `@Category` exclusion. This is problematic for thredds because many of those parameter-provider methods do expensive and unnecessary operations, such as scanning all or part of `TestDir.cdmUnitTestDir`. They're SUPER spammy in our log output as well.

I'd like to find a test runner that doesn't invoke the parameter-provider unless necessary. Junit's included `Parameterized` class doesn't fit the bill, but [JUnitParams](https://github.com/Pragmatists/JUnitParams) might. Spock might work as well.

cwardgar:
This is a particular problem on Travis, where `cdmUnitTest/` is not available, but those parameter-provider methods are run anyway. The result is a bunch of spurious error messages like:
```
ucar.nc2.dataset.TestOpenWithEnhanceP STANDARD_OUT
    ---------------Reading directory /home/travis/build/JohnLCaron/thredds/build/NO/unidata.testdata.path/FOUND/cdmUnitTest/conventions
    ---------------INVALID /home/travis/build/JohnLCaron/thredds/build/NO/unidata.testdata.path/FOUND/cdmUnitTest/conventions

ucar.nc2.dataset.TestOpenWithEnhanceP STANDARD_ERROR
    java.io.FileNotFoundException: Cant open /home/travis/build/JohnLCaron/thredds/build/NO/unidata.testdata.path/FOUND/cdmUnitTest/conventions
    	at ucar.unidata.test.util.TestDir.actOnAll(TestDir.java:344)
    	at ucar.unidata.test.util.TestDir.actOnAllParameterized(TestDir.java:311)
    	at ucar.nc2.dataset.TestOpenWithEnhanceP.getTestParameters(TestOpenWithEnhanceP.java:58)
```

Note that `TestOpenWithEnhanceP` is marked with `@Category(NeedsCdmUnitTest.class)`, and is thus excluded on Travis. Those messages are still printed.

cwardgar:
It turns out that JUnitParams has the same behavior as JUnit: the parameter-provider method is always executed, even if its test class is excluded.

A simple solution that just occurred to me is to check the directory pointed to by `TestDir.cdmUnitTestDir`**in the parameter-provider method** and return an empty list if it doesn't exist.

PR #782, opened 2017-03-18
Bring the Java DAP4 and the (new) netcdf-c DAP4 into consistent operation.

DennisHeimbigner:
The primary goal of this PR is to bring the Java DAP4 and the (new) netcdf-c
DAP4 into consistent operation.

This pr has some currently unresolved issues.
1. It looks like we are not using the latest netcdf-c library (on Jenkins) and
   this apparently is causing some test to fail on Jenkins that work locally.
2. There is at least one test --
   ucar.nc2.jni.netcdf.TestNc4JniReadCompare.compareDatasets[/share/testdata/cdmUnitTest/formats/netcdf4/tst/tst_enums.nc] --
   that I could fix if I had access to the cdmunittests directory;
   it is the result of some changes made in this pull request.

From the commit log:
1. Fix major bug in computing serialization checksums
   because of endianness issues.
2. Fix some more bugs in dap4 detected as a
   result of testing dap4 against dap4 in netcdf-c.
3. Refactor to convert DapStructure to be a
   subclass of DapType instead of DapVariable;
   effectively make all variables consistent
   wrt having a basetype. This necessitated
   some significant refactoring of some dap4 code.
4. Add Suppress category to tag tests that should
   be temporarily suppressed because something
   they use (e.g. remotetest.unidata.ucar.edu/d4ts)
   is unavailable. Assumption is that this fixable
   and this category can be removed when the fix is in.
5. Add ability for the client to control the endianness
   of server generated replies: primarily for testing
   since not part of the spec.
6. Significant modification of the handling of enumeration typed
   attributes. This affects at least:
	EnumTypedef.java
	DataType.java
	Attribute.java
	Nc4Iosp.java
7. Modify TestHang to properly check for timed out connections.
8. Modify test cases as needed to conform to the above changes.
9. Misc. cleanup of httpservices
10. Modify HTTPMethod to support testing using Spring servlet mocking
11. cleanup DTSServlet (in opendap) to generate proper return codes;
   currently, it always returns 200.
12. UnitTestCommon- It appears that under
   Jenkins, the Java property "user.dir" is set
   incorrectly for our purposes. In this case, we
   want to use the WORKSPACE environment variable
   set by Jenkins.
13. Modify Attribute.java#writecdl to
   1. (-strict) Properly print the attribute type
      for strings. This is to conform to the netcdf-c practice
      which avoids ambiguity of char attributes vs string attributes.
   2. (-strict) Properly print the enum const name (as a string)
      rather than its integer value; this requires storing the
      EnumTypeDef in the attribute
   3. Modify constructors to allow passing of an EnumTypeDef (try to keep
      maximum back compatibility at the expense of constructor
      proliferation).
   Note that the non-strict case is left as is.
14. Modify writeCDL (with strict == true) to properly write the attribute
   type to conform to the netcdf-c cdl form. Also had to modify some tests
   to use lower case string as the type.
   Affects at least:
	Group.java
	Variable.java
	Struct.java
	Attribute.java
15. Fix some code to conform to enum fixes
	- TestGrid1.java
	- TestByteRange.java
	- Group.java
	- Variable.java
16. Fix some code to use the new attribute constructors (re: #5.3)
	- TestWriteMisc.java
	- H4header.java
	- HdfEos.java
	- H5header.java
17. Modify NCdumpw to allow overriding of the dataset name.
18. Modify TestCEEvaluator to use #17
19. Convert test to use logging instead of System.out/err;
    this was somewhat haphazard, so more to do.
20. Copyright changes (again, haphazard).


lesserwhirls:
Thanks @DennisHeimbigner! Here is the file you are looking for:

[tst_enums.nc.zip](https://github.com/Unidata/thredds/files/852521/tst_enums.nc.zip)

I had to zip it in order to attach it here.

It looks like there are a couple of tests failing on Travis - are those related to this file?

DennisHeimbigner:
Most of those tests seem to be failing on "file not found", which I do not think
is caused by my changes. The TestUserProblems failure is an authentication failure.
The only one I see that is mine is ucar.nc2.jni.netcdf.TestNc4JniReadCompare
and you sent me the data file to try to fix it.

DennisHeimbigner:
When you get a chance, please replace that old tst_enums.nc with the 
attached.
=Dennis

lesserwhirls:
Hi Dennis, the file didn't attach to the issue. You may have to zip it to get it to attach (should be able to drag-and-drop using the github interface)

DennisHeimbigner:
Oops, I attached it to the email and it apparently does not pass thru.
I will attach to this comment.
[tst_enums.zip](https://github.com/Unidata/thredds/files/852780/tst_enums.zip)

lesserwhirls:
Hi @DennisHeimbigner - I updated the file on our aws jenkins test data. Sorry for the delay!

cwardgar:
I see only 2 test failures:
```
dap4.test.TestServletConstraints > testServletConstraints FAILED
    java.lang.AssertionError: ***Fail
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.assertTrue(Assert.java:41)
        at dap4.test.TestServletConstraints.dodmr(TestServletConstraints.java:283)
        at dap4.test.TestServletConstraints.doOneTest(TestServletConstraints.java:241)
        at dap4.test.TestServletConstraints.testServletConstraints(TestServletConstraints.java:222)
```
and
```
dap4.test.TestServlet > testServlet FAILED
    java.lang.AssertionError: ***Fail
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.assertTrue(Assert.java:41)
        at dap4.test.TestServlet.dodmr(TestServlet.java:246)
        at dap4.test.TestServlet.doOneTest(TestServlet.java:207)
        at dap4.test.TestServlet.testServlet(TestServlet.java:188)
```
There's a bunch of exception stack traces related to #537, but those aren't actual test failures.

lesserwhirls:
I just ran this branch on [Jenkins](https://jenkins-aws.unidata.ucar.edu/job/dap4.part4.dmh/lastCompletedBuild/testReport/), and in addition to the two failing tests that @cwardgar points out, we still have this one dap4 related test failing (although this one has been failing on 5.0 for awhile):

`thredds.server.dap4.TestDap4.testSimpleDap4GridDataset`

~~~java
java.lang.AssertionError
	at thredds.server.dap4.TestDap4.testSimpleDap4GridDataset(TestDap4.java:37)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
~~~

### Standard Output

~~~bash
open= http://localhost:8081/thredds/catalog.xml
fatalError= Cant open http://localhost:8081/thredds/dap4/cdmUnitTest/ft/coverage/03061219_ruc.nc 
err=DapNetcdfFile constructor failed: Cannot open http://localhost:8081/thredds/dap4/cdmUnitTest/ft/coverage/03061219_ruc.nc
~~~

### Standard Error

~~~bash
Mar 19, 2017 4:03:47 PM ucar.nc2.dataset.NetcdfDataset openDap4ByReflection
SEVERE: DapNetcdfFile constructor failed: Cannot open http://localhost:8081/thredds/dap4/cdmUnitTest/ft/coverage/03061219_ruc.nc
~~~

DennisHeimbigner:
As far as I can tell, the situation is as follows.

> thredds.server.dap4.TestDap4.testSimpleDap4GridDataset
Failing on file not found:
    http://localhost:8081/thredds/dap4/cdmUnitTest/ft/coverage/03061219_ruc.nc 

>dap4.test.TestServlet.testServlet
>dap4.test.TestServlet.testServletConstraints
These appear to be failing because an older version of the netcdf-c library
is being used (4.3.x).

All the other failing tests under jenkins appear to be the same
ones that fail under 5.0.0.

cwardgar:
We're running `netcdf-4.4.1` on the Jenkins slaves. Is that too old?

DennisHeimbigner:
Odd. I have a test to read the version thru the netcdf-c API,
and it is returning 4.3.x
In any case 4.4.1 should be recent enough.
=Dennis

dopplershift:
I wouldn't ignore that version discrepancy...

lesserwhirls:
Yeah, I'm concerned about that difference in version info. I'll take a look.

DennisHeimbigner:
It is possible, though unlikely, that the nc_inq_libvers function in the netcdf-c library
is returning the wrong version #.

lesserwhirls:
Nope, @DennisHeimbigner, your function is working properly. Our test is looking for the netCDF library on our slave machines under the directory `/opt/netcdf-4.4.1/lib`, but that directory does not exist. What does exist is an install of netCDF v4.3.3.1 in `/usr/local/lib`, and that is what is being picked up and used. I'm building a new install for netCDF-C using the latest hdf5 and netcdf-c, and will create a new shapshot for our jenkins slave machines to use. I will place the install in `/usr/local/lib`, and then we won't need to specify anything special in our Jenkins config file. I'll let you know when this is updated and working. Good catch!

lesserwhirls:
Actually, that netCDF install is in `/usr/`...not for long, though...

cwardgar:
Sean, are you sure? In the log, I see:
```
INFO: Nc4Iosp: NetCDF-4 C library loaded (jna_path='/opt/netcdf-4.4.1/lib', libname='netcdf').
```

lesserwhirls:
Yep, I'm pretty sure. I'm on a machine created by the slave AMI, and this is what I see:

~~~bash
[ec2-user@ip-172-31-13-191 ~]$ ls -l /opt
total 4
drwxr-xr-x 5 root root 4096 Oct 29  2015 aws
[ec2-user@ip-172-31-13-191 ~]$ which nc-config
/usr/bin/nc-config
[ec2-user@ip-172-31-13-191 ~]$ nc-config --version
netCDF 4.3.3.1
[ec2-user@ip-172-31-13-191 ~]$ 
~~~

I wonder if there is a bug in that if the `-Djna.library.path` property is set and any library is found (even if the path the property points to does not exist), then the IOSP reports a successful loading based on that property?

cwardgar:
That's very weird. Ryan and I sat down and created that AMI together...and we installed 4.4.1 in /opt/.

@dopplershift Do you have any idea what's going on?

lesserwhirls:
While I'm in slave machine, I see that it's running the amazon java:

~~~
[ec2-user@ip-172-31-13-191 ~]$ which java
/usr/bin/java
[ec2-user@ip-172-31-13-191 ~]$ java -version
java version "1.7.0_131"
OpenJDK Runtime Environment (amzn-2.6.9.0.71.amzn1-x86_64 u131-b00)
OpenJDK 64-Bit Server VM (build 24.131-b00, mixed mode)
~~~

Should I go ahead and set things up to use Oracle?

lesserwhirls:
Ok, I ended up creating a new ami for our build slaves. Now we have this:

~~~bash
[ec2-user@ip-172-31-0-225 ~]$ java -version
java version "1.8.0_121"
Java(TM) SE Runtime Environment (build 1.8.0_121-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.121-b13, mixed mode)
[ec2-user@ip-172-31-0-225 ~]$ h5dump --version
h5dump: Version 1.10.0-patch1
[ec2-user@ip-172-31-0-225 ~]$ nc-config --all

This netCDF 4.4.1.1 has been built with the following features: 

  --cc        -> gcc
  --cflags    -> -I/usr/local/include -I/usr/local/include
  --libs      -> -L/usr/local/lib64 -lnetcdf

  --has-c++   -> no
  --cxx       -> 

  --has-c++4  -> no
  --cxx4      -> 

  --has-fortran-> no
  --has-dap   -> no
  --has-nc2   -> yes
  --has-nc4   -> yes
  --has-hdf5  -> yes
  --has-hdf4  -> no
  --has-logging-> no
  --has-pnetcdf-> no
  --has-szlib -> 

  --prefix    -> /usr/local
  --includedir-> /usr/local/include
  --libdir    -> /usr/local/lib64
  --version   -> netCDF 4.4.1.1
[ec2-user@ip-172-31-0-225 ~]$ 
~~~

I'm running a test to make sure everything is working with the new AMI...I have my doubts that our test data will mount though.

lesserwhirls:
As I expected. I made sure to add the correct entry to `fstab` in the slave ami for the test data snapshot, but it didn't mount. @dopplershift - was there some magic somewhere for this?

dopplershift:
The only thing I can think of is creating the mount point, which I don't see on the instance I just started...

lesserwhirls:
I half-way created the mount point...re-imaging now...

lesserwhirls:
Ok, it mounted this time...let's see how this run goes. At any rate, I'm ok merging this, although I am interested to see if @DennisHeimbigner's version report shows 4.4.1.1.

lesserwhirls:
The test data are mounted, but now I'm getting a bunch of `java.io.IOException: No space left on device` errors...I'll get this all fixed up today (sorry)

DennisHeimbigner:
Let me know when Jenkins is ready to use again.
=Dennis

lesserwhirls:
Ok, Jenkins is up and running again, and I've updated all the configs to point to the new netCDF-C location. Both `v5.0` and this PR report 19 failures.

It looks like there are just two failing tests on Travis (https://github.com/Unidata/thredds/pull/782#issuecomment-287601456), but it would be good to get those knocked out so we can cleanly merge.

DennisHeimbigner:
I am working on the TestServlet and TestServletConstraints failures.

DennisHeimbigner:
Ok, it passes Travis, so I think we can consider merging it.

lesserwhirls:
Thanks @DennisHeimbigner!

PR #790, opened 2017-03-31
Nc4Iosp fixes

### LINKED ISSUE #761: Closing NetcdfFileWriter after redefine
When closing a NetcdfFileWriter that has been set to redefine mode and back, some internal filestream is kept open. Example:

@Test
public void testCloseFileWriter() throws IOException {
    final File theFile = new File("somewhere.nc");

    final NetcdfFileWriter netcdfFileWriter = NetcdfFileWriter.openExisting(theFile.getAbsolutePath());

    netcdfFileWriter.setRedefineMode(true);
    netcdfFileWriter.setRedefineMode(false);

    netcdfFileWriter.close();

    if (!theFile.delete()) {
        fail();
    }
}

This test fails on Windows OS, on Linux the FS just deletes, no matter if there are streams open.

Has been verified with netcdf4 version 4.6.7 and 4.6.8.

cwardgar:
This PR contains 3 separate bug fixes for `Nc4Iosp`. See individual commits for details.
* Fixes [DBP-690959](https://andy.unidata.ucar.edu/esupport/staff/index.php?_m=tickets&_a=viewticket&ticketid=28098&departmentid=15&ticketstatusid=3).
* Fixes #761 

The most surprising was that it wasn't possible to write an unlimited dimension to a NetCDF-4 file. How the hell didn't we know that before?


lesserwhirls:
Hmmm...I literally just wrote a netCDF file with an unlimited dimension using the `addUnlimitedDimension` method of `NetcdfFileWriter`. What way was failing previously?

cwardgar:
And it was NetCDF-4?

lesserwhirls:
Yes.

lesserwhirls:
Quick and dirty, but...

~~~java
import java.io.IOException;

import ucar.ma2.Array;
import ucar.ma2.DataType;
import ucar.ma2.InvalidRangeException;
import ucar.nc2.Attribute;
import ucar.nc2.Group;
import ucar.nc2.NetcdfFileWriter;
import ucar.nc2.Variable;

public class nc4tst {

    public static void main(String[] args) {
        String fname = "/Users/sarms/Desktop/hi.nc";
        try (NetcdfFileWriter ncFile = NetcdfFileWriter.createNew(NetcdfFileWriter.Version.netcdf4, fname)) {
            // create a null group (although could just use null for group in method calls)
            Group group = ncFile.addGroup(null, "");

            // Create a dimension
            ncFile.addUnlimitedDimension("x");
            // Create a variable
            Variable arr = ncFile.addVariable(group, "arr", DataType.FLOAT, "x");
            // Add a variable attribute
            arr.addAttribute(new Attribute("attr", "val"));

            // add a global attribute
            group.addAttribute(new Attribute("globalAttr", "val"));

            // Create file and exit redefine
            ncFile.create();

            // Create an array of data
            float[] data = new float[]{1.f, 2.f, 3.f, 4.f, 5.f, 6.f, 7.f, 8.f, 9.f, 10.f};
            int[] shape = new int[]{10};
            Array arrData = Array.factory(DataType.FLOAT, shape, data);

            // Write data to file
            ncFile.write(arr, arrData);
        } catch (IOException e) {
            e.printStackTrace();
        } catch (InvalidRangeException e) {
            e.printStackTrace();
        }
    }
}
~~~

~~~bash
netcdf hi {
dimensions:
	x = UNLIMITED ; // (10 currently)
variables:
	float arr(x) ;
		arr:attr = "val" ;

// global attributes:
		:globalAttr = "val" ;
data:

 arr = 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 ;
}
~~~

netcdf4/src/main/java/ucar/nc2/jni/netcdf/Nc4Iosp.java
```diff
@@ -2339,8 +2350,10 @@ private void createGroup(Group4 g4) throws IOException {
 
     // dimensions
     for (Dimension dim : g4.g.getDimensions()) {
-      int dimid = addDimension(g4.grpid, dim.getShortName(), dim.getLength());
+      int dimLength = dim.isUnlimited() ? Nc4prototypes.NC_UNLIMITED : dim.getLength();
+      int dimid = addDimension(g4.grpid, dim.getShortName(), dimLength);
```
cwardgar: This is the relevant bit. In order to create an unlimited dimension, we must pass a `dimLength` of `NC_UNLIMITED` to `addDimension()`. The value of `NC_UNLIMITED` is `0`. So, your specific example worked ONLY because the length of your unlimited dimension was `0` at the time. For any other value, it would have failed. Maybe you can give it a shot real quick and double-check that I'm right?

lesserwhirls: I'm confused - I didn't add the dimension using `addDimension`, I added it using `addUnlimitedDimension`.  Does `addUnlimitedDimension` set the length to zero by default? If I try using:

~~~java
boolean isShared = true;
boolean isUnlimited = true;
boolean isVlen = false;
ncFile.addDimension(group, "x", 10, isShared, isUnlimited, isVlen);
~~~

then I get an error, which is not what I expected. Is that the bug? Sorry, Friday brain here...

cwardgar: Yes, `addUnlimitedDimension` creates a zero-length dimension.
```java
  public Dimension addUnlimitedDimension(String dimName) {
    return addDimension(null, dimName, 0, true, true, false);
  }
```

cwardgar: And yeah, you'll get an error which is COMPLETELY unhelpful in determining the cause. That was fun to unwind.

lesserwhirls: Ok, I see now. Yes, the `NetCDF: Bad chunk sizes.` error message is completely unhelpful here. Yuck.

cwardgar: Yep, and I originally created a fix that targeted the chunk sizes generated by `Nc4ChunkingDefault`. As it turns out, the chunk sizes were fine, but ONLY for an unlimited variable. And of course, we weren't marking the variable's dimension as unlimited.

cwardgar:
*pokes Travis*

dopplershift:
Travis seems to be having problems...

lesserwhirls:
@cwardgar - can you have travis rerun this PR? I think it's working now.

cwardgar:
Looks like reopening the PR finally kicked off a Travis build.

Issue #800, opened 2017-04-12
Enable reading remote NcML files


lesserwhirls:
Currently the CDM, specifically NetcdfDataset, does not allow users to read NcML files that are hosted on remote servers and served via HTTP. For example, a user cannot pass in the following URL:

http://weather.rsmas.miami.edu/repository/entry/get/entryid:dab6c075-cf2a-45d5-bfa3-966e8a7ae38f/merra_aimcpasm.ncml

even though it only references other remote datasets (i.e. no local datasets need to be downloaded before the NcML file can be used).

If, in `NetcdfDataset` (specifically the `disambiguateHttp` method), the code is able to identify the file hosted at the URL as an NcML file, then netCDF-Java will read it without any issues. With a little modification of the code to detect that the hosted file is indeed an NcML file, things "just work" (™️)

PR on the way.

lesserwhirls:
@yuanho - I hope to get it in our next 4.6 release (likely next week).

cofinoa:
@lesserwhirls this is related also to issue #797 

Please take a look my last comment https://github.com/Unidata/thredds/issues/797#issuecomment-293881425

lesserwhirls:
Hi @cofinoa, the issues are somewhat related. However, in this case, it's possible that the `NcML` file does not reference any local file at all (i.e. the `location` attribute of the `netcdf` element references an `OPeNDAP` URL). Given that `NcML` files are generally small and have this ability, it makes sense to at least give it a try. With something like a `GRIB` file, you need to index the file first, which requires scanning through the entire file. If you already have to read through the entire file once, you are probably better off downloading it locally one time (assuming you are not accessing the file through a webservice like `OPeNDAP` on the TDS).

cofinoa:
You can enforce to read as ncml by adding the fragment `protocol=ncmdl` to the URL like this:
http://weather.rsmas.miami.edu/repository/entry/get/entryid:dab6c075-cf2a-45d5-bfa3-966e8a7ae38f/merra_aimcpasm.ncml#protocol=ncmdl

PR #801, opened 2017-04-12
Allow NetcdfDataset to open remote NcML Files

lesserwhirls:
fixes Unidata/thredds#800

### DIFF
--- a/cdm/src/main/java/ucar/nc2/dataset/NetcdfDataset.java
+++ b/cdm/src/main/java/ucar/nc2/dataset/NetcdfDataset.java
@@ -704,6 +704,13 @@ public class NetcdfDataset extends ucar.nc2.NetcdfFile {
         svctype = decodePathExtension(trueurl); // look at the path extension
       } else {
         svctype = disambiguateHttp(trueurl);
+        // special cases
+        if ((svctype == null || svctype == ServiceType.HTTPServer)) {
+          // ncml file being served over http?
+          if (checkIfRemoteNcml(trueurl)) {
+            svctype = ServiceType.NCML;
+          }
+        }
       }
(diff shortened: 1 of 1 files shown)


cofinoa:
@lesserwhirls This PR diverges from how the NCML service type detection is made for the file protocol: 

https://github.com/Unidata/thredds/blob/v4.6.8/cdm/src/main/java/ucar/nc2/dataset/NetcdfDataset.java#L697-L708

may be it woul better to do something like this:

```java
    if (svctype == null) {
      //There are several possibilities at this point; all of which
      // require further info to disambiguate
      //  - we have file://<path> or file:<path>; we need to see if
      //    the extension can help, otherwise, start defaulting.
      //  - we have a simple url: e.g. http://... ; contact the server
      if (leadprotocol.equals("file")) {
        svctype = decodePathExtension(trueurl); // look at the path extension
      } else {
          svctype = disambiguateHttp(trueurl);
          if ((svctype == null || svctype == ServiceType.HTTPServer) && decodePathExtension(trueurl)==ServiceType.NCML)
              svctype = ServiceType.NCML;
      }
    }
```

in case the `svctype` is null or is HTTPServer, try with `decodePathExtension`

The other option it's would be to implement your appoach by reading the head and discover the ncml namespace ... or make both just in case.

lesserwhirls:
Thanks for the feedback @cofinoa!

The idea behind reading the first bit of the file is like the IOSP "fail fast" approach, especially since we would possibly be trying to download something. However, maybe it's not a big deal given we are looking at an NcML file (which are typically small). Things started to get a bit crowded, so I offloaded the code to `checkIfRemoteNcml`.

PR #803, opened 2017-04-13
Generate Web Start JNLP files programmatically

cwardgar:
* Created 2 custom Gradle tasks, plus unit and integration tests for those tasks.
* Removed `:tds:filterWebStartFile`, as its functionality has been subsumed by `ToolsUiJnlpBaseTask`.
* Updated release procedure doc.
* Updated Gradle from 2.14 to 3.4.1, which was necessary to resolve a bug I was seeing in the integration tests.


buildSrc/build.gradle
```diff
@@ -27,6 +31,73 @@ dependencies {
...
+ *
+ * The external-only route is easiest: we already have "xerces:xercesImpl:2.9.1" on the classpath due to
+ * gradleApi(), so we merely need to add an external Xalan. Something like this works:
+ *     testRuntime "xalan:xalan:2.7.2"
+ *
+ * However, we're gonna go with the built-in-only option instead, so that we can avoid adding another
+ * dependency to the project. How do we do that, given that an external Xerces is stuck on the classpath?
+ * It turns out that there are system properties that we can set to specify which implementations of the key
+ * JAXP classes to use (see https://goo.gl/WgJ0EM). With those, we can force only the built-in libraries to be used.
+ */
```
cwardgar: OMFG DEPENDENCY HELL. Sometimes I hate Java.

lesserwhirls: That's...fun?

cwardgar:
Reopened PR to see if I can kick-start Travis.

buildSrc/src/test/resources/edu/ucar/build/ui/netCDFtoolsExtraJars.jnlp
```diff
@@ -0,0 +1,13 @@
...
+<jnlp spec="7.0" codebase="http://www.unidata.ucar.edu/software/thredds/current/netcdf-java/webstart"
+        href="netCDFtoolsExtraJars.jnlp" version="1.5">
+    <component-desc/>
+    
+    <resources>
+        <jar href='root-1.5.jar' main='true' download='eager' />
+    
+        <jar href='slf4j-api-1.7.7.jar' download='lazy' />
+        <jar href='objenesis-2.2.jar' download='lazy' />
+        <jar href='hamcrest-core-1.3.jar' download='lazy' />
```
lesserwhirls: Is there any way to tie these dependency versions into what is used in `dependencies.gradle`? The `objenesis` dep in `dependencies.gradle` is version 2.4, and `slf4j` is 1.7.22, whereas the ones here are older. I can just imagine the dependencies in netCDFtoolsExtraJars.jnlp will get accidentally forgotten when we update other libraries in `dependencies.gradle`

buildSrc/src/test/groovy/edu/ucar/build/ui/ToolsUiJnlpExtensionTaskSpec.groovy
```diff
@@ -0,0 +1,133 @@
...
+        GradleRunner gradleRunner = GradleRunner.create().withProjectDir(tempFolder.root).withArguments(":$taskName")
+        
+        expect: "if we execute the task, it will succeed"
+        gradleRunner.build().task(":$taskName")?.outcome == TaskOutcome.SUCCESS
+        
+        and: "if we re-execute the task without changing anything, it'll be UP-TO-DATE"
+        gradleRunner.build().task(":$taskName")?.outcome == TaskOutcome.UP_TO_DATE
+        
+        when: "we update a dependency, which will change the runtime configuration Input that the task uses by default"
+        buildFile.text = buildFileContent.replace("compile 'org.slf4j:slf4j-api:1.7.7'",
```
lesserwhirls: see https://github.com/Unidata/thredds/pull/803/files#r111667152

buildSrc/src/test/resources/edu/ucar/build/ui/netCDFtoolsExtraJars.jnlp
cwardgar: This file is only a test resource (a "control" file used for comparison -- look at its path). I know that's confusing, because there used to be a file with the same name (but slightly different path) that we edited and then published every release. In any event, the versions of these dependencies are what they are because of an integration test.

buildSrc/src/test/groovy/edu/ucar/build/ui/ToolsUiJnlpExtensionTaskSpec.groovy
```diff
@@ -0,0 +1,133 @@
...
+            targetCompatibility = JavaVersion.VERSION_1_7
+            
+            repositories {
+                jcenter()
+            }
+            dependencies {
+                compile 'org.slf4j:slf4j-api:1.7.7'
+                compile 'org.objenesis:objenesis:2.2'
+                runtime 'org.hamcrest:hamcrest-core:1.3'
+                testRuntime 'org.codehaus.groovy:groovy-all:2.4.5'
```
cwardgar: Specifically, this integration test. Notice how the versions match. Notice also that `'org.codehaus.groovy:groovy-all:2.4.5'` doesn't appear in the control file. and won't be written by `ToolsUiJnlpExtensionTask.Writer`. That's because, below, we have:
```
dependenciesConfig = rootProject.configurations.runtime
```
meaning "only include jars that appear in the runtime configuration". `groovy-all` doesn't appear there (it's in `testRuntime`), but the other 3 do (`runtime` extends the `compile` configuration). I was just testing that ONLY jars in the specified configuration appear in the output JNLP file.

buildSrc/src/main/groovy/edu/ucar/build/ui/ToolsUiJnlpExtensionTask.groovy
```diff
@@ -0,0 +1,94 @@
...
+            
+                xml.jnlp(spec: '7.0', codebase: codebase, href: outputFile.name, version: applicationVersion) {
+                    'component-desc'()
+                
+                    resources() {
+                        jar(href: applicationJarName, main: 'true', download: 'eager')
+                    
+                        dependenciesConfig.resolve().each {
+                            jar(href: it.name, download: 'lazy')
+                        }
```
cwardgar: Here's the relevant code that selects JARs for the JNLP output file. As you can see, it's not fixed; it'll include whatever is in `dependenciesConfig`. (Okay, that was my last comment on the topic. Hopefully you were able to read the previous ones chronologically in the order that I posted them.)

lesserwhirls:
Ok, this makes sense. Thanks for your comments @cwardgar!

PR #804, opened 2017-04-15
fixed ignored error from getRequestState()

DennisHeimbigner:
re: e-support CZN-403791
AbstractServlet.getRequestState is throwing
a null pointer exception and it is being ignored.
Modify to properly report failures in getRequestState.
Does not fix cause of the failure, but at least it gets reported now.


opendap/src/main/java/opendap/servlet/AbstractServlet.java
```diff
@@ -413,19 +413,22 @@ public void IOExceptionHandler(IOException e,  ReqState rs) {
    * @param e        The exception that caused the problem.
    * @param rs The <code>ReqState</code> for the client.
    */
-  public void anyExceptionHandler(Throwable e, ReqState rs) {
-    HttpServletResponse response = rs.getResponse();
+  public void anyExceptionHandler(Throwable e, ReqState rs)
+  {
+    log.error("DODServlet ERROR (anyExceptionHandler): " + e);
+    e.printStackTrace();
+    printThrowable(e);
```
cwardgar: I dislike adding new code that prints to STDOUT or STDERR. Why not just use [this method](https://www.slf4j.org/api/org/slf4j/Logger.html#error(java.lang.String,%20java.lang.Throwable)), which will print the entire exception message and stack trace to the log? Better still would be if you removed *all* printing to STDERR from the whole class.

cwardgar: Now that I think about it, I'm not completely sure that that method will print the exception's stack trace to the log. Maybe 95% sure. Wouldn't hurt to test.

DennisHeimbigner:
Agree about changing to use log; but this code is same as before, just moved location.

cwardgar:
Just moved location? I don't understand. Is there some harm in making those changes? Seems pretty straightforward.

cwardgar:
Oh, you're saying that the code isn't exactly "new". Okay, but those changes would still be good.

cwardgar:
Can you rebase this?

DennisHeimbigner:
ok, up to date

cwardgar:
We've got conflicts. Possibly because you merged in master instead of rebasing on top of it? I dunno.

DennisHeimbigner:
I did rebase it. Why does the pull request say there are no conflicts?

cwardgar:
Not sure, but I see a merge commit (https://github.com/Unidata/thredds/pull/804/commits/7eaaaaac3f4cab211e42ff51642340edbfdd7dc2) and a PR in conflict: 
screenshot from 2017-04-15 19-08-00
Do you see something diffrent?

DennisHeimbigner:
Yes.ee atached
[pr804.pdf](https://github.com/Unidata/thredds/files/924021/pr804.pdf)

cwardgar:
Well, that's weird as hell. I wonder what @lesserwhirls sees.

Still, even on your end, I see a merge commit, as well as the inclusion of commits from me and Sean in the PR. That tells me that something went wrong with your rebase. Try the rebase again, maybe? Possibly in a new PR.

DennisHeimbigner:
ok I will see what I can do.
=Dennis

DennisHeimbigner:
Ok, I started over and created pr 806 from most recent version
of master + changes from 804.
Hope that fixes things.
=Dennis

cwardgar:
Superseded by #806.

PR #805, opened 2017-04-15
Change XSLT version from 1.0 to 2.0

cofinoa:
The XSLT document is using features from version 2.0 but the header mentions version 1.0. 

For example for the some `xsl:variable` elements the `as` attribute is been used. 

[version 1.0] https://www.w3.org/TR/xslt#variables
[version 2.0] https://www.w3.org/TR/xslt20/#variables


cwardgar:
@lesserwhirls Should this change be made to [threddsIso](https://github.com/Unidata/threddsIso/blob/master/src/main/resources/xsl/nciso/UnidataDD2MI.xsl) instead? Why do we have duplicate files in the first place? I'm not really sure how the iso stuff works.

lesserwhirls:
Things are a bit interesting for threddsIso. The code lives in the threddsIso repository, and the canonical transforms and other resource files live there as well. From what I understand, we have to manually pull those over to the THREDDS code base to make sure they are accessible in the `thredds.war` file we create.

Currently there is a PR (threddsIso#31) to update `UnidataDD2MI.xsl` to use ACDD 1.3, although that PR still uses version 1.0 in the header. I'll mention this PR in that PR to see if @noaaroland can include the change.

lesserwhirls:
Actually, the xls files in @noaaroland's PR do not use `as`, so maybe not an issue anymore? Once his PR is accepted, then we can copy over those new transforms to use in the TDS.

PR #756, opened 2017-02-15
Add ':visadCdm' to netcdfAll and ToolsUI, for Gempak IOSPs

cwardgar:
Fixes [ADT-625226](https://andy.unidata.ucar.edu/esupport/staff/index.php?_m=tickets&_a=viewticket&ticketid=27956).

PR #808, opened 2017-04-18
Release 4.6.9

cwardgar:


msdsoftware:
The size of netcdfAll-4.6.x.jar jumped from 26 MB to 32 between 4.6.8 and 4.6.9. Any info available on why?

cwardgar:
Dependency updates and the [addition of visadCdm](https://github.com/Unidata/thredds/pull/756). [Here](https://www.diffchecker.com/87AQ2Umj) is the dependency diff.

If we're trying to reduce the size of the fat jar, none of those deps jump out at me as easy cuts. Looks like it might be time to investigate [ProGuard](https://www.guardsquare.com/en/proguard/manual/introduction). NetCDF-Java loads a ton of classes by reflection, so it probably won't be straightforward to shrink it. I anticipate lots of trial-and-error.

cwardgar:
I take that back: `com.amazonaws:aws-java-sdk-s3:1.11.113` is a relatively easy cut, and would save a lot of space. I'll think about that.

msdsoftware:
Please do think about it. I was one of those who whined a couple years ago about the bloat in netcdfAll, so was overjoyed in the size was halved going from 4.6.0 to 4.6.1. Hate to see that bloat creep back in.

lesserwhirls:
Our 5.0 `netCDFAll.jar` artifact is at 26 currently. I wonder what we are missing in the 5.0 version, and if we can keep missing it for the sake of size.

cwardgar:
I've created an issue for this: #818. I doubt we'll get to it before the 5.0.0 release.

msdsoftware:
Please remind me... what's the timetable for 5.0?

I'm pondering how to manage the collision of the netCDF-Java distro going to 5.0, Java going to 9.0, and my own app going up a major version number. A long vacation might be required.

lesserwhirls:
Ah, 5.0...it's almost mythical at this point. We are actually pretty close to releasing a beta of the TDS (looking at this month). Once we have the TDS released as stable, we will get a beta of netCDF-Java out the door. My guess is netCDF-Java 5.0 will not be stable until early next year at the earliest. I would bet that Java 9 will be out first, which I think is scheduled for this summer I think.

msdsoftware:
Java 9 is scheduled for July 27. I have concerns that early adopters on macOS are going to find my code breaks. I need to do some testing but don't want to mess up my current dev tools.

PR #811, opened 2017-04-19
Tweak generation of ToolsUI JNLP files so they work on our web server

cwardgar:
ui.jar's 'Codebase' manifest attribute and the `/jnlp/@codebase` attributes of ToolsUI JNLP files are now kept in sync. They are controlled by a single variable in `:ui`: `webstartCodebase`. It turns out that different values for all of these codebases was the principal cause of the "Application Blocked by Java Security" failures we were seeing in javaws.
* The ToolsUiJnlp tasks now have a `codebase` property. Updated unit tests to exercise the new property.
* The aforementioned property may be `null`, which suppresses output of the `/jnlp/@codebase` attribute. This is useful for local Web Start testing.
* Restored my (now simpler) notes about how to test Web Start locally.

Other changes:
* Don't include `/jnlp/@href` attribute in `ToolsUiJnlp*Task` output files.
* `ToolsUiJnlpExtensionTask` outputs must include `<security><all-permissions/></security>` element.
* Refactored `ToolsUiJnlp*Task` tests to be simpler and more consistent.
* For all URLs in our JAR manifests and JNLP files that point to our web sever, use HTTPS instead of HTTP.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,7 +26,7 @@ ext {
     vendor = "UCAR/Unidata"
 
     // Matches Maven's "project.url". Used in MANIFEST.MF for "Implementation-URL".
-    url = "http://www.unidata.ucar.edu/software/thredds/current/netcdf-java/documentation.htm"
+    url = "https://www.unidata.ucar.edu/software/thredds/current/netcdf-java/documentation.htm"
 
     SimpleDateFormat iso_8601_format = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ")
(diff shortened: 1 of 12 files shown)


cwardgar:
I prepared the 4.6.10 Web Start release using this code and verified that it works. [You can too](http://www.unidata.ucar.edu/software/thredds/current/netcdf-java/webstart/netCDFtools.jnlp). So, I'm gonna go ahead and merge this in myself, as it's blocking me from doing the rest of the 4.6.10 release. Plus, it's all Gradle and JNLP stuff anyway, which is hard for others to review.

PR #813, opened 2017-04-20
Cherry-pick ToolsUI Web Start work from master onto 5.0.0

cwardgar:
Applies changes from #803 and #811.

I knew that these edits would cause lots of conflicts down the road when we tried to merge 
master into 5.0.0, so I decided I'd just move them over now, while the code is still fresh in my head.


cwardgar:
Oh, I tried to do a full merge at first, but there were some other conflicts that I didn't feel comfortable trying to resolve by myself.

PR #814, opened 2017-04-23
fix data plot to timezone UTC

petejan:
fix so that data in dataplot is fixed in UTC timezone, not local


lesserwhirls:
Would it be possible to try to use the time zone defined in the `unit` attribute of the time coordinate associated with the variable being plotted, if there is one, before defaulting to UTC?

lesserwhirls:
I rebased this 5.0.0 to get rid of the merge commits introduced by me using the update branch feature of github. New PR can be found at Unidata/thredds#837. Thanks @petejan!

PR #815, opened 2017-04-24
Include additional information on web pages for catalogs if markdown files are present

cskarby:
Markdown is a popular lightweight markup language with plain text formatting syntax. Markdown is often used to format readme files, for writing messages in online discussion forums, and to create rich text using a plain text editor. As an example is the [README.md](https://github.com/Unidata/thredds/blob/5.0.0/README.md) in the thredds source code repository written in Markdown syntax.

TDS Catalog pages: Render additional information from markdown if present

1. For all catalog pages render ${tds.content.path}/README.md if present. This can be used to include a general information for all catalogs on one server.
2. For static catalogs e.g. ${tds.content.path}/subpath/example.xml render
${tds.content.path}/subpath/example.md
3. For dynamically generated catalogs render README.md in the directory that
was scanned for files, if that directory has no README.md, try in the parent
directory until reaching the parent catalog for the collection.

Markdown can be tested with a [markdown editor](https://stackedit.io/), with github or similar.


tds/src/main/java/thredds/servlet/HtmlWriter.java
```diff
@@ -542,6 +552,84 @@ public int writeCatalog(HttpServletRequest req, HttpServletResponse res, InvCata
...
+      return false;
+    }
+    Parser parser = Parser.builder().build();
+    Node node;
+    try {
+      node = parser.parseReader(reader);
+    } catch (IOException e) {
+      return false;
+    }
+    HtmlRenderer.builder().escapeHtml(true).build().render(node, output);
```
cskarby: Please note that HTML-tags are escaped, in order to keep the site maintainer in control, preventing html code injection.

lesserwhirls:
Very nice @cskarby! Let me download these changes and play around a bit so that I can provide some feedback.

Note that in the long term we would like to move catalog html generation over to using a template system, like `thymeleaf`, but in the meantime, this would be an excellent addition!

lesserwhirls:
Greetings @cskarby!

This is very nice! I have just a few points of feedback:

1. I could not get this to work using the `<datasetScan>` element (i.e. with a `README.md` file in the directory structure being scanned). Were you able to get that to work?
2. Given that markdown can add horizontal rule elements, I would suggest not hard coding that element in the `HtmlWriter` class.
3. For styling purposes (which can be controlled using the `TDS.css` file in the `${tds.content.root.path}/public` directory), would it be possible to add a class to all of the HTML elements generated from the markdown files? Something like 

    `<... class="tdsMarkdownGenerated ...>`
4. Would you be willing to add a bit of documentation? Maybe at the end of the page [ConfigCatalogs.html](https://github.com/Unidata/thredds/blob/master/docs/website/tds/tutorial/ConfigCatalogs.html)? (currently online [here](http://www.unidata.ucar.edu/software/thredds/current/tds/tutorial/ConfigCatalogs.html)).

Again, thank you for the work you did getting this up and running. Very useful contribution!

PR #816, opened 2017-04-25
Add tests for a fresh installation of TDS

cwardgar:
An installation is considered "fresh" if the `tds.content.root.path` property:
* points to a non-existent directory. In this case, TDS will create that directory, if it can.
* points to an existing directory that does not contain a "catalog.xml" file.

The tests ensure that default startup files are copied over and we can resolve the top-level catalog.xml.

Also:
* Improved Jacoco reporting.
* Replaced use of 'provided' configuration with 'compileOnly'.

See individual commits for more detail.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,7 +25,6 @@ buildscript {
         classpath libraries["gretty"]
         classpath libraries["shadow"]
         classpath libraries["coveralls-gradle-plugin"]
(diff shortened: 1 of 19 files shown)


lesserwhirls:
Dang - 🥇 for pushing through all that to get the test!

cwardgar:
Rebased.

PR #817, opened 2017-04-25
Change WCS example links from motherlode to thredds

lesserwhirls:


cwardgar:
Some or all of the links don't work. Shouldn't we replace them with links that do?

lesserwhirls:
Which links? All of the WCS queries work. The example WCS urls 400 because there is no query on them. I think these are the links you would put into a WCS client, and from that the client would add queries to talk to the service.

I did see a few broken links in the WCS client list, so I've updated those based on what I've done int he 5.0 docs.

cwardgar:
Okay, the example links do work when the HTML is rendered. I was trying to copy/paste and the `&amp;`s were screwing me up. Oops.

PR #823, opened 2017-04-28
Added -D controls for HTTPSession timeouts

DennisHeimbigner:
Added some JVM -D flags to control the behavior of HTTPSession.
Specifically, added the following two flags:
1. -Dtds.http.conntimeout - set the timout (in seconds) for initial connection to a server
2. -Dtds.http.sotimeout - set the timeout (in seconds) for read requests on a socket


cwardgar:
Looks good. Thanks for fixing the `buildSrc` tests.

Issue #48, opened 2014-04-16
Integrate ncSOS into TDS codebase


rsignell-usgs:
@JohnLCaron  and @cwardgar, I spoke with Dave Foster (@daf) today, and although he said he would have liked to fix a few more bugs first, he was fine with you guys incorporating ncSOS into this Unidata/thredds repo, and work out the bugs here.   The question we had was how this should work.  

If you clone ncSOS and add to the thredds repo, do we then post issues  on github or do we need to use the Unidata Jira?  

Or do you have other ideas about how to manage the shared work?

@daf, please jump in if I've munged this topic up.

Thanks,
Rich

lesserwhirls:
Moving to https://bugtracking.unidata.ucar.edu/browse/TDS-428

lesserwhirls:
Our Jenkins site is no longer used and has been removed, as we've moved to github issues, so opening this back up.

See issue asascience-open/ncSOS#186 and PR asascience-open/ncSOS#187. This is the only thing at this point to be done. The idea is to incorporate the plugin into the standard TDS build, starting with TDS v4.6 (perhaps next release, depending on how fast the PR above moves).

PR #826, opened 2017-05-03
Add ncsos service

lesserwhirls:
Addresses Unidata/thredds#48

### DIFF
--- a/tds/build.gradle
+++ b/tds/build.gradle
@@ -39,6 +39,7 @@ dependencies {
     compile libraries["spring-webmvc"]
 
     runtime libraries["threddsIso"]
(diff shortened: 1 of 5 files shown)


tds/build.gradle
```diff
@@ -39,6 +39,7 @@ dependencies {
     compile libraries["spring-webmvc"]
 
     runtime libraries["threddsIso"]
+    compile libraries["ncsos"]
```
cwardgar: Can this be a 'runtime' dep?

lesserwhirls: Actually, yes! Good catch!

PR #829, opened 2017-05-05
update docs for wms

lesserwhirls:
Now that the reading trac site is gone, we should pull in the
relevant docs for detailed wms config, godiva2, and update links
to the reading site


cwardgar:
Can you make relative links instead of full URLs, where possible?

lesserwhirls:
Good call - thanks! Updated.

PR #843, opened 2017-05-19
Make sure checkFileType leaves correct RAF file position

DennisHeimbigner:
Turns out that at least H4header assumes that checking
the file type leaves the raf file pos just after the header.
So modify NCheader.checkFileType to guarantee that constraint.


cdm/src/main/java/ucar/nc2/iosp/NCheader.java
```diff
@@ -58,41 +74,47 @@
...
-            return 0; // unknown
+        int hdrlen = 0;
+        byte[] byte4 = Arrays.copyOfRange(magic,0,CDF1HEAD.length);
+        hdrlen = CDF1HEAD.length;
+        format = 0;
+        if(Arrays.equals(CDF1HEAD,byte4))
+            format = NC_FORMAT_CLASSIC;
+        else if(Arrays.equals(CDF2HEAD,byte4))
+            format = NC_FORMAT_64BIT_OFFSET;
+        else if(Arrays.equals(CDF1HEAD,byte4))
```
cwardgar: Should be `CDF5HEAD` here. Shouldn't our tests have caught this?

cdm/src/main/java/ucar/nc2/iosp/NCheader.java
```diff
@@ -58,41 +74,47 @@
...
+            format = NC_FORMAT_CDF5;
+        if(format != 0) {
+            raf.seek(hdrlen);
+            return format;
         }
         // For HDF5/4, we need to search forward
```
cwardgar: According to [the docs](https://support.hdfgroup.org/release4/doc/UsrGuide_html/UG_html-2.htm#906736832_pgfId-92358), the HDF4 magic bytes are always at the beginning of the file. This is in contrast to [HDF5](https://support.hdfgroup.org/HDF5/doc/H5.format.html#Superblock). It looks like we've had code in `H4header` to search forward for the magic since 2007, but it's been unnecessary all that time. So, you could simplify the code here and search for `H4HEAD` in the same way that you search for `CDF1HEAD`, `CDF2HEAD`, and `CDF5HEAD`.

DennisHeimbigner:
TestCDF5Reading does not actually invoke NCheader.chackFileType:
it assumed that its file was a cdf5 file.
So, added TestCheckFileType test and some test files.
Then fixed any errors it reported.

DennisHeimbigner:
Interesting, the old code in H4header did search forward like HDF5.
BTW, I could not see this in the document you referenced, but it is clear
in this one.
https://support.hdfgroup.org/release4/doc/DSpec_html/DS.pdf

cwardgar:
Looks good!

PR #848, opened 2017-05-30
Add detailed usage and license information for others' code in thredds.

cwardgar:
* So far, we're only covering code that we've included from Gretty and ERDDAP.
* Added READMEs to `:buildSrc` and `:waterml`, which give very specific details of our usage of that code.
* Added a "Licensing" section to the top-level README that links to the new subproject READMEs.
* Added licenses to `docs/licenses/third-party`.

Hopefully all of this can act as a template for future inclusions of third-party open-source code.


lesserwhirls:
👍

Issue #849, opened 2017-06-02
NCSS Response does not include subset file size in header  (TDS 5.0.0)


bonnland:
Hi, the ESG catalog is currently based on TDS 5.0.0-snapshot. 

We're hoping to capture successful downloads from our Tomcat logs for NetCDF subset requests.  It currently looks like the expected size of the downloaded file is not being included in the response header, so the Tomcat log does not include enough information to know whether the client received the entire file.   Example:

curl -v -o florida.nc https://tds.ucar.edu/thredds/ncss/grid/glade/p/vetssg/data/cordex/data/raw/NAM-22i/day/CanRCM4/CanESM2/hist/pr/pr.hist.CanESM2.CanRCM4.day.NAM-22i.raw.nc?var=pr&north=31.500&west=-88.000&east=-77.000&south=25.000&horizStride=1&time_start=1950-01-01T12%3A00%3A00Z&time_end=2005-12-31T12%3A00%3A00Z&timeStride=1&vertCoord=&addLatLon=true&accept=netcdf3

...

> GET /thredds/ncss/grid/glade/p/vetssg/data/cordex/data/raw/NAM-22i/day/CanRCM4/CanESM2/hist/pr/pr.hist.CanESM2.CanRCM4.day.NAM-22i.raw.nc?var=pr HTTP/1.1
> Host: tds.ucar.edu
> User-Agent: curl/7.49.0
> Accept: */*
>
{ [5 bytes data]
< HTTP/1.1 200 200
< Date: Wed, 31 May 2017 20:18:47 GMT
< Server: Apache
< Content-Disposition: attachment; filename=pr.hist.CanESM2.CanRCM4.day.NAM-22i.raw.nc
< Content-Type: application/x-netcdf
< Transfer-Encoding: chunked
<
{ [5 bytes data]
100 616k 0 616k 0 0 5522k 0 --:--:-- --:--:-- --:--:-- 5658k
* Connection #0 to host tds.ucar.edu left intact

The resulting file size was 631560 bytes. 

The tomcat log shows that 631560 bytes were sent, but it shows nothing for the expected size.   I believe this is because the content length is not included in the response header.  

If the content length could be included in the response header, we could calculate byte savings from the request, which would be valuable for our stakeholders.

PR #850, opened 2017-06-03
set content length for NCSS grid return

lesserwhirls:
Fixes one part of Unidata/thredds#849

It is not clear how to go about setting content-length on the Grid As Point service of NCSS, as the content headers appear to be set [before](https://github.com/Unidata/thredds/blob/5.0.0/tds/src/main/java/thredds/server/ncss/controller/NcssPointController.java#L123) the file is generated.

### DIFF
--- a/tdcommon/src/main/java/thredds/util/Constants.java
+++ b/tdcommon/src/main/java/thredds/util/Constants.java
@@ -1,37 +1,11 @@
 /*
- * Copyright 1998-2014 University Corporation for Atmospheric Research/Unidata
- *
- *   Portions of this software were developed by the Unidata Program at the
- *   University Corporation for Atmospheric Research.
- *
- *   Access and use of this software shall impose the following obligations
- *   and understandings on the user. The user is granted the right, without
- *   any fee or cost, to use, copy, modify, alter, enhance and distribute
- *   this software, and any derivative works thereof, and its supporting
- *   documentation for any purpose whatsoever, provided that this entire
- *   notice appears in all copies of the software, derivative works and
- *   supporting documentation.  Further, UCAR requests that the user credit
- *   UCAR/Unidata in any publications that result from the use of this
- *   software or in any product that includes this software. The names UCAR
- *   and/or Unidata, however, may not be used in any advertising or publicity
- *   to endorse or promote any products or commercial entity unless specific
- *   written permission is obtained from UCAR/Unidata. The user also
- *   understands that UCAR/Unidata is not obligated to provide the user with
- *   any support, consulting, training or assistance of any kind with regard
- *   to the use, operation and performance of this software nor to provide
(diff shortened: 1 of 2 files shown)


cwardgar:
Getting this to work for Grid-as-point and DSGs is a solid piece of work, especially since you'd need to edit each of the `DsgSubsetWriter` subclasses. And many of them use different writers for I/O.

So, just grid for now? Can you make an issue for the remaining work so we don't forget? Also, this needs a rebase.

lesserwhirls:
Yes, I am going to hold off on everything else but GRID requests. Once we have the NCSS Grid as Point html interface set, I can take a stab at the `DsgSubsetWriter` portion of this (new issue is Unidata/thredds#852)

ethanrd:
According to RFC 2616 (HTTP/1.1), Content-Length is not allowed if there is a Transfer-Encoding (e.g., chunk, deflate, gzip). @lesserwhirls and @cwardgar , do either of you know why, in the example in #849, the response would be "chunked". Is it automatically set if Content-Length is not set?

lesserwhirls:
Good question @ethanrd - I have no idea.

cwardgar:
It looks like it's automatically set is the response is over a given size. But, it is suppressed if you include`Content-Length`.
[Source](https://stackoverflow.com/a/7881249/3874643)

Can you try this out, Sean, and see what headers we're returning?

lesserwhirls:
Here is what the request looks like via Chrome:

```
Request URL:http://localhost:8080/thredds/ncss/grid/testAll/2004050412_eta_211.nc?var=Z_sfc&horizStride=1&time=2003-09-25T00%3A00%3A00Z&vertCoord=&accept=netcdf3
Request Method:GET
Status Code:200 
Remote Address:[::1]:8080
Referrer Policy:no-referrer-when-downgrade
```

and here is the response using the code in this PR:

```
Content-Disposition:attachment; filename=2004050412_eta_211.nc
Content-Length:54788
Content-Type:application/x-netcdf
Date:Wed, 07 Jun 2017 22:08:44 GMT
Server:Apache
```

and here is the response when making the request via curl (as done in Unidata/thredds#849):

```
* Connected to localhost (::1) port 8080 (#0)
> GET /thredds/ncss/grid/testAll/2004050412_eta_211.nc?var=Z_sfc HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.51.0
> Accept: */*
> 
< HTTP/1.1 200 
< Content-Disposition: attachment; filename=2004050412_eta_211.nc
< Content-Type: application/x-netcdf
< Content-Length: 54788
< Date: Wed, 07 Jun 2017 22:17:16 GMT
< Server: Apache
```

Looks like the right thing happens and `Transfer-Encoding` is dropped since the new code sets `Content-Length`.

cwardgar:
Good enough for me.

Issue #853, opened 2017-06-08
Inconsistent representation for Grib Level Type


ywangd:
For variables in a `NetcdfFile` object that is populated from a GRIB file, they have a Grib Level Type attribute, named `Grib1_Level_Type` and `Grib2_Level_Type` for GRIB-1 and GRIB-2 files, respectively.

The issue is that the data type of the attribute's value is different. 
* For `Grib1_Level_Type` attribute, the value is of an Integer, i.e. the level type code. The corresponding string representation is held by another attribute `Grib1_Level_Desc`.
* For `Grib2_Level_Type`, the value is of a String, i.e. the corresponding text for the level type code. The Integer value of the level type is simply NOT available.

For our usage purpose, we'd like both attributes to have Integer values. So maybe a possible solution is to store Integer code to `Grib2_Level_Type` and add a new `Grib2_Level_Desc` to store its string form.

lesserwhirls:
This seems more than reasonable. We'll look at getting this straightened out for our next release.

PR #855, opened 2017-06-08
Fix inconsistent representation for Grib Level

lesserwhirls:
Addresses Unidata/thredds#853

Grib1_Level_Type and Grib2_Level_Type were not consistent. Grib2_Level_Type was
the String description of the level, which is actual the analog of Grib1_Level_Desc.

Adds new attribute, Grib2_Level_Desc, and makes Grib2_Level_Type the integer code
value, just link Grib1_Level_Type.

### DIFF
--- a/grib/src/main/java/ucar/nc2/grib/collection/Grib2Iosp.java
+++ b/grib/src/main/java/ucar/nc2/grib/collection/Grib2Iosp.java
@@ -429,13 +429,11 @@ public class Grib2Iosp extends GribIosp {
     Grib2Customizer.Parameter entry = cust.getParameter(vindex.getDiscipline(), vindex.getCategory(), vindex.getParameter());
     if (entry != null) v.addAttribute(new Attribute("Grib2_Parameter_Name", entry.getName()));
 
-    if (vindex.getLevelType() != GribNumbers.MISSING) {
-      String levelTypeName = cust.getLevelName(vindex.getLevelType());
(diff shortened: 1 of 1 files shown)


grib/src/main/java/ucar/nc2/grib/collection/Grib2Iosp.java
```diff
@@ -429,13 +429,11 @@ protected void addVariableAttributes(Variable v, GribCollectionImmutable.Variabl
...
-    if (vindex.getLevelType() != GribNumbers.MISSING) {
-      String levelTypeName = cust.getLevelName(vindex.getLevelType());
-      if (levelTypeName != null)
-        v.addAttribute(new Attribute("Grib2_Level_Type", levelTypeName));
-      else
-        v.addAttribute(new Attribute("Grib2_Level_Type", vindex.getLevelType()));
-    }
+    if (vindex.getLevelType() != GribNumbers.MISSING)
+      v.addAttribute(new Attribute("Grib2_Level_Type", vindex.getLevelType()));
+    String ldesc = cust.getLevelName(vindex.getLevelType());
```
cwardgar: Should be `cust.getLevelDescription`

cwardgar: Nevermind, it looks like that method doesn't exist on `Grib2Customizer` like it does on `Grib1Customizer`. Weird.

Issue #854, opened 2017-06-08
Error constructing GRIB2 var names


msdsoftware:
I've heard from some NOAA people about trouble using Panoply to read some GRIB2 data, but IDV also has the problem. So assuming that the datasets are okay (and I'm assured that they are), then the problem would seem to lie in the NJ code for reading/constructing GRIB2 metadata.

The issue is that variable IDs and variable names in some datasets are reported with the wrong accumulation or averaging period, e.g.., the var names are shown with "XXX_36_Hour_Accumulation_XXX" when they should be "XXX_6_Hour_Accumulation_XXX".

Sample datasets are at
http://nomads.ncep.noaa.gov/pub/data/nccf/com/gefs_legacy/prod/gefs_legacy.20170606/00/pgrb2a/

An example that shows the var names as 36-hour is geavg.t00z.pgrb2af258

However, a dataset that shows them as the expected 6-hour is geavg.t00z.pgrb2af252

lesserwhirls:
Greetings @msdsoftware,

The key difference between the two files is the time unit encoded into the GRIB record. In file `geavg.t00z.pgrb2af252` (with anticipated behavior), octet 18 of the `PDS` for `Total Precipitation` indicates: 

```
 18: Indicator of unit of time range == 1 (table 4.4: Hour)
```

whereas in file `geavg.t00z.pgrb2af258` (unanticipated behavior), octet 18 for the same grid indicates:

```
18: Indicator of unit of time range == 11 (table 4.4: 6 hours)
```

In both cases, a time interval of `6` is computed based on the forecast time (octet 19) and the time interval end (octets 37-43), but once the time unit is applied, we get the discrepancy (6 hours vs 36 hours).

msdsoftware:
So do you think this constitutes a bug in the NJ code. Or if not necessarily a bug, then something that NJ might catch with a sanity test and try to "correct"?

I forwarded your comments to the NOAA people -- they said thanks much for the analysis. They are going to trying to figure out where/why in the processing chain some files are generated with octet 18 value of 11. The theory has been advanced that someone fat-fingered an 11 rather than a 1 when writing some code.

msdsoftware:
I received some follow-up from NOAA about the datasets that cause the unexpected behavior:

> It seems as though the unidata library in question may need to calculate that "I" value differently for forecast hours over 255.  The information that the developer provided regarding the octets in the PDS section are below:
> 
> As far as I can tell, the GRIB1 PDS time specifications are correct:
[...]

PR #861, opened 2017-06-13
Fix time interval name string construction

lesserwhirls:
Fixes Unidata/thredds#854

### DIFF
--- a/grib/src/main/java/ucar/coord/CoordinateTimeIntv.java
+++ b/grib/src/main/java/ucar/coord/CoordinateTimeIntv.java
@@ -86,7 +86,6 @@ public class CoordinateTimeIntv extends CoordinateTimeAbstract implements Coordi
       else if (value != firstValue) return MIXED_INTERVALS;
     }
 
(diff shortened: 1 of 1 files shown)


lesserwhirls:
Need to do a full Jenkins run before considering this one.

lesserwhirls:
Nope...will update with something better.

JohnLCaron:
Theres a bunch of work on 5.0 with regards to time interval grib naming.
Also, the issues are convoluted, and in some cases the results are unexpected.
Would be glad to review issues on current version of 5.0. 
4.6, not so much.

lesserwhirls:
Yes, from what I recall, the way time units (and perhaps time info more generally) are handled between GRIB1 and GRIB2 are not consistent. I've been meaning to get back to these issues, but if you see any you'd like to take a stab at on 5, then 👍 . I'm also happy to get you an account on Jenkins again if you'd like to be able to run the full test suite. In terms of GRIB, in my ideal world, 4.6 will likely get one more table refresh and that's it (actually, in my ideal world, 4.6 would be long gone).

PR #869, opened 2017-06-21
BUG: Fix geostationary projection (Fixes #799)

### LINKED ISSUE #799: Problem accessing GOES-16 full disk on TDS 5
So trying to hit some full disk data from NOAAPORT using NCSS:

http://thredds-test.unidata.ucar.edu/thredds/ncss/grid/satellite/goes16/GOES16/20170412/FullDisk/Channel01/GOES16_20170412_000037_0.47_1km_0.0S_89.5W.nc4/dataset.html

I get the following:
```
IllegalArgumentException: Must specify semi_minor_axis
```
Looking at the OPENDAP output, This would seem to be the problem:
```
grid_mapping_name: geostationary
latitude_of_projection_origin: 0.0
longitude_of_projection_origin: -89.5
semi_major: 6378137.0
semi_minor: 6356752.31414
perspective_point_height: 3.5785831E7
sweep_angle_axis: x
```
I have no idea if that's actually CF-compliant.

dopplershift:
While files with this projection would show their coordinate system
properly understood by toolsUI, NCSS would complain as the projection
was recreated from the internal object--dropping the perspective point.

Of course this problem traces back to a LOOK line.

I'm happy to add some tests if anyone has a suggestion on how to write it--it's not as simple as just doing the coordinate system build--there's some two-step process where that's converted back where the problem lies. (Or just running it through NCSS.)

### DIFF
--- a/cdm/src/main/java/ucar/unidata/geoloc/projection/sat/Geostationary.java
+++ b/cdm/src/main/java/ucar/unidata/geoloc/projection/sat/Geostationary.java
@@ -147,7 +147,7 @@ public class Geostationary extends ProjectionImpl {
     addParameter(CF.GRID_MAPPING_NAME, NAME);
     addParameter(CF.LONGITUDE_OF_PROJECTION_ORIGIN, navigation.sub_lon_degrees);
     addParameter(CF.LATITUDE_OF_PROJECTION_ORIGIN, 0.0);
-    // addParameter(CF.PERSPECTIVE_POINT_HEIGHT, navigation.sub_lon_degrees);   LOOK NOT USED ??
+    addParameter(CF.PERSPECTIVE_POINT_HEIGHT, navigation.sat_height * 1000.0);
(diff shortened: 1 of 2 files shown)


cdm/src/main/java/ucar/unidata/geoloc/projection/sat/Geostationary.java
```diff
@@ -147,7 +147,7 @@ private void makePP() {
     addParameter(CF.GRID_MAPPING_NAME, NAME);
     addParameter(CF.LONGITUDE_OF_PROJECTION_ORIGIN, navigation.sub_lon_degrees);
     addParameter(CF.LATITUDE_OF_PROJECTION_ORIGIN, 0.0);
-    // addParameter(CF.PERSPECTIVE_POINT_HEIGHT, navigation.sub_lon_degrees);   LOOK NOT USED ??
```
lesserwhirls: w00t!

lesserwhirls: made ya look, so I guess it worked! 😝

PR #871, opened 2017-06-22
JoinNew NcML Aggregations with scalar coordinate variables

dopplershift:
For joinNew aggregations, allow the use of an existing scalar variable with the same name, and aggregate it. This allows, for instance, an aggregation to create a new time dimension for CF-compliant files that have a scalar coordinate variable for time (since there's only a single time in the file)--and have the resulting aggregation have a the aggregated time variable on that new time dimension.

The second commit hoists up the code for handling changing time dimensions into the `AggregationOuterDimension` class so that this new type of joinNew aggregation can make use of them.


cwardgar:
Looks good. Nice job integrating your test with the existing ones. Can you add some documentation? It looks like `ncml/Cookbook.adoc` and `Aggregation.adoc` are the places to do it. You should be able to follow the examples of the other aggregation types that are already there.

dopplershift:
It seemed to me to fit as another option in the list given. Does this work for you?

cwardgar:
Yep, that works.

PR #874, opened 2017-07-10
Allow default cache sizes of Variables to be modified by user.

cwardgar:
* Variables need to be static, public, and non-final.
* From the [mailing list](http://www.unidata.ucar.edu/mailing_lists/archives/netcdf-java/2017/msg00054.html).

### DIFF
--- a/cdm/src/main/java/ucar/nc2/dataset/CoordinateAxis.java
+++ b/cdm/src/main/java/ucar/nc2/dataset/CoordinateAxis.java
@@ -73,7 +73,7 @@ import java.util.Formatter;
 
 public class CoordinateAxis extends VariableDS {
   static private org.slf4j.Logger log = org.slf4j.LoggerFactory.getLogger(CoordinateAxis.class);
-  static private final int axisSizeToCache = 100 * 1000; // bytes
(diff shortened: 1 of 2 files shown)


cdm/src/main/java/ucar/nc2/dataset/CoordinateAxis.java
```diff
@@ -73,7 +73,7 @@
 
 public class CoordinateAxis extends VariableDS {
   static private org.slf4j.Logger log = org.slf4j.LoggerFactory.getLogger(CoordinateAxis.class);
-  static private final int axisSizeToCache = 100 * 1000; // bytes
+  static public int axisSizeToCache = 100 * 1000; // bytes
```
lesserwhirls: I suppose there is no way to make this package private? If not, that's ok, just checking to make sure.

cwardgar: Not if we want the user to easily be able to modify it.

lesserwhirls: Ok, no worries.

PR #888, opened 2017-07-17
Update script.sh

lesserwhirls:


lesserwhirls:
So this one we should actually merge, as we no longer are building docs using asciidoctor. The others are just tests and can be closed.

cwardgar:
Any reason you're commenting-out instead of deleting? And what about the `adoc2html.sh` file?

lesserwhirls:
Uh...umm...that was a test? Yeah, that's it! It was a test! See, reviews work. Yay. Super good job #1. 👍 👍 👍 💯

cwardgar:
Travis is being dumb.

lesserwhirls:
I think I may need to open a new PR...

PR #897, opened 2017-07-20
MNT: bump spring dependencies

lesserwhirls:


cwardgar:
BTW, what does "MNT" mean? I've seen those three-letter codes on some of Ryan's commits as well. Is it some convention to describe commits?

lesserwhirls:
@cwardgar - just trying it out to see how it feels. @dopplershift uses 

 * `DOC` - documentation
 * `MNT` - maintenance
 * `BUG` - bug fix
 * `ENH` - enhancement

in the python projects. It could be a way to make release notes easer to categorize, especially `BUG` and `ENH`.  I could see adding another label, `API`, to indicate API changes...that one could be super useful to highlight. We could just use and modify github labels, which we already have some under the "type" tag.

PR #908, opened 2017-08-26
Enable individually defined datasets in an FMRC

lesserwhirls:
Allow individually defined datasets in an FMRC configured using NcML


lesserwhirls:
Waiting for Jenkins runs to complete, just to make sure nothing breaks...at least to the best of our ability to make sure nothing breaks!

lesserwhirls:
@guygriffiths - this should fix the FMRC in NcML problem :-)

lesserwhirls:
The sad part is that it's already documented in the netcdf-java docs, but I could not find a version of toolsUI that could open it (I went back to 4.2).
w00t

guygriffiths:
@lesserwhirls - Thanks, but it's still not fully working for me.  I've been testing with 2 separate methods, one for an FMRC defined in a file, which uses:
```
Fmrc fmrc = Fmrc.open(location, errlog);
NetcdfDataset nc = fmrc.getDatasetBest().getNetcdfDataset()
```
This returns `null` from the `Fmrc.open()` method, specifically when calling `NcmlCollectionReader.open()`.  It gets to line 113 of `NcmlCollectionReader`, where it doesn't find either a "scan" or a "scanFmrc" element, sends an appropriate message to the errorlog, and returns `null`.  However, I wouldn't expect either of these elements to be present when explicitly listing files.  My example FMRC file is pretty much the same as the one you have at:
https://github.com/lesserwhirls/thredds/blob/3e36b4586e282c56269412b24aefe1874c14e4c0/cdm/src/test/data/ncml/fmrc/testAggFmrcExplicit.ncml

The other method I've been trying is constructing the NcML on-the-fly and passing it to `NcMLReader.readNcML()`.  This returns successfully, using the constructed NcML:
```
<netcdf xmlns="http://www.unidata.ucar.edu/namespaces/netcdf/ncml-2.2" enhance="true">
    <aggregation dimName="run" type="forecastModelRunCollection" enhance="true">
        <netcdf location="/home/guy/Data/fmrctest/2017072206_Glb15Min_CMAP.nc" coordValue="2017-07-21T19:00:00.000+01:00" />
        <netcdf location="/home/guy/Data/fmrctest/2017072218_Glb15Min_CMAP.nc" coordValue="2017-07-22T07:00:00.000+01:00" />
        <netcdf location="/home/guy/Data/fmrctest/2017072306_Glb15Min_CMAP.nc" coordValue="2017-07-22T19:00:00.000+01:00" />
        <netcdf location="/home/guy/Data/fmrctest/2017072318_Glb15Min_CMAP.nc" coordValue="2017-07-23T07:00:00.000+01:00" />
    </aggregation>
</netcdf>
```
and the dataset has added the `run` dimension, as expected.  However, I'd then like to get the "best" version of the dataset - i.e. with no `run` axis, just the `time` axis populated with the data values from the final file which contains the time step..

So I guess there are two issues here:
 - The fact that `Fmrc.open()` doesn't work with explicitly-listed datasets.  This is not a big deal for me, I was only using the file-based dataset for testing, but it is the `Fmrc` class which has the `getDatasetBest()` method which I want.
 - Not being able to extract the best dataset from a NetcdfDataset which has been opened as an FMRC.  This is what I really need to do, but I could just be ignorant of *how* to do this, rather than it being a bug.

lesserwhirls:
Ok, I think I have this working. I just need to update the tests to make sure and make a new PR.

lesserwhirls:
Oh boy, that was fun. This is now working with Unidata/thredds#910 :-)

guygriffiths:
@lesserwhirls - Thanks, I've just had a chance to test, and it's looking close...

Currently if I define an FMRC in a file, with explicitly listed files, it works perfectly (using `Fmrc.open(location, errlog).getDatasetBest().getNetcdfDataset()`).

However, I'm not sure how I can get the same behaviour with the NcML defined as a string (i.e. with no NcML file), which is my main use case (it comes from trying to automatically interpret a glob expression where the files are part of an FMRC).  Looking through the code it doesn't appear to be possible at the moment.  For standard (i.e. not FMRC) NcML files, I use the `NcMLReader.readNcML(Reader r, CancelTask c)` method.  It would be really useful if there was a method `Fmrc.readNcML(Reader r, CancelTask c)` or similar, which returned an `Fmrc` object.

The following changes work for my situation:
https://github.com/guygriffiths/thredds/commit/4e95b9d620816334c483370383fd4190a18f42fd
but may cause issues if these methods get used in another situation (particularly passing `null` as the `ncmlLocation` for the (new) `NcmlCollectionReader.readXML()` method - I'm not really familiar enough with the overall codebase to know what impacts it might have).  I'd be grateful if you could take a look and see what else needs to be done before I'd get a pull request accepted.

One final request - once this is all working and tested, could you do another 5.0.0 release (probably 5.0.0-alpha2?), so that we have a permanent version to work from?

lesserwhirls:
Greetings @guygriffiths - this looks ok to me. The only suggestion I would have is to add a test, but I can do that. If you make a PR, I will run it against our full test suite and, if that passes, then I say we are good to go. I don't mind cutting a 5.0.0-alpha2 once this change is merged. Thanks for iterating on this with me!

guygriffiths:
Excellent, thanks @lesserwhirls - I've just put in that pull request now :)

lesserwhirls:
Ok @guygriffiths - you should be good to go with `<version>5.0.0-alpha2</version>`

Cheers!

guygriffiths:
@lesserwhirls - Perfect, thanks :)

lesserwhirls:
Woah! @cwardgar - any idea why these two modules ballooned in size?

cwardgar:
Fixed in #916. Awaiting review and merge.

lesserwhirls:
Ok, I think we have it now. @guygriffiths - try the 5.0.0-alpha3 artifacts. Thanks @cwardgar!

guygriffiths:
Excellent, thanks.  Just released ncWMS 2.2.10 with these libs. :)

lesserwhirls:
woohoo 👍 - thanks for sticking with us through this!

guygriffiths:
No worries, thanks for fixing it so quickly!

PR #910, opened 2017-08-29
Allow non-scan based FMRC ojbects

lesserwhirls:
Allow explicility listed files in NcML FMRC to be read in as FMRC objects

Allow https namespaces in NcMLCollectionReader

### DIFF
--- a/cdm/src/main/java/thredds/inventory/NcmlCollectionReader.java
+++ b/cdm/src/main/java/thredds/inventory/NcmlCollectionReader.java
@@ -1,35 +1,8 @@
 /*
- * Copyright 1998-2015 University Corporation for Atmospheric Research/Unidata
- *
- *   Portions of this software were developed by the Unidata Program at the
- *   University Corporation for Atmospheric Research.
- *
- *   Access and use of this software shall impose the following obligations
- *   and understandings on the user. The user is granted the right, without
- *   any fee or cost, to use, copy, modify, alter, enhance and distribute
- *   this software, and any derivative works thereof, and its supporting
- *   documentation for any purpose whatsoever, provided that this entire
- *   notice appears in all copies of the software, derivative works and
- *   supporting documentation.  Further, UCAR requests that the user credit
- *   UCAR/Unidata in any publications that result from the use of this
- *   software or in any product that includes this software. The names UCAR
- *   and/or Unidata, however, may not be used in any advertising or publicity
- *   to endorse or promote any products or commercial entity unless specific
- *   written permission is obtained from UCAR/Unidata. The user also
- *   understands that UCAR/Unidata is not obligated to provide the user with
- *   any support, consulting, training or assistance of any kind with regard
- *   to the use, operation and performance of this software nor to provide
- *   the user with any updates, revisions, new versions or "bug fixes."
- *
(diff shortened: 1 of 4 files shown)


lesserwhirls:
Will start a run on Jenkins to make sure life is good-ish.

lesserwhirls:
Ok, no new failures on Jenkins, so I think we're good.

cdm/src/test/java/ucar/nc2/ncml/TestAggFmrc.java
```diff
@@ -159,6 +176,95 @@ public void testAggDataValuesWithOrig() throws IOException {
...
+
+    try {
+      List<CalendarDate> scanRunTimes = fmrcScan.getRunDates();
+      List<CalendarDate> explicitRunTimes = fmrcExplicit.getRunDates();
+
+      Assert.assertEquals(scanRunTimes.size(), 3);
+      Assert.assertEquals(explicitRunTimes.size(), 3);
+
+      for (int i = 0; i < 3; i++) {
+        scanRunTimes.get(i).compareTo(explicitRunTimes.get(i));
```
cwardgar: This is a no-op.

cdm/src/main/java/thredds/inventory/NcmlCollectionReader.java
```diff
@@ -108,12 +90,12 @@ static public NcmlCollectionReader open(String ncmlLocation, Formatter errlog) t
...
-    if (scanElem == null) {
-      errlog.format("NcML must have aggregation scan or scanFmrc element");
-      return null;
-    }
+    Element scanElem = aggElem.getChild("scan", myNS);
+    //if (scanElem == null) scanElem = aggElem.getChild("scanFmrc", Catalog.ncmlNS);
+    //if (scanElem == null) {
+    //  errlog.format("NcML must have aggregation scan or scanFmrc element");
+    //  return null;
+   //}
```
cwardgar: Can this block be removed? Commented-out code is a form of technical debt that we have a lot of already.

cdm/src/test/java/ucar/nc2/ncml/TestAggFmrc.java
```diff
@@ -159,6 +176,95 @@ public void testAggDataValuesWithOrig() throws IOException {
...
+
+    try {
+      List<CalendarDate> scanForecastTimes = fmrcScan.getForecastDates();
+      List<CalendarDate> explicitForecastTimes = fmrcExplicit.getForecastDates();
+
+      Assert.assertEquals(scanForecastTimes.size(), 24);
+      Assert.assertEquals(explicitForecastTimes.size(), 24);
+
+      for (int i = 0; i < 3; i++) {
+        scanForecastTimes.get(i).compareTo(explicitForecastTimes.get(i));
```
cwardgar: This too.

lesserwhirls:
Good catch on the no-ops...uncovered a bug dealing with the correct identification of runtime, and that the `coord` attribute, which is the explicit runtime value for each explicitly listed dataset, was not being applied correctly. w00t. I'll also re-run jenkins, to be safe.

lesserwhirls:
Ok, again, no new failures. Yay.

PR #914, opened 2017-09-07
Read Fmrc directly from string

guygriffiths:
Minor modifications to allow `Fmrc` to read an NcML forecast model run collection directly from a string, in addition to the existing ability to read it from a file.


lesserwhirls:
So far so good - I'm running the full test suite on Jenkins now and will report back when it's finished. Thanks @guygriffiths!

lesserwhirls:
There was one new failure in our tests, but it's unrelated to this PR.

PR #916, opened 2017-09-13
Don't include Gretty overlays in final project archives

cwardgar:
Fixes issue from https://github.com/Unidata/thredds/pull/908#issuecomment-328892503

Other miscellaneous changes:
* Use Markdown's automatic list numbering, instead of doing it manually.
* Add SonarQube and Coveralls badges to README.


cwardgar:
You can see the new badges here: https://github.com/Unidata/thredds/blob/fee916a549f7461a5589ae64af8bbbadf4a608c6/README.md

I added them so that our SonarQube and Coveralls results are more visible to US, as I noticed a tendency for them to be forgotten, probably because we're not getting notifications. Then again, the numbers don't exactly make us look good.

Add the badges or no?

lesserwhirls:
Wow, so you do have to review your own code if you are the code owner *laughs*

cwardgar:
Hmm, so you can't merge?

lesserwhirls:
Here is what I see

screen shot 2017-09-13 at 9 19 12 am

lesserwhirls:
I forgot to add - let's keep the badges. Good or bad, no matter.

Issue #907, opened 2017-08-25
Opening datasets with insufficiently defined projected grids


msdsoftware:
Twice within a week, I have heard from users of my Panoply application about netCDF datasets that cannot be opened. In a nutshell, the problem is that in both cases Panoply tries to acquire the dataset in enhanced mode, and that the datasets use projected lon-lat grids which are apparently considered insufficiently defined. In other words, the `grid_mapping` variable does not include one or more attributes that the netCDF-Java library requires be there, and so an exception is thrown when NJ tries to construct the coordinate system and the dataset is not acquired/opened.

In one case, the exception is "IllegalArgumentException: Mercator projection must have attribute standard_parallel or scale_factor_at_projection_origin", thrown at ucar.nc2.dataset.transform.Mercator, line 46.

In the second case, the exception is "IllegalArgumentException: Must specify semi_minor_axis", thrown at ucar.nc2.dataset.transform.Geostationary, line 111. (In this case, the dataset had an attribute named semi_minor, rather than semi_minor_axis.)

Since the exceptions block the datasets from even being opened, there is nothing I can do in my own code to trap for them. The question then is, is there a reasonable way to modify the NJ library code to allow the datasets to be opened? For example, rather than through an exception, could NJ instead apply default values for missing projection attributes, possibly logging a warning that the gridding results might not be what the user expects?

Note that issue #611 somewhat addressed this problem but only for the case of the `LambertCylindricalEqualArea` projected grid.

PR #917, opened 2017-09-13
Geostationary -- Calculate semi_minor_axis or inv_flattening

msdsoftware:
If either of semi_minor_axis or inv_flattening is present in the grid_mapping attributes but other is not, then the missing value can be calculated from the other and from the semi_major_axis.


cwardgar:
In the event that both `semi_minor_axis` and `inv_flattening` are present, your final `else` block will be executed. Is that your intent?

You've got 4 possible outcomes based on the presence of the two variables. To me, it's clearer to list them explicitly, e.g.:
```java
      if (Double.isNaN(semi_minor_axis) && Double.isNaN(inv_flattening)) {
         throw new IllegalArgumentException("Must specify "+CF.SEMI_MINOR_AXIS+" or "+CF.INVERSE_FLATTENING);
      } else if (Double.isNaN(semi_minor_axis)) {
          final double flattening = 1. / inv_flattening;
          semi_minor_axis = semi_major_axis * (1. - flattening);
      } else if (Double.isNaN(inv_flattening)) {
          final double flattening = (semi_major_axis - semi_minor_axis) / semi_major_axis;
          inv_flattening = 1. / flattening;
      } else {
        assert !Double.isNaN(semi_minor_axis) && !Double.isNaN(inv_flattening);
        if (semi_minor_axis == semi_major_axis) {
          // Do nothing. inv_flattening = 1. / 0.
        }
      }
```

cwardgar:
Updated code snippet. Also, should it be an error if we can't calculate `inv_flattening` due to division by zero?

msdsoftware:
My bad about blocking on the case of both being present. (Gives self head slap.)

I wasn't quite sure what to do about the case of semi_minor = semi_major, or if anything needed to be done. The thing is, if that condition occurs, then inv_flattening would have already been set to NaN above. One could argue it should instead be reset to POSITIVE_INFINITY.

The cases I have run into for using the geostationary projection have all been "real" data and have specified the spheroid attributes. It would be more than a bit odd for such data to specify data on the sphere rather than a spheroid.

msdsoftware:
What's the easiest way for me to correct the code in my pull request?

cwardgar:
In that case, I think that emitting an error if `semi_minor == semi_major` is reasonable. Or maybe just an assertion?
```java
assert semi_minor == semi_major : "Should never happen with real data!";
```
Any change you make to the `msdsoftware:master` branch will automatically be incorporated into this PR. So go ahead and just add another commit. Then, if you want a cleaner history, you can do an interactive rebase. It should all get picked up here.

msdsoftware:
I've updated the code.

I left the case of semi-minor == semi_major as "do nothing" -- i.e., leave inv_flattening = NaN -- due to the possibility that someone might generate a sample dataset using the sphere. A very slim possibility but it's there.

Also, throwing an exceptionwould block the dataset from being acquired, and I'm already running into elsewhere with projected grid datasets that cannot be acquired/opened because grid mapping projection attributes are under- or poorly specified (cf #907).

msdsoftware:
@cwardgar, Travis seems to have borked building after I posted that code fix, but I can't tell whether it's something I got wrong in the code or if it's something else. It builds fine with Gradle on a local copy.

cwardgar:
It's some sort of weird error in Travis's JVM v7. That's not our fault.

Can you close and reopen this PR? That should force Travis to rerun the checks.

dopplershift:
You can restart Travis from the web UI, I just did it.

cwardgar:
This seems to be a problem with Travis. Here is the last successful Travis build for the master branch:
https://travis-ci.org/Unidata/thredds/jobs/269734167

Here is a build *of the same commit* that I reran just a couple hours ago, now failing:
https://travis-ci.org/Unidata/thredds/jobs/269734167

The difference? `java version "1.7.0_75"` vs `java version "1.7.0_121"`. Looks like Travis updated its JDK and things broke. I'll look more into this later. I wonder if the build would break using a local copy of 1.7.0_121?

cwardgar:
I added a commit in #919 and then merged it into this branch. Fixed! So this is good to go now.

PR #921, opened 2017-09-16
Create only one MavenPublication per project

cwardgar:
Previously, webapp projects created two MavenPublications: 'java' and 'web', each with their own POM. When I tried to publish both to the new Nexus repo, the second publication was refused because I was trying to overwrite the first POM and repository policy forbids redeployment. Now, all artifacts for a project are included in one, big MavenPublication.

Updated tests to account for the change.


lesserwhirls:
Does the 5.0 branch "do the right thing" here, or will it need to change, too?

cwardgar:
No, it doesn't, and I need the change to be on 5.0 soon as well. I'll try to merge master into 5.0 later today.

Issue #929, opened 2017-10-03
This line generates NPE if org is set to null in the while loop just above this line.


ghansham:
https://github.com/Unidata/thredds/blob/a05e9fc00ac3ac7fab215c3db053fada57678675/cdm/src/main/java/ucar/nc2/dataset/VariableDS.java#L280

ghansham:
This is specific to NcML only.

DennisHeimbigner:
Do we have a test program for this so I can see if pull request
https://github.com/Unidata/thredds/pull/930
also fixes this.

ghansham:
I have a fix for that within variableDS class. But that may be a temporary fix. But still I will share it with you

ghansham:
The variables were unsigned short variables

cwardgar:
I was able to recreate this on 5.0.0 with the following NcML:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<ncml:netcdf xmlns:ncml="http://www.unidata.ucar.edu/namespaces/netcdf/ncml-2.2" enhance="all">
    <ncml:variable name="my_var" shape="" type="ushort">
        <ncml:attribute name="long_name" value="my_var_name" />
        <ncml:attribute name="_FillValue" type="ushort" value="65535" />
    </ncml:variable>
</ncml:netcdf>
```
The fix seems pretty straightforward.

ghansham:
What happens if my data is 10 bit and my fillvalue is 0. And I have another
variable where fillvalue is 1023. And I hv this fillvalue defined along
with variable as per CF conventions.

Regards
Ghansham

DennisHeimbigner:
Ghansham- Can you give specific ncml for your 10bit example?

ghansham:
Sure..i will send ncml and data as well.
Although I have not checked the behavior after the recent commits.

DennisHeimbigner:
The proximate cause is at VariableDS line 2803:
> needUnsignedConversion = (this.getDataType().isUnsigned() && !org.getDataType().isUnsigned());
this generates NPE because 'org' is null.
However, this occurs because of the code at  NcMLReader line 1218
> v = new VariableDS(ds, g, parentS, name, dtype, shape, null, null);
This seems incorrect to me. Instead, we should be creating an instance of Variable
instead of VariableDS.

cwardgar:
I already have a fix for this:
```java
    // do we need to do unsigned conversion ?
    Variable ultimateOrigVar = this;
[...]

Issue #923, opened 2017-09-19
Ncml mishandling signedness


DennisHeimbigner:
re: e-support PCM-232977
Given test.ncml:
````
    <?xml version="1.0" encoding="UTF-8"?>
    <ncml:netcdf xmlns:ncml="http://www.unidata.ucar.edu/namespaces/netcdf/ncml-2.2">
    <ncml:variable name="my_var" shape="" type="short">
    <ncml:attribute name="long_name" value="my_var_name" />
    <ncml:attribute name="_FillValue" type="short" value="65535" isUnsigned="true" />
    <ncml:attribute name="_Unsigned" value="true" />
    </ncml:variable>
    </ncml:netcdf>
````
The command:
````j
ava -Xmx1g -classpath netcdfAll-4.6.10.jar ucar.nc2.write.Nccopy --input test.ncml --output test.nc --format netcdf4 --deflateLevel 0
````
produces:
````
    netcdf test {
    variables:
        ushort my_var ;
            my_var:long_name = "my_var_name" ;
            my_var:_Unsigned = "true" ;
    data:
        my_var = 32769 ;
    }
````
Appear to be two problems:
1. my_var should be of type short, not ushort
2. _FillValue should be defined and as ushort

Looks like ncml is not handling signedness correctly.
Correction: the bug may be in Nc4Iosp or Nccopy.

DennisHeimbigner:
Fixed in pull request https://github.com/Unidata/thredds/pull/930

PR #930, opened 2017-10-04
Fix ncml handling of isUnsigned xml atribute

DennisHeimbigner:
re: issue https://github.com/Unidata/thredds/issues/923
NcMLReader appears to be ignoring the isUnsigned xml attribute.
in the master branch. Added code to make it be recognized.
Note that this change should not be propagated to 5.0.

### DIFF
--- a/cdm/src/main/java/ucar/nc2/ncml/NcMLReader.java
+++ b/cdm/src/main/java/ucar/nc2/ncml/NcMLReader.java
@@ -562,6 +562,14 @@ public class NcMLReader {
       if (debugConstruct) System.out.println(" add new att = " + name);
       try {
         ucar.ma2.Array values = readAttributeValues(attElem);
+        // Check for unsigned
+        String signedness = attElem.getAttributeValue("isUnsigned");
+        boolean unsigned = false;
+        if(signedness != null && (
+                signedness.equalsIgnoreCase("true")
+                || signedness.equalsIgnoreCase("1")))
+          unsigned = true;
+        values.setUnsigned(unsigned);
         addAttribute(parent, new ucar.nc2.Attribute(name, values));
       } catch (RuntimeException e) {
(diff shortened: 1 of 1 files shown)


cwardgar:
So this issue doesn't affect 5.0? I was going to ask for a unit test, but I guess that's not really applicable if the problem is fixed in some other way there.

cwardgar:
Also, does this fix #929? If not, the 2 issues are at least closely related. Seems worthwhile to try to knock out both at the same time.

DennisHeimbigner:
Good question. I need a test program for that issue to see.

DennisHeimbigner:
The NcmlReader code in 5.0.0 is significantly different.
It appears to allow the unsigned types (e.g. ushort)
directly, so the problem should not occur. I confess
I have not tried it to check. In any case, if a fix is
needed in 5.0.0 it will look quite different to the 4.x change
in this pr.

ghansham:
I might hv a sample datasets and corresponding ncml to reproduce this error.  Will try to share with you. If not possible I will test it at my end.

DennisHeimbigner:
I did a quick test using the original test.ncml, but with short->ushort.
Nccopy properly creates the .nc file with ushort and with the unsigned
value for _FillValue (65535).  However, Ncdumpw prints out the _Fillvalue
as -1. So there is an error in handling (at least) unsigned short in the print
code used by Ncdumpw.

ghansham:
I basically encountered NPE issue when I tried to load samle ncml existing time dimension aggegation with unsigned variables. The input files where CF complaint hdf5 files.

cwardgar:
This fix is appropriate for 4.6. However, if merged into master, it'll eventually find its way into 5.0.0, where it is NOT appropriate, due to extensive changes vis-a-vis unsigned types. To avoid conflicts, I propose that we leave these changes out and satisfy ourselves with the fact that #923 is fixed in 5.0.0 by #934.

Issue #932, opened 2017-10-16
HdfEos handling of featureType attribute


msdsoftware:
I have a couple sample HDF-EOS files containing what is essentially featureType trajectory data. However, netCDF-Java is adding a `:featureType = "PROFILE"` global attribute to the files. Poking around in the code, I find that this comes from the `HdfEos` class examining the geolocation info, finding that the lat and lon axes are not 2D and defaulting to PROFILE:

```
    if ((latAxis != null) && (lonAxis != null)) {
        int xyDomainSize = CoordinateSystem.countDomain(new Variable[]{latAxis, lonAxis});
        if (xyDomainSize < 2) featureType = FeatureType.PROFILE;  // ??
    }
```

Note the double question mark in the comment. It would seem that when this written, someone wasn't sure that they were doing the right thing?

I'm looking at fixing this by having the geolocation examination also look for a time axes, and then replacing the above with

```
        if ((latAxis != null) && (lonAxis != null)) {
            log.trace ("found lonAxis and latAxis");
            List<Dimension> xyDomain = CoordinateSystem.makeDomain(new Variable[] {latAxis, lonAxis});
            log.trace ("xyDomain {}", xyDomain);
            if (xyDomain.size() < 2) {
              if (timeAxis != null)
              {
                Dimension dd1 = timeAxis.getDimension(0);
                Dimension dd2 = latAxis.getDimension(0);
                Dimension dd3 = lonAxis.getDimension(0);
             
                if (dd1.equals (dd2) && dd1.equals(dd3)) {
                    featureType = FeatureType.TRAJECTORY;  // ??
                } else {
                    featureType = FeatureType.PROFILE;  // ??
                }
              } else {
                featureType = FeatureType.PROFILE;  // ??
              }
            }
        }
```

This solution isn't perfect — certainly it omits the possibility of other featureTypes — but it should be an improvement over what's in `HdfEos` now.

Any comments, thoughts, criticism?

lesserwhirls:
I think you've interpreted the question marks correctly - it's much worse to see:

~~~java
// LOOK - WTF?!?
~~~

which generally means the author didn't know why or could not remember why that line was needed or what it was trying to do.

[...]

PR #933, opened 2017-10-17
Add trajectory FT detection to HdfEos, clean up code a bit

msdsoftware:
This adds detection of the trajectory featureType to the HdfEos class discussed in #932.

I also did some code clean-up in that class: wrapping all if blocks in braces, systematic indentation in a couple methods, and converting all System.out.println/printf to log.debug. I left in the showWork boolean on the debugs but it's redundant and could be removed in a future update.


lesserwhirls:
The test suite results look good, so I'll go ahead and merge. Thank you @msdsoftware!

PR #934, opened 2017-10-18
Fix various issues with unsigned data types and NcDumpW.

### LINKED ISSUE #651: ToolsUI and Unsigned variables
It would be really appreciated if the ToolsUI could handle unsigned values as at the moment the results end up looking odd as it's just printing the signed version.

I know that Java 8 has the "toUnsignedString" which would work well for displaying the values, but also understand if the source is targeted to old java version.

An alternative may be to use [Gauva's unsigned int/long's](https://github.com/google/guava/wiki/PrimitivesExplained#unsigned-support) to handle it as it doesn't require java 8

### LINKED ISSUE #929: This line generates NPE if org is set to null in the while loop just above this line.

cwardgar:
Fix NPE when enhancing unsigned variable that is newly created in NcML. Closes #929.
* Such variables return null for VariableDS.getOriginalVariable(). Previously, the code did not account for that possibility.

Correctly interpret unsigned data as unsigned when printing it. Closes #651.
* This applies to NcDumpW.printArray(), Array.toString(), and Attribute.writeCDL().
* Remove the "-unsigned" flag from NcDumpW. The functionality it enabled is now the default. All scripts, tests, and documentation that used or mentioned the flag have been updated.
* In DataType, add support for widening unsigned longs to BigIntegers. Test included. Also add a generic widenNumber() method that will widen any integral data type.
* Refactor MAMath.convertUnsigned() to support unsigned longs and to use DataType.widenNumber(). So we can now widen arbitrary Arrays of integral data types. An extensive test is included.
* Remove NcDumpW.fixUnsigned(). We already have methods in DataType that do the exact same thing.

When a variable is enhanced, if the "_Unsigned"=="true" attribute is found, the variable's data type will be changed to the unsigned equivalent.
* For example, a "short" variable will be changed to a "ushort" variable.

Fix NcDumpW formatting and indentation.
* This was mostly about creating an Indent object, incrementing and decrementing it as appropriate, and making sure to use it every time something is printed.

Together, these fixes resolve the eSupport ticket PCM-232977 and the GitHub issue that was created from it, #923.
* Add a regression test that asserts that when the example dataset from that ticket is ncdumped, the correct output is written.
* The test lives in TestNcDump, which was renamed from TestDump. The file used to contain a really bad, old test, which I've nuked.

Replace all instances of the string "_Unsigned" with the static variable CDM.UNSIGNED.


cwardgar:
@DennisHeimbigner Can you take a look at this? I've made some changes that might conflict with what you've done in the past. In particular, NcDumpW has different whitespace formatting in its output. Are there certain whitespace requirements that we must match in order to be compatible with the C lib's ncdump and ncgen?

cdm/src/main/java/ucar/nc2/dataset/EnhanceScaleMissingImpl.java
```diff
@@ -133,10 +133,15 @@
...
 
-    // the other possibility is that you want to apply scale and offset to a signed value, then declare the result unsigned
-    // this.isUnsigned = (orgVar != null) ? orgVar.isUnsigned() : forVar.isUnsigned();
-    this.isUnsigned = isUnsigned(forVar);  // LOOK switch to unsigned var
-    this.convertedDataType = forVar.getDataType().withSign(this.isUnsigned);  // only for netcDF !!??
+    // deal with legacy use of attribute with Unsigned = true
+    Attribute unsignedAttrib = forVar.findAttributeIgnoreCase(CDM.UNSIGNED);
+    boolean isUnsignedSet = unsignedAttrib != null && unsignedAttrib.getStringValue().equalsIgnoreCase("true");
+    if (isUnsignedSet) {
+      forVar.setDataType(forVar.getDataType().withSign(true));
```
DennisHeimbigner: Is this correct; should it it be withSign(flase)?

cwardgar: It is correct, but `DataType.withSign(boolean)` is needlessly confusing. Intuitively, `withSign(boolean)` ought to produce a *signed* data type, right? But it actually does the opposite.

I've refactored that method to accept a `Signedness` enum instead. See my new commit.

DennisHeimbigner:
When I tried to execute TestNcDump, it failed in a peculiar way.
The relevant output was:
> org.junit.ComparisonFailure:  <Click to see difference>
> at ucar.nc2.TestNcDump.testNcDumpUnsigned(TestNcDump.java:61)
What is weird is that if I click the " <Click to see difference>"
It brings up a comparison window that says that the two strings are identical,
but the assertion fails anyway.

cwardgar:
That strikes me as a line ending issue, given that the test works on Mac (my box) and Linux (Jenkins). Maybe I can make the test smarter; it's a simple string comparison right now.

DennisHeimbigner:
The obvious thing to do is preprocess the two strings to remove carriage returns ('\r').

DennisHeimbigner:
One more thing. You might add a test of NcDumpw with a file that has
nested groups to make sure that your indenting is correct.

cwardgar:
Okay, I've fixed the failure on Windows. Also, I've added a test case with nested groups.

Given the NcML input:
```xml
<netcdf xmlns="http://www.unidata.ucar.edu/namespaces/netcdf/ncml-2.2">
    <dimension name="time" length="2" isUnlimited="true" />

    <attribute name="global_att" type="String" value="foo" />

    <variable name="time" shape="time" type="int">
        <attribute name="units" type="String" value="hours" />
        <values>6 18</values>
    </variable>

    <group name="outer_group">
        <dimension name="lat" length="3" />

        <attribute name="outer_att" type="String" value="bar" />

        <variable name="lat" shape="lat" type="float">
            <attribute name="units" type="String" value="degrees_north" />
            <values>41.0 40.0 39.0</values>
        </variable>

        <group name="inner_group">
            <dimension name="lon" length="4" />

            <attribute name="inner_att" type="String" value="baz" />

            <variable name="lon" shape="lon" type="float">
                <attribute name="units" type="String" value="degrees_east" />
                <values>-109.0 -107.0 -105.0 -103.0</values>
            </variable>
        </group>
    </group>
</netcdf>
```
The output of NcDumpW will be:
```
netcdf file:../cdm/src/test/data/testNestedGroups.ncml {
  dimensions:
    time = UNLIMITED;   // (2 currently)
  variables:
    int time(time=2);
      :units = "hours";

  group: outer_group {
    dimensions:
      lat = 3;
    variables:
      float lat(lat=3);
        :units = "degrees_north";

    group: inner_group {
      dimensions:
        lon = 4;
      variables:
        float lon(lon=4);
          :units = "degrees_east";

      // group attributes:
      :inner_att = "baz";
    }

    // group attributes:
    :outer_att = "bar";
  }

  // global attributes:
  :global_att = "foo";

  data:
    time = 
      {6, 18}
    outer_group/lat = 
      {41.0, 40.0, 39.0}
    outer_group/inner_group/lon = 
      {-109.0, -107.0, -105.0, -103.0}
}

```

lesserwhirls:
Awesome 👍

PR #935, opened 2017-10-19
Return CDL notation from Group.toString (master)

msdsoftware:
The toString() method in ucar.nc2.Group returned only the group short name rather than the CDL description as do other dataset structure classes.


cwardgar:
Also good. Thanks!

PR #939, opened 2017-10-28
Fix dataset tracker bug for dataset element

lesserwhirls:
datasets added via the dataset element were not being added to the dataset tracker for cases where default services were being used.


lesserwhirls:
This fix allows restricted access to work properly for `<dataset>` elements. Because the `<dataset>` elements were not being tracked, the `resourceControlOk` method of `DatasetManager` was unable to detect that they were restricted (see [here](https://github.com/Unidata/thredds/blob/5.0.0/tds/src/main/java/thredds/core/DatasetManager.java#L405) for where the restriction check is done). This only affected `<dataset>` elements without explicit services (i.e. the new default service code in v5).

PR #942, opened 2017-11-08
Use new Nexus v3 repos at artifacts.unidata.ucar.edu

cwardgar:
* Explicitly list each and every repository from which we download dependencies. This means eschewing the "unidata-all" group repo!
* Drop "geoapi-pending" dependency from tds. It's actually just a dep that "ncwms" needs.
* Update docs and all other files to use the new endpoints.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,8 +26,8 @@ Source code for each of these is available from GitHub at
 The latest released and snapshot software artifacts (.jar and .war files e.g.)
 are available from Unidata's Maven repositories
 
-* https://artifacts.unidata.ucar.edu/content/repositories/unidata-releases/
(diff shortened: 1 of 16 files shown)


cwardgar:
This PR is in anticipation of the new Nexus v3 server being available at https://artifacts.unidata.ucar.edu. **That has not happened yet.** However, I hope for the DNS switch to take place this week. Expect for the Travis build to fail until then.

So yeah, don't merge this yet. Similar changes will have to be made for 5.0.0, but I want to get this into master and then merge master into 5.0.0 first. Then I'll fix any remaining issues.

cwardgar:
Okay, this is good to go.

PR #947, opened 2017-11-09
Clean-up of tests

lesserwhirls:
This PR cleans up a few things with our tests:

1. Tests that look at the number of grids in real time datasets and check to make sure that the number does not change - removed, as this should be done as part outside of the main tests
2. Some remote datasources had moved and needed to be updated. Should consider making these local at some point.
3. Added logging with `log4j` to `cdm-tests`. The logger name is `testLogger`, and it is configured to print to the console at the `DEBUG` level. `System.out.print*` methods were converted to `DEBUG` messages for the `testLogger` log. This preserves current behavior, but will allow us to tune the `log4j.xml` config as needed to cut down on noise.


lesserwhirls:
Removed 14 failing "real time" tests
Fixed 3 tests by updating remote resource locations.
Fixed remote resource for one test which seems to legit be failing now.

This should take the number of failing 5.0.0 tests on Jenkins from 46 to 29.

cwardgar:
You've got to be careful when converting `printf()` to slf4j. Two reasons:
1. slf4j has different placeholders. So instead of using `%s`, `%d`, `%f`, etc, you should use `{}`.
2. Every log message will be printed with a newline appended, so it's unnecessary to add your own. Besides, slf4j won't know how to interpret `%n` anyway.

So, a line like:
`System.out.printf("open %s%n", filename);`
ought to be:
`log.debug("open {}", filename);`
Alternatively, if you want to retain the printf format string, you could do something like:
`log.debug(String.format("open %s", filename));`

lesserwhirls:
Thanks @cwardgar - I think I have it updated correctly now.

PR #951, opened 2017-11-13
Merge master into 5.0.0 and use new Nexus v3 repos on 5.0.0

cwardgar:
5.0.0 ought to build on Travis now. *fingers crossed*

I also updated all remaining hyperlinks that point to artifacts.unidata.ucar.edu.


cdm/src/main/java/ucar/nc2/Group.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 1998-2017 John Caron and University Corporation for Atmospheric Research/Unidata
```
lesserwhirls: I would leave John in this copyright line since he made changes after he left. I don't think it matters too much for contributors who have not added their name to the copyright explicitly, as they still keep their copyright regardless, but I don't think we should remove it if it is explicitly added. Just my 2c.

DennisHeimbigner: I disagre. As far as we know John was still an employee when those changes were made.

cwardgar:
Yeah, I suspected that the godiva2 stuff wasn't needed in 5.0.0. But when I did the merge, there was already a bunch of stuff in `tds/src/main/webapp/godiva2`.

cdm/src/main/java/ucar/nc2/Group.java
lesserwhirls: Ok, that makes sense then.

cwardgar:
The `HdfEos` changes come from #933. Most of it is just reformatting and converting System.out to logger.debug(). There are about 20 novel lines added, dealing with detection of trajectories. Seems okay.

cwardgar:
As usual, this will require a merge commit.

lesserwhirls:
I can take care of this as soon as I get back to my computer (at a Dr.
Appt).

Issue #948, opened 2017-11-10
Group.getAttributes() returns secret attribute


TomBlock:
When requesting attributes from the root-group the secret attribute "_NCProperties" is returned, among the other custom attributes.

We are copying the content of an already existing file to a newly created (to add/modify variables), a code segment copying the root-attributes is like:

`for (Attribute att : rootGroup.getAttributes()) {`
`     System.out.println(att.getFullName());` 
`     newGroup.addAttribute(att);` 
`}`

The console dump is:

_NCProperties
title
institution
contact
... etc

When creating the new file with NetcdfFileWriter.create() an exception is thrown, because setting the "_NCProperties" attribute fails - it already exists.

java.io.IOException: -42 (NetCDF: String match to name in use) on attribute '_NCProperties = "version=1|netcdflibversion=4.4.1.1|hdf5libversion=1.10.1"' on global or group attribute

In our opinion, the secret attribute "_NCProperties" should be suppressed in a call to Group.getAttributes(), as also stated in the documentation (http://www.unidata.ucar.edu/software/netcdf/docs/attribute_conventions.html)

DennisHeimbigner:
The problem is that clients need to be able to get the value of _NCProperties
(and other hidden) attributes. In the netcdf-c library, as noted, we did this by
allowing the client to ask for these special attributes only by name. Asking by
e.g. attribute id suppressed them. We will probably need to do something
like that here. I suppose a short term solution is to just ignore attempts to
add a special attribute to a file.

DennisHeimbigner:
BTW, I am guessing that you are using version 4.6.x and not 5.0.0.
5.0.0 has code to ignore attempts to write such attributes.

TomBlock:
Hi Dennis,
you`re right, I forgot to mention:
we are using version 4.6.10 of the java lib, v 4.4.1.1 of netcdf.so and 1.10.1 of hdf5.so

And I agree, the client should explicitely ask for the secret attributes. I have modified our code to suppress copying the _NCProperties attribute - easy enough to overcome the problem.

Tom

DennisHeimbigner:
I am adding fixes to both 4.6 and 5.0 to address this problem.
Thanks for catching it.

DennisHeimbigner:
Fixed by pull requests:
https://github.com/Unidata/thredds/pull/949
and
https://github.com/Unidata/thredds/pull/952

PR #952, opened 2017-11-13
Make special attributes accessible by name only; version 5

DennisHeimbigner:
Re: git pull request https://github.com/Unidata/thredds/pull/949
Re: git issue https://github.com/Unidata/thredds/issues/948

This adds the change of pull request https://github.com/Unidata/thredds/pull/949
to branch 5.0.0. Changes in 5.0 code make it impossible to cherry pick.

From master pull request:
Currently, special attributes like _NCProperties are visible
when iterating thru the attributes of the root group.
This pr changes that so that they are suppressed when using
Group.getAttributes(). They can still be accessed by name using
Group.findAttribute(String name).
This change enforces the similar rule used in the netcdf-C library.

Added a test dataset and a test case for this change.


lesserwhirls:
👍

PR #953, opened 2017-11-14
AzimuthalEquidistant did not handle spherical Earth

msdsoftware:
Almost identical to the previous problem with handling the cylindrical equal area (#614), the azimuthal equidistant was experiencing an NPE if the grid mapping variable attributes described a spherical Earth. Solution, create a spherical Earth object if the call to makeCoordinateTransform did not construct an object.

Also set handling the spherical case of both these projections to watch for a specified Earth radius, and to pass that along when creating the Earth object. The previous update to handling the CEA did not do so.


lesserwhirls:
Thank you @msdsoftware! We plan to get a 4.6.11 release the week after Thanksgiving, and this fix will be in there.

msdsoftware:
@lesserwhirls, Good to know. I'm planning a Panoply +0.0.1 release before the end of the month and will try to wait for 4.6.11.

PR #958, opened 2017-11-22
Bump some dependencies

lesserwhirls:


lesserwhirls:
pending jenkins test results

build.gradle
```diff
@@ -56,6 +56,7 @@ buildscript {
         maven {
             url "https://dl.bintray.com/cwardgar/maven/"  // For 'com.cwardgar.gretty-fork:gretty'.
         }
+        mavenCentral() // for org.jsoup
```
cwardgar: I doubt very much that this line is necessary. The repositories in this file are only searched for Gradle plugins.

gradle/dependencies.gradle
```diff
@@ -187,9 +187,9 @@ libraries["jna"] = "net.java.dev.jna:jna:4.2.2"
 
 libraries["joda-time"] = "joda-time:joda-time:2.8.1"
 
-libraries["jsoup"] = "org.jsoup:jsoup:1.7.2"
+libraries["jsoup"] = "org.jsoup:jsoup:1.11.2"
 
-libraries["protobuf-java"] = "com.google.protobuf:protobuf-java:2.5.0"
+libraries["protobuf-java"] = "com.google.protobuf:protobuf-java:3.5.0"
```
cwardgar: This one will be interesting. I vaguely remember John considering an upgrade to 3.X but shying away because it wasn't a drop-in replacement.

build.gradle
cwardgar: To test, you can do: `./gradlew --refresh-dependencies assemble`.

lesserwhirls: Yep, didn't need it. Need to keep the `--refresh-dependencies` flag handy for the future! I suppose intellij does that automatically when it detects a change in dependencies?

lesserwhirls:
Ok, no new failures on Jenkins, and no failures related to `cdmr` or `cdmremote`.

lesserwhirls:
btw - I did the jenkins build under my 4.6.x [workspace](https://jenkins-aws.unidata.ucar.edu/job/sarms.4.6.x/7/)

cwardgar:
Good to see the protobuf changes didn't break anything. Btw, I was clearly thinking of this post from John on SO: https://stackoverflow.com/questions/33204321

So we're already on v3.X in 5.0.0. I think there **might** be breakage if we tried to generate new protobuf-handing code on master, so uh, let's not do that. I'm content as long as it works on 5.0.

PR #961, opened 2017-11-29
add dependency check suppression file

lesserwhirls:


lesserwhirls:
This PR suppresses false warnings on the tdm artifact (related to a php lib called tdm), as well as false warnings stemming from ncwms have trac urls in comments (thus dependency check thinks ncmws uses a vulnerable trac lib, which it does not).

PR #964, opened 2017-12-03
Don't include servlet-api JARs in WARs that we build

cwardgar:
* The culprit here was Gretty. I've updated to "com.cwardgar.gretty-fork:gretty:2.0.1", which fixes the bug.
* In :tds:war, fail the build if we find that a "servlet-api" JAR is slated for inclusion in the WAR.
* Use the "compileOnly" configuration that was introduced in Gradle 2.12, instead of "provided" or "providedCompile". As a result, we no longer need the "gradle-extra-configurations-plugin" plugin.
* Maven proxy repositories are proving to be a pain. Simply use the associated source repos instead. If those repos go away, we can host the needed artifacts in "unidata-3rdparty".

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -61,7 +61,6 @@ buildscript {
         classpath libraries["gretty"]
         classpath libraries["shadow"]
         classpath libraries["coveralls-gradle-plugin"]
-        classpath libraries["gradle-extra-configurations-plugin"]
     
         classpath libraries["guava"]  // For various utility methods used in the build scripts.
     }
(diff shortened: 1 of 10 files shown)


cwardgar:
So is there anything left to get in before the release?

lesserwhirls:
Nope, I don't think so. I will put a new build on thredds.ucar.edu to see how it looks there, and if everything looks ok, then I think we are good.

cwardgar:
Ok, I'll do the release today then.

yuanho:
Is it still version 4 release?

Yuan

cwardgar:
Yes, it'll be v4.6.11

yuanho:
I need to check in one small change for dorade radar, I will create a pull
request in a few minutes.

Yuan

yuanho:
Christian,

Can you make the change for me? only one line in the DoradeSweep.java, line
557, change from

if (findName(file, "SSWB") )

to

if (findName(file, "SSWB") || findName(file, "COMM"))


I have too many other modifications in my local repo.

Yuan

cwardgar:
Done: #965

yuanho:
Thanks.

Yuan

PR #965, opened 2017-12-04
Dorade radar tweak from Yuan

cwardgar:


cwardgar:
Merging my own PR, because it's minor and blocking the release.

PR #969, opened 2017-12-08
 Test on Travis's Ubuntu 14.04.5 environment using new libnetcdf 4.5.0 binaries

cwardgar:
* Other dependencies included are hdf5 1.10.1 and zlib 1.2.11.
* Nuked the old libnetcdf 4.4.1 binaries, which targeted Ubuntu 12.04.5.
* Removed a section of .travis.yml that sets the machine hostname because I *think* it's no longer necessary now that we're on Ubuntu 14. We'll see.

Remove "oraclejdk7" and "openjdk7" from Travis build matrix.  …
* On Ubuntu 14, we cannot build thredds with either of those JDKs. The reason is different for each one, and neither is Gradle's fault.

@Ignore TestNc4IospWriting.expandUnlimitedDimensions()  …
* The test is broken when we link against libnetcdf 4.5.0.
* Waiting on Unidata/netcdf-c#718 to be fixed.

Issue #895, opened 2017-07-19
dap4 test failing on travis


lesserwhirls:
See http://unidata-tds.s3.amazonaws.com/Travis/5.0.0-4483/4483.1/classes/dap4.test.TestServlet.html

~~~bash
dap4.core.util.DapException: CDMDSP: cannot process: /home/travis/build/Unidata/thredds/dap4/d4tests/src/test/data/resources/testfiles/test_atomic_array.nc
	at dap4.cdm.dsp.CDMDSP.open(CDMDSP.java:142)
	at dap4.cdm.dsp.CDMDSP.open(CDMDSP.java:35)
	at dap4.servlet.DapCache.open(DapCache.java:94)
	at dap4.servlet.DapController.doDMR(DapController.java:307)
	at dap4.servlet.DapController.handleRequest(DapController.java:230)
	at thredds.server.dap4.Dap4Controller.handleRequest(Dap4Controller.java:73)
	at sun.reflect.GeneratedMethodAccessor23.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.springframework.web.method.support.InvocableHandlerMethod.doInvoke(InvocableHandlerMethod.java:205)
	at org.springframework.web.method.support.InvocableHandlerMethod.invokeForRequest(InvocableHandlerMethod.java:133)
	at org.springframework.web.servlet.mvc.method.annotation.ServletInvocableHandlerMethod.invokeAndHandle(ServletInvocableHandlerMethod.java:97)
	at org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerAdapter.invokeHandlerMethod(RequestMappingHandlerAdapter.java:827)
	at org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerAdapter.handleInternal(RequestMappingHandlerAdapter.java:738)
	at org.springframework.web.servlet.mvc.method.AbstractHandlerMethodAdapter.handle(AbstractHandlerMethodAdapter.java:85)
	at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:963)
	at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:897)
	at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:970)
[...]

PR #970, opened 2017-12-08
Test on Travis's Ubuntu 14.04.5 environment using new libnetcdf 4.5.0 binaries (take 2)

cwardgar:
* Other dependencies include hdf5 1.10.1 and zlib 1.2.11.
* Nuked the old libnetcdf 4.4.1 binaries, which targeted Ubuntu 12.04.5.

Remove "oraclejdk7" and "openjdk7" from Travis build matrix.
* On Ubuntu 14, we cannot build thredds with either of those JDKs. The reason is different for each one, but neither is Gradle's fault. See https://github.com/DiUS/java-faker/pull/253

@Ignore TestNc4IospWriting.expandUnlimitedDimensions()
* The test is broken when we link against libnetcdf 4.5.0.
* Waiting on Unidata/netcdf-c#718 to be fixed.


cwardgar:
This PR is the same as #969, except that this time I build libnetcdf with `--disable-dap` and `--enable-logging`, as we talked about [here](https://github.com/Unidata/thredds/issues/895#issuecomment-317003803).

cwardgar:
The binaries were built using [this playbook](https://github.com/cwardgar/jenkins-slave-image/blob/master/ansible/roles/libnetcdf-and-deps/tasks/main.yml). Binaries for Jenkins will be generated the same way.

PR #976, opened 2017-12-14
Set the inflate output stream buffer size

DennisHeimbigner:
re: e-support ZIZ-818863
re: gihub pr https://github.com/Unidata/thredds/pull/955

The user has asked for the ability to use a larger output buffer
size for the stream to which the inflated file is written.  This
is a follow-on to the previous request to allow the user to set
the inflate buffer size.

The solution we have arrived at is to set the output buffer size
to be the same size as 8*chunk size being inflated (which is as
large or larger than the deflated size).  This means that the
chunk size for the input file will also determine the output
buffer size.

This fix is in master, if it works out, then I will add it to 5.0.0.
The fix only changes code in H5tiledLayoutBB.java, method
Datachunk#inflate().


cwardgar:
This'll get merged into 5.0.0 eventually, if you just want to leave it on master. I don't imagine it'll cause a conflict.

DennisHeimbigner:
I am not entriely sure that this fix will actually solve
the user's problem. We need to give them a new
jar for them to test.  There is no point in moving it
to 5.0.0 until we see if it is what they want.

PR #978, opened 2017-12-16
Fix FeatureDatasetCapabilitiesXML.doOne()

cwardgar:
* We must call FeatureDatasetPoint.calcBounds(), or else the capabilities document will lack the "TimeSpan" and "LatLonBox" elements.
* Do proper resource release with try/finally.
* Convert System.out.print calls to slf4j.
* User JUnit's TempoaryFolder @rule, not TestDir.getTempFile().

Other stuff:
* Update Gradle from 3.4.1 to 3.5.1.
* Add "tds/src/test/content/thredds/state" to .gitignore.

### DIFF
--- a/cdm-test/src/test/java/thredds/featurecollection/TestFeatureDatasetCapabilitiesXML.java
+++ b/cdm-test/src/test/java/thredds/featurecollection/TestFeatureDatasetCapabilitiesXML.java
@@ -38,12 +38,15 @@ import org.jdom2.JDOMException;
 import org.jdom2.output.Format;
 import org.jdom2.output.XMLOutputter;
 import org.junit.Assert;
+import org.junit.Rule;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
+import org.junit.rules.TemporaryFolder;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import ucar.nc2.constants.FeatureType;
-import ucar.nc2.ft.FeatureDataset;
 import ucar.nc2.ft.FeatureDatasetFactoryManager;
 import ucar.nc2.ft.FeatureDatasetPoint;
 import ucar.nc2.ft.point.writer.FeatureDatasetCapabilitiesWriter;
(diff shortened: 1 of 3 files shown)


cdm-test/src/test/java/thredds/featurecollection/TestFeatureDatasetCapabilitiesXML.java
```diff
@@ -65,11 +68,15 @@
  * Test FeatureDatasetCapabilitiesXML
  *
  * @author caron
- * @since 9/23/2015.
+ * @since 9/23/2015
  */
 @RunWith(Parameterized.class)
 @Category(NeedsCdmUnitTest.class)
 public class TestFeatureDatasetCapabilitiesXML {
+  private static final Logger logger = LoggerFactory.getLogger(TestFeatureDatasetCapabilitiesXML.class);
```
lesserwhirls: In Unidata/thredds#947, I started using the logger name `testLogger`, and configured `log4j` to print to the console at the `DEBUG` level. That basically preserves current behavior, but will allow us to tweak `log4j.xml` to control all of the test output. Not sure if that was a good idea or not, as this way will allow us to turn on individual test logs, which is nice. What do you think?

cwardgar: Best practice seems to be one logger per class, where the logger's name is the full name of the class. Primary benefits are better customizability and the ability to tell which class a log message came from, by including the logger's name in the message.

One change I would make is to declare `Logger`s like so:
```java
final static Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
as recommended by the [SLF4J FAQ](https://www.slf4j.org/faq.html#declaration_pattern). That one can be copy-and-pasted into different classes.

I plan to do a mini-overhaul of our logging system next week, with the goal of making the messages printed by our tests more useful. I'll include the new declaration style in that work.

cwardgar: A potential downside is that it won't be easy to enable logging for JUST test classes. For example, if I do:
```xml
    <logger name="thredds.featurecollection" level="DEBUG" additivity="false">
        <appender-ref ref="STDOUT" />
    </logger>
```
That'll cause DEBUG messages to be written for ALL classes in the "thredds.featurecollection" package, main and test. However, I think that's okay. I mean, if you're trying to debug `thredds.featurecollection.FeatureDatasetCapabilitiesXML`, you're probably going to want to see the messages from the main classes in the package as well.

lesserwhirls: 👍 Sounds good to me. Would you mind updating this PR to change what I did for the logger in Unidata/thredds#947?

cwardgar: I will do that in my logging mini-overhaul PR next week, if you don't mind.

lesserwhirls: Sounds good!

PR #979, opened 2017-12-17
Change Maven repository for geotoolkit JARs

cwardgar:
This seems to fix the threddsMaster failures we've been seeing on Jenkins, such as [Job 357](https://jenkins-aws.unidata.ucar.edu/job/threddsMaster/357/console). I noticed that all of the corrupt JARs were coming from http://maven.geotoolkit.org. That's not a valid Maven repository URL; the correct URL is https://nexus.geomatys.com/repository/geotoolkit/. Obviously they are doing some Apache magic to make the first URL work.

On a lark, I decided to try the second URL, and lo and behold, [the issue vanished](https://jenkins-aws.unidata.ucar.edu/job/cwardgar/1/console). I can't tell you why. Maybe the redirect/rewrite operation is corrupting the bytes, either on the client or the server. Who knows?

In any event, I'm not very confident that this commit really fixes the problem. Once this is merged into master, I want to run Jenkins again.

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -31,9 +31,9 @@ allprojects {  // Doesn't apply any plugins: safe to run closure on all projects
     
         // Third-party repositories.
         maven {
-            // For "geotk-referencing", which is needed by "ncwms". Repositories are not inherited from a dependency,
+            // For geotoolkit JARs, which are needed by "ncwms". Repositories are not inherited from a dependency,
             // so we must duplicate ncwms's needed repos here. See https://stackoverflow.com/a/19908009/3874643.
-            url "http://maven.geotoolkit.org/"
+            url "https://nexus.geomatys.com/repository/geotoolkit/"
         }
         maven {
             url "https://dl.bintray.com/cwardgar/maven/"  // For 'com.cwardgar.gretty-fork:gretty'.


lesserwhirls:
w00t - let's give it a shot!

cwardgar:
Yep, that did the trick: https://jenkins-aws.unidata.ucar.edu/job/threddsMaster/359/console

Jenkins is still marking the build as failed due to:
```
[threddsMaster] $ bash -xe /tmp/jenkins2857688722174231698.sh
+ . /etc/profile
+ '[' -e js-scan.txt ']'
+ retire --outputpath js-scan.txt
Build step 'Execute shell' marked build as failure
```
Is that expected? Or does it signal a problem with the AMI?

lesserwhirls:
By default, Jenkins runs the shell build step using:

~~~bash
bash -ex
~~~

The `-e` flag tells the shell to exit as soon as a non-zero status code is returned from a command, which is exactly what happens when retire finds vulnerable JavaScript libraries. However, we don’t want the whole build to fail because of that, so to fix it, we need to put:

~~~bash
#!/bin/sh
~~~~

at the top of the build script commands in the Jenkins config. This will cause the build step to fail only if the last command has a non-zero exit status. Retire will still return a non-zero status, which is why I put that lame echo command at the end, to ensure we don’t fail the whole build because of retire.

I’ve updated the config for master and started a new build, just to be sure. If that does fix things, I’ll update the config for 5.0.

lesserwhirls:
Uh... ¯\_(ツ)_/¯ 

~~~bash
FAILURE: Build failed with an exception.

* What went wrong:
Could not resolve all files for configuration ':tds:compileClasspath'.
> Could not download geotk-referencing.jar (org.geotoolkit:geotk-referencing:3.21)
   > Could not get resource 'https://nexus.geomatys.com/repository/geotoolkit/org/geotoolkit/geotk-referencing/3.21/geotk-referencing-3.21.jar'.
      > Tag mismatch!
~~~

Do you think we should just host the geotoolkit artifacts we use?

cwardgar:
Ah, I didn't realize you were depending on the shebang. I accidentally nuked that when I got rid of the `apt-get install` steps. My bad. I've set the shebang to `#!/usr/bin/env bash` for those configs.

As for the build failure, WHAT THE FUCK? My best guess now is that this is an SSL issue. In particular, that host only works on clients with SNI support (see https://www.ssllabs.com/ssltest/analyze.html?d=nexus.geomatys.com&hideResults=on), and Gradle may not have it (see https://github.com/gradle/gradle/issues/1868). That doesn't explain why the failure is intermittent though.

In any event, creating a proxy repository for it might solve the problem. I'll try that tomorrow.

PR #981, opened 2017-12-19
Use JUnit's TemporaryFolder Rule for temporary files

cwardgar:
* `TemporaryFolder` automatically cleans up after itself, instead of leaving temp files in the build directory.
* Nuked `TestDir.getTempFile()` and `TestDir.temporaryLocalDataDir`.
* Don't bother naming any of the files we create with `TemporaryFolder`. Doing so could cause an exception if we'd already created a file with the same name.
* Remove all use of `TestFileDirUtils.createTempDirectory()` and `TestFileDirUtils.deleteDirectoryAndContent()`. That functionality is provided by `TemporaryFolder`.


cwardgar:
This is a big commit that involved a lot of copy/paste. I bet I broke something. Let me run Jenkins before merging.

cwardgar:
Yep, broke a ton of shit, but almost all of it was due to poorly-written tests. Fixed now. I also discovered and fixed two bugs:

* `NetcdfDataset` can't open local NcML files that lacks a ".xml" or ".ncml" extension.
* `XMLStore.createFromFile()` treats an empty file as an existing storage, instead of a brand new one. It tries to parse the empty file, which results in an exception.

This is ready to be merged.

Issue #646, opened 2016-09-23
Switch all logging to logback


cwardgar:
Currently we're using log4j2 in TDS and JUL elsewhere ("slf4j-jdk14"). Logback is superior to these in [just about every way](http://stackoverflow.com/questions/11359187/why-not-use-java-util-logging/11360517#11360517).

Switching from JUL to logback will be simple, not least because we're not using any `logging.properties` config files anywhere, which are specific to JUL. Transitioning TDS to logback will be trickier. We actually have some logging code that specifically depends on log4j2.

JohnLCaron:
Im not sure you should take Ceki's word for it. theres a reason why he has forked twice.=, and only does one person projects.

Having said that, his libraries have been quite good. And hes dedicated his whole life to it (it seems). Just saying maybe see who else is using what.

SLF4J (which is his) of course makes the choice painless, since its mostly a runtime decision. which is good.

Google has developed its own library called flogger, which has very elegant fluid API, and is apparently very performant. Dont think its available outside of google but ill check. It is used in the code to form the log messages, ie not a facade. And it uses standard JUL logging for the actual output. So very different tradeoff.

syntax is like:  logger.asWarning().withException(e).log("what %d up %s%n", i, s);

the asWarning() prevents the rest being evaluated if warning is not on, so very efficient.

cwardgar:
For me, the main benefits of logback vs log4j2 are simpler configuration, saner default behavior when no config is provided, and better documentation. It seems to be the popular choice for newer Java projects as well, such as Gradle.

What issues do we need to be aware of for transitioning TDS from log4j to logback? I know a couple:
* TdsMonitor. Would we be okay here as long as we generated log messages in logback with the exact same pattern as those we generated in log4j2?
* [LoggerFactorySpecial](https://github.com/Unidata/thredds/blob/5.0.0/tdcommon/src/main/java/thredds/util/LoggerFactorySpecial.java). Generate `Logger`s with a custom `RollingFileAppender`. Seems like a hack: you left the comment `all wrong see http://logging.apache.org/log4j/2.x/manual/extending.html`. Perhaps there's an easier way to extend in logback (or even log4j2, for that matter).

Anything else you can think of?

JohnLCaron:
AFAIU, as long as we use slf4j to create the log messages, you can just
relink any app, including tdm, with logback.

I just looked and it appears tds 5.0 no longer uses loggerFactory. So i
think we can rip it out and then just relink.

[...]

PR #983, opened 2017-12-21
Add a logger to every test class and replace existing test loggers

cwardgar:
I now have Carpal Tunnel Syndrome. Ow. It was nice to not have to use my brain for a while though.

Next I'll be switching the logging backend from log4j2 to logback (see #646). Then I'll provide a default `logback-test.xml` config for each subproject.

I will NOT go through and convert all `System.out.print*` calls to SLF4J for each test class. That sounds like a Herculean (and mind-numbing) task that could only be partially automated. However, I will do the conversion for individual test classes that I happen to work on in the future.


lesserwhirls:
That looks like it was...fun...

lesserwhirls:
I'm going to override the code owners check on this one since it's just test logger stuff.

PR #984, opened 2017-12-21
Update threddsConfig.xml

lesserwhirls:
Change `<GribIndexing>` element to `<GribIndex>` and update the default options.


lesserwhirls:
Already updated link in live docs.

Issue #896, opened 2017-07-20
Bug for 360_day calendar and time period edges between 29 and 30th February


glevava:
Hi, 
I encountered an issue regarding the THREDDS calendar support.
Even THREDDS is able to support 360_day calendar, it raises an `IOexception for Invalid Catalog` if the start or end coverage date time is between the 29th and 30th February.

The XML catalog is properly generate but java raises the following error when it try to read it:
`java.lang.IllegalArgumentException: Illegal base time specification: '2066-02-30' Value 30 for dayOfMonth must be in the range [1,28]`

If I manually change the date in the XML catalog to `2066-03-01` for instance, it works as expected.

cofinoa:
@glevava, could you attach the .XML catalog file with the problematic date?

glevava:
Hi Antonio, attached is (one of) the "affected" catalog (in txt format).

[invalid_catalog.txt](https://github.com/Unidata/thredds/files/1163637/invalid_catalog.txt)

cofinoa:
@glevava 
Currently the XML elements in the `<timeCoverage>` doesn't support other calendars/dates/periods than the Gregorian one: 
```xml
<timeCoverage>
  <start>1980-12-01T00:00:00</start>
  <end>2066-02-30T00:00:00</end>
</timeCoverage>
...
```
A modification on the catalog builder should allow to an attribute like this: 
```xml
<timeCoverage calendar="360_day">
  <start>1980-12-01T00:00:00</start>
  <end>2066-02-30T00:00:00</end>
</timeCoverage>
```

then any date/period element refers to the calendar been specified in the attribute.

cofinoa:
It may be also the `<addTimeCoverage>` for the server-side catalog specification should be also modified. 

Any other thoughts?

glevava:
I manually modified the catalog to get:
```
<timeCoverage calendar="360_day">
  <start>1980-12-01T00:00:00</start>
  <end>2066-02-30T00:00:00</end>
</timeCoverage>
```
And try publication again. It fails with the same error.

PR #985, opened 2017-12-22
Resolves Unidata/thredds#896

cofinoa:
The thredds catalog element timeCoverage can accept a calendar attribute.

### DIFF
--- a/cdm/src/main/java/thredds/client/catalog/builder/CatalogBuilder.java
+++ b/cdm/src/main/java/thredds/client/catalog/builder/CatalogBuilder.java
@@ -654,7 +654,7 @@ public class CatalogBuilder {
     // look for dates
     list = parent.getChildren("date", Catalog.defNS);
     for (Element e : list) {
-      DatasetBuilder.addToList(flds, Dataset.Dates, readDate(e));
+      DatasetBuilder.addToList(flds, Dataset.Dates, readDate(e,null));
     }
 
     // look for documentation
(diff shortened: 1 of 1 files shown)


cwardgar:
@cofinoa Thanks for the contribution! I think we should add a test for this before merging it in.

lesserwhirls:
We will also need to update the xsd ([InvCatalog.1.2.xsd](https://github.com/Unidata/thredds/blob/b731bcb45b6e10b7e6102e97a9ef35e9fef43c93/cdm/src/main/resources/resources/thredds/schemas/InvCatalog.1.2.xsd)). Since 5.0 isn't released yet, I think it is ok to keep the version as-is.

cofinoa:
@cwardgar do you want to me to add the tests? I'm not experienced on that, but if you provide to me some "clues" I can try. what is the next step?

@lesserwhirls do you think we can integrate this onto InvCatalog.1.0.7.xsd too? and what about TDS4.6?

cwardgar:
Well, it turns out that we're not testing `timeCoverageType` at all currently, so adding something for your new `calendar` attribute wouldn't be as easy as I hoped. We really need to overhaul our catalog testing, with a goal of 100% coverage of the various capabilities.

But I won't make you wait until we've done that; your PR looks good to go in now. It's pretty simple after all.

@lesserwhirls So what is the difference between `InvCatalog.1.0.7.xsd` and `InvCatalog.1.2.xsd`? Is 1.0.7 for TDS 4.6 and 1.2 for TDS 5.0? Do you think we should add the `calendar` attribute to both?

cwardgar:
This PR is superseded by #1002. Thanks for getting this feature rolling, @cofinoa.

PR #987, opened 2017-12-23
Use "logback-classic" as the logging backend for our tests

cwardgar:
Clean up log4j cruft.
* tdcommon no longer implements a custom log4j component, so we can eliminate its "log4j-api" and "log4j-core" deps. "log4j-api" isn't used anywhere else in the project, so we can nuke it.
* "log4j-core" needn't be explicitly included because it will be dragged in transitively by "log4j-slf4j-impl". So we can nuke it too.
* Now that TDS includes "log4j-web", we no longer need to do explicit startup/shutdown of log4j in TdsContext.
* Update to the latest versions of SLF4J and Log4J.

Use "logback-classic" as the logging backend for our tests.
* Previously, each subproject used either "slf4j-jdk14" or "log4j-slf4j-impl". Now it's all "logback-classic".
* Added a basic "logback-test.xml" to all test SourceSets. It includes instructions for how it should be used during development/debugging.
* Nuked "slf4j-jdk14" from the project as we're no longer using it anywhere. Nuked "spring-security-taglibs" for the same reason.
* Removed old log4j config files that were used for testing.
* dap4lib and toolsUI also have been moved to logback-classic. They don't have dedicated logback.xml files, so they will get the default logback config, which is actually quite sane and useful (unlike in log4j).


lesserwhirls:
I'll go ahead and merge this without all of the reviews since this is logging config related stuff.

PR #989, opened 2017-12-24
Run ClientCatalogBasicSpec against local embedded thredds server

cwardgar:
* This required moving the class out of `:cdm` and into `:it`. Also, some utility code needed to be moved out of `:cdm` and into `:testUtil`.
* Previously, the tests were running against `TestDir.threddsTestServer` and would sometimes fail.
* Updated the tests in `thredds.client.catalog` to use the new `ClientCatalogUtil` in `:testUtil`.
* Added `logback-test.xml` to a few test SourceSets that I missed.
* Converted from `System.out` to slf4j.


cwardgar:
As you can see, I'm a bit backed up with PRs.

lesserwhirls:
Waiting in line at the store...worst idea ever...might as well merge some PRs!

PR #990, opened 2017-12-24
Change logger declarations in Groovy test classes.

cwardgar:
* The `MethodHandles.lookup().lookupClass()` trick doesn't work in Groovy, unfortunately.


cwardgar:
Okay, this is my last PR for a while. I'll run Jenkins once it's merged in. I think we're down to like 5 or 6 outstanding test failures!

PR #991, opened 2017-12-28
Eliminate dependence on thredds-test from our test suite

cwardgar:
External dependencies are bad, unless it's something rock-solid like Maven Central. thredds-test does not fit that bill, so I've eliminated its use in our test suite. This ought to eliminate several periodic failures.

The process mostly consisted of replacing references to datasets on thredds-test with references to comparable datasets on the embedded TDS that `:it` stands up. Also, since the embedded TDS is only available when `:it:test` runs, most of the relevant tests had to be moved out of their former subprojects and into `:it`.

In other cases, there were tests making use of `TestDir.threddsTestServer` that I deemed too terrible to save. I've gone ahead and nuked them.

To review this PR, it's probably easier to just go through the individual commits, one-by-one.


cdm-test/src/test/java/ucar/nc2/dt/grid/TestGridSubset.java
```diff
@@ -202,50 +199,10 @@ public void testMSG() throws Exception {
     }
   }
 
-  @Test
-  @Ignore("Bad URL, as of 2015/03/11.")
```
lesserwhirls: :-/

cdm/src/test/java/thredds/client/catalog/TestResolve.java
```diff
@@ -33,22 +33,15 @@
...
 import java.util.List;
 
-/** Test reletive URL resolution. */
-
-public class TestResolve1 {
+/** Test relative URL resolution. */
+public class TestResolve {
   private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
   String base="http://www.unidata.ucar.edu/";
```
lesserwhirls: I suppose the tests that use `base` are a good check to see if we can handle a redirect. Would it be better to change `base` to use https, and only do one test with the non-https url, or maybe parameterize this test to run it with both http and https? I'm probably overcomplicating things here.

it/src/test/java/thredds/client/catalog/TestResolve.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 1998-2009 University Corporation for Atmospheric Research/Unidata
+ * Copyright 1998-2015 University Corporation for Atmospheric Research/Unidata
```
lesserwhirls: 2017?

lesserwhirls: and remove rest of copyright block?

cdm-test/src/test/java/ucar/nc2/dt/grid/TestGridSubset.java
cwardgar: Are you reacting to the bad URL, or that the test was removed?

FYI, the test is not removed; merely moved to `TestGridSubsetThredds`.

cdm/src/test/java/thredds/client/catalog/TestResolve.java
cwardgar: As far as I can tell, we never connect to the URL; it's used only for string comparison.

it/src/test/java/thredds/client/catalog/TestResolve.java
cwardgar: Yeah, I've made no effort whatsoever to update copyright blocks. The only reason this changed is because of a copy/paste, I assume. I think those blocks should be handled by automation, not us.

cdm-test/src/test/java/ucar/nc2/dt/grid/TestGridSubset.java
lesserwhirls: The bad url comment :-)

it/src/test/java/thredds/client/catalog/TestResolve.java
lesserwhirls: Ideally automation, but that's not in place yet. No need to update now - we'll address that later.

cdm/src/test/java/thredds/client/catalog/TestResolve.java
lesserwhirls: I though I saw a connection in there, but nope, just good ol' string comparisons. My bad.

lesserwhirls:
👍

PR #992, opened 2017-12-28
Handle embedded TDS URLs in one place

cwardgar:
* Instead of having a bunch of strings like "localhost:8081/thredds/" strewn all over the place, we now define and manage that URL in one place: `TestOnLocalServer`. As a result, we can easily move the embedded TDS to a new port or context path in the future, if necessary.
* Renamed `TestWithLocalServer` to `TestOnLocalServer`. I did this simply because a method call like `TestOnLocalServer.withHttpPath()` sounds better when read aloud than `TestWithLocalServer.withHttpPath()`. I admit that this is a totally non-functional change resulting from OCD.
* For clarity, `TestOnLocalServer.withPath()` was renamed to `withHttpPath()`, because we now also have `withDodsPath()` and `withProtocolAndPath()`.


cwardgar:
This had a Travis failure at first, but now it's ready to go in.

PR #993, opened 2017-12-30
Remove cruft from TestDir

cwardgar:
* `TestDir` is a crucial utility for our test suite; almost every class in `:cdm-test` uses it in some way. So, I figured that cleaning it up could be a significant win.
* Removed "thredds.properties" and everything related to it from `TestDir`. Apparently it was an alternate way to specify "unidata.testdata.path", but I don't think we're making use of it at all.
* `Misc` also had a method that processed thredds.properties called `getTestdataDirPath()`. However, that method was only used by features in the BUFR section of ToolsUI, and even when a valid thredds.properties was available, those features no longer worked. So, I nuked `Misc.getTestdataDirPath()`, as well as the relevant ToolsUI buttons.
* I identified that `getAllFilesInDirectoryStandardFilter()`, `getAllFilesInDirectory()`, and `showMem()` were each only used by one class. So, I moved those methods to the appropriate classes.
* Completely refactored `TestOffAggFmrcGrib`. It was pretty damn crufty before.
* Removed `TestAggMisc.readAllData()` and `TestAggMisc.makeSubset()` as we already have those methods in `TestDir`.
* Log to SLF4J, not STDOUT.

### DIFF
--- a/cdm-test/src/test/java/ucar/nc2/dataset/TestConventionFeatureTypes.java
+++ b/cdm-test/src/test/java/ucar/nc2/dataset/TestConventionFeatureTypes.java
@@ -19,6 +19,7 @@ import java.io.File;
 import java.io.IOException;
 import java.lang.invoke.MethodHandles;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Formatter;
 import java.util.List;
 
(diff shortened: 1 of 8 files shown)


lesserwhirls:
What are the changes in the ui BURF stuff do? Were those methods not used anywhere?

lesserwhirls:
Doh, just reread the PR.

PR #994, opened 2017-12-31
In tests, log all NCdumpW output to SLF4J

cwardgar:
* Previously, most of it was logged to STDOUT. It was a huge contributor to the STDOUT spam we get when running tests.
* Most of the changes involved switching from the `NCdumpW.print*` methods to the `NCdumpW.toString()` methods. The latter produces a string that we just pass on to SLF4J.
* Nuked some methods in `NCdumpW` that were no longer being used. One of the `printArray()` methods was only being used by a (useless) `main()` method in `WRFConvention`, so I went ahead and nuked the `main()` too.
* I only converted `NCdumpW` output to SLF4J in this commit; many of the edited classes still have loads of other crap being printed to STDOUT. We'll get to that stuff in time.

### DIFF
--- a/cdm-test/src/test/java/ucar/nc2/dataset/TestCoordinates.java
+++ b/cdm-test/src/test/java/ucar/nc2/dataset/TestCoordinates.java
@@ -35,7 +35,7 @@
 package ucar.nc2.dataset;
 
 
-import junit.framework.*;
+import junit.framework.TestCase;
 import org.junit.experimental.categories.Category;
 import org.slf4j.Logger;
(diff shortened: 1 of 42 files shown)


cwardgar:
This will probably be my last PR that affects a large number of test classes. Next, I'll focus on our remaining test failures. I think I can probably fix all but the DAP4 ones.

lesserwhirls:
Dude, you're a test cleanin' machine!

lesserwhirls:
Not sure what's up with CLAHub, but I'll turn it off and merge.

PR #997, opened 2018-01-03
Bug fix for creating logical view of a constant Array

cwardgar:
* Previously, such Arrays would produce a `ClassCastException` when you called `createView()` on them. Public methods like `reduce()` and `flip()` rely on `createView()`.
* The fix required modifying a `factory()` method in most of the `Array` subtypes.
* Added a regression test to `ArrayTest`.

Other:
* Converted STDOUT to SLF4J in `GribCordsMatchGbx` and `TestCoordinatesMatchGbx`. I was originally trying to fix a test failure related to those classes but decided to change course.
* ToolsUI (main) now uses `logback.xml` for logging configuration. It replaced `jdk1.4-logging.properties`.


cwardgar:
The failure in `TestLogicalReduce.testStationProfile()` suffered from this. Fixing the bug only revealed a deeper problem, unfortunately. I have a separate commit that addresses it.

PR #1002, opened 2018-01-09
Add calendar attribute to timeCoverage in client catalog

cwardgar:
* Updated InvCatalog.1.2.xsd.
* Added test to `TestClientCatalog.testTimeCoverage()` that asserts that the element is being processed and that it is having the expected effect on start and end times.
* Removed duplicate method from `TestClientCatalog` and converted its STDOUT prints to SLF4J.
* Added documentation to `InvCatalogSpec.adoc`.

This implements the additions we discussed in #985. I decided to change `calendar` from an attribute to an element because it is more in line with how the other `timeCoverage` data are specified.

@lesserwhirls This PR includes documentation changes that you may want to keep track of.

### DIFF
--- a/cdm/src/main/java/thredds/client/catalog/builder/CatalogBuilder.java
+++ b/cdm/src/main/java/thredds/client/catalog/builder/CatalogBuilder.java
@@ -651,7 +651,7 @@ public class CatalogBuilder {
     // look for dates
     list = parent.getChildren("date", Catalog.defNS);
     for (Element e : list) {
-      DatasetBuilder.addToList(flds, Dataset.Dates, readDate(e,null));
+      DatasetBuilder.addToList(flds, Dataset.Dates, readDate(e, null));
     }
 
     // look for documentation
@@ -981,14 +981,13 @@ public class CatalogBuilder {
   /*
     <xsd:complexType name="timeCoverageType">
     <xsd:sequence>
+      <xsd:attribute name="calendar" type="xsd:string"/>
       <xsd:choice minOccurs="2" maxOccurs="3">
         <xsd:element name="start" type="dateTypeFormatted"/>
         <xsd:element name="end" type="dateTypeFormatted"/>
         <xsd:element name="duration" type="duration"/>
       </xsd:choice>
-
       <xsd:element name="resolution" type="duration" minOccurs="0"/>
-      <xsd:element name="calendar" type="xsd:string" minOccurs="0"/>
     </xsd:sequence>
   </xsd:complexType>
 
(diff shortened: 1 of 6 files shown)


cwardgar:
I'm inclined to limit this addition to `InvCatalog.1.2.xsd` only. TDS v5 is not yet released, so we can do anything we want to `InvCatalog.1.2.xsd`. On the other hand, if we added it to `InvCatalog.1.0.X.xsd`, we'd have to create a new version of the schema. Bleh.

At this point, I think the only changes we should be making to 4.6 should be bug fixes, and serious bugs at that. This doesn't qualify.

cdm/src/main/java/thredds/client/catalog/builder/CatalogBuilder.java
```diff
@@ -991,6 +988,7 @@ private Element readMetadataFromUrl(java.net.URI uri) throws java.io.IOException
       </xsd:choice>
 
       <xsd:element name="resolution" type="duration" minOccurs="0"/>
+      <xsd:element name="calendar" type="xsd:string" minOccurs="0"/>
```
cofinoa: I thougth calendar as an attribute of `<timeCoverage>` instead of an element like the others (start, end, duration and resolution). The reason in my mind is that calendar value provides the semantic for the time/periods represented in the inner elements of timeCoverage.

cdm/src/test/data/thredds/catalog/TestTimeCoverage.xml
```diff
@@ -1,57 +1,55 @@
...
   </dataset>
-
-
-</catalog>
\ No newline at end of file
+  
+  <dataset name="test5" ID="test5">
+    <metadata inherited="true">
+      <timeCoverage>
+        <start>2017-02-12</start>
```
cofinoa: To test also the date representation I would put an start date represented by 2017-02-30 and the end 30 day later in the 30 days uniform calendar, i.e. end 2017-03-30. The 2017-02-30 should fire an exception in the netcdf-java if standard calendar is been used.

cdm/src/test/java/thredds/client/catalog/TestClientCatalog.java
```diff
@@ -190,6 +192,21 @@ public void testTC() throws Exception {
...
     TimeUnit tu = d.getTimeUnit();
     assert tu.getUnitString().equals("days") : tu.getUnitString(); // LOOK should be 3 hours, or hours or ??
+
+    ds = cat.findDatasetByID("test5");
+    tc = ds.getTimeCoverage();
+    assert null != tc;
+    logger.debug("tc = {}", tc);
+
+    CalendarDate start = tc.getStart().getCalendarDate();
+    assert CalendarDateFormatter.toDateString(start).equals("2017-02-12");
```
cofinoa: see comment above

docs/website/tds/catalog/InvCatalogSpec.adoc
```diff
@@ -880,6 +885,12 @@ Example:
   <duration>10 days</duration>
   <resolution>15 minutes</resolution>
 </timeCoverage>
+
+<timeCoverage>
+  <start>2017-02-12</start>
```
cofinoa: see comment above

cwardgar:
@cofinoa I agree with both of your suggestions. See my changes in the new commit. And thanks for the review.

lesserwhirls:
Looks good to me. We're good to go here?

cwardgar:
Yep, good to go.

PR #1010, opened 2018-01-19
Enable NcML written in TDS config catalogs to be opened as a FeatureDatasetCoverage

lesserwhirls:
Fixes the issue where WCS service did not work for NcML, as reported on the THREDDS mailing list.

### DIFF
--- a/cdm/src/main/java/ucar/nc2/ft2/coverage/CoverageDatasetFactory.java
+++ b/cdm/src/main/java/ucar/nc2/ft2/coverage/CoverageDatasetFactory.java
@@ -1,16 +1,20 @@
 /*
- * (c) 1998-2017 John Caron and University Corporation for Atmospheric Research/Unidata
+ * (c) 1998-2018 John Caron and University Corporation for Atmospheric Research/Unidata
  */
 package ucar.nc2.ft2.coverage;
 
 import ucar.nc2.constants.FeatureType;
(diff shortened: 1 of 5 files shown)


cwardgar:
Can you add a test for this?

lesserwhirls:
I will. I found an issue with our ChronicleMap dataset tracker that might cause some issues, so let me fix that one first.

lesserwhirls:
So this PR really needs Unidata/thredds/pull/1011 first, otherwise you have to remove the dataset tracker database every time you restart the TDS. Once that goes in, I will write a test and update this PR.

lesserwhirls:
Ok, this should be ready to review as long as travis passes.

cwardgar:
So you didn't need to add a `serviceName` for WCS to the catalog?

lesserwhirls:
I was able to switch [this](https://github.com/Unidata/thredds/pull/1010/files#diff-ffa9afb62a6f8f764317c4f0f7ff8225R172) one to use `serviceName=all`, which includes wcs. However, that agg was sort of broken before, so I also had to fix the `DiskCache` path in [threddsConfig.xml](https://github.com/Unidata/thredds/pull/1010/files#diff-e765faf390d844eea8fec203b3619f89L90).

cwardgar:
Ah, okay. Make sense.

PR #1007, opened 2018-01-18
Cleanup

lesserwhirls:
Update a few dependencies and remove old godiva2 files. Will need to run on Jenkins just to be sure about the chronicle-map upgrade.

PR #1011, opened 2018-01-20
Fix bug in config catalog initialization

lesserwhirls:
In the case where a dataset tracker database exists on disk and the server uses read mode = always on startup, there was a bug in that the dataset tracker database was not properly being removed because it was not closed first. This resulted in an empty dataset tracker after startup, which means the TDS was not aware that it had any datasets. For some services, this isn't a big deal, as they do not use the dataset tracker (for example, opendap), but for others, this was a major problem (wcs, wms, for example). This issue showed up with the [update to ChronicleMap](https://github.com/Unidata/thredds/pull/1007).

### DIFF
--- a/tds/src/main/java/thredds/core/ConfigCatalogInitialization.java
+++ b/tds/src/main/java/thredds/core/ConfigCatalogInitialization.java
@@ -181,7 +181,17 @@ public class ConfigCatalogInitialization {
 
     switch (readMode) {
       case always:
-        if (databaseAlreadyExists) this.datasetTracker.reinit();
+        // if the database already exists, we need to close it
+        // before we reinit
+        if (databaseAlreadyExists) {
+          logCatalogInit.info("ConfigCatalogInitializion datasetTracker database already exists - closing it before reinitialization.");
+          try {
+            this.datasetTracker.close();
+          } catch (IOException e) {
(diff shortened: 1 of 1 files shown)


tds/src/main/java/thredds/core/ConfigCatalogInitialization.java
```diff
@@ -181,7 +181,16 @@ public synchronized boolean reread(ReadMode readMode, boolean isStartup) {
...
     switch (readMode) {
       case always:
-        if (databaseAlreadyExists) this.datasetTracker.reinit();
+        // if the database already exists, we need to close it
+        // before we reinit
+        if (databaseAlreadyExists) {
+          try {
+            this.datasetTracker.close();
+          } catch (IOException e) {
+            e.printStackTrace();
```
cwardgar: It's better to log this to `logCatalogInit` than dump the stack trace to STDERR. I think SLF4J provides a `error()` method that takes a `Throwable` as an argument.

Also, you might want to log an INFO message about  closing `datasetTracker` because the database  already exists.  Might be useful for debugging later.

lesserwhirls: Good call - thanks!

lesserwhirls:
Good call. Ok, should be good to go now.

Issue #834, opened 2017-05-10
reading grib returns error java.lang.NullPointerException


jbadag:
Hi, 
I'm working with subseasonal data of ECMWF. I download the data of the TIGGE page for the NCEP model, for real time data:
(http://apps.ecmwf.int/datasets/data/s2s-realtime-instantaneous-accum-kwbc/levtype=sfc/type=cf/)

and reforecasts data:
(http://apps.ecmwf.int/datasets/data/s2s-reforecasts-instantaneous-accum-kwbc/levtype=sfc/type=cf/). 

I open and process this gribs data files using matlab 2017 with nctoolbox 1.1.3 and i don't have any problem with real time data, but when I try to open the reforecasts data gribs, I get a java error (see below).

On the other hand I tried to view the file using the panoply software and got the same error message (see attached). 
![unnamed](https://cloud.githubusercontent.com/assets/16575671/25910092/96b8fc1a-3585-11e7-83aa-e664d2c3f01e.png)

An example file with which I have problems can be downloaded from here:
https://www.dropbox.com/s/j4mtsk0jsj2jiz2/20101229_31.grib?dl=0

My Java version is:
Java 1.7.0_60-b19 with Oracle Corporation Java HotSpot(TM) 64-Bit Server VM mixed mode

My Operative System is: 
Microsoft WIndows 10 Pro

Does anyone have an idea what could be causing the problem?

Error using ncdataset (line 97)
Failed to open C:

Error in cfdataset (line 59)
obj = obj@ncdataset(url);

Error in ncgeodataset (line 74)
obj = obj@cfdataset(url);

Error in bias_NCEP (line 11)
nc=ncgeodataset(archivo_con);

Caused by:
Error using ncdataset (line 81)
Java exception occurred:
java.lang.NullPointerException

	at ucar.nc2.grib.grib2.Grib2Rectilyser.cdmVariableHash(Grib2Rectilyser.java:453)

	at ucar.nc2.grib.grib2.Grib2Rectilyser.make(Grib2Rectilyser.java:101)

	at ucar.nc2.grib.grib2.Grib2CollectionBuilder.makeAggregatedGroups(Grib2CollectionBuilder.java:489)

	at ucar.nc2.grib.grib2.Grib2CollectionBuilder.createIndex(Grib2CollectionBuilder.java:414)

	at ucar.nc2.grib.grib2.Grib2CollectionBuilder.readOrCreateIndex(Grib2CollectionBuilder.java:157)

[...]

PR #1012, opened 2018-01-23
Fixes issue #834 :  Support GRIB2 PDS Template 61

cofinoa:
The PDS template 11 has been used as base to support template 61. 

A public method for for getting model version date has been added (i.e. `getModelVersionDate` ).


lesserwhirls:
Thanks so much for this contribution @cofinoa! I'm going to add a test for this, but otherwise, it looks good.

lesserwhirls:
I added a test for this on https://github.com/Unidata/thredds/pull/1013 - closing this one in favor of that.

PR #1015, opened 2018-01-25
Allow to to specify in the filter element to match the full path

cofinoa:
New path attribute on filters to allow matching the full path instead only the name. See:
https://stackoverflow.com/questions/48445792/filtering-thredds-datasetscan-by-both-directory-and-file-pattern

Example of directory structure:

```
/GGG/alpha/a
          /b
          /c
/GGG/beta/a
         /b
         /c
```

the following `<datasetScan>`

```xml
<datasetScan name="Test Filtering" ID="testFiltering"
               path="testFiltering" location="/GGG">
    <filter>
      <include wildcard="*" atomic="true" collection="true"/>
      <exclude wildcard="*alpha/a" path="true" collection="true"/>
    </filter>
  </datasetScan>
```

will show all the directories except the  `alpha/a` directory

### DIFF
--- a/cdm/src/main/java/thredds/catalog/parser/jdom/InvCatalogFactory10.java
+++ b/cdm/src/main/java/thredds/catalog/parser/jdom/InvCatalogFactory10.java
@@ -661,6 +661,15 @@ public class InvCatalogFactory10 implements InvCatalogConvertIF, MetadataConvert
               collection = true;
           }
 
+          // Determine if applies to datasets path, default false.
+          boolean isPath = false;
+          String pathAttVal = curElem.getAttributeValue("path");
+          if (pathAttVal != null) {
+            // If not "false", set to true.
+            if (!pathAttVal.equalsIgnoreCase("false"))
+            	isPath = true;
+          }
+
           // Determine if include or exclude selectors.
           boolean includer = true;
           if (curElem.getName().equals("exclude")) {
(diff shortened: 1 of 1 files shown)


cwardgar:
This is definitely a worthwhile addition, but I hesitate to merge it into master, because catalog handling is vastly different in 5.0.0. I think we should target 5.0.0 only.

We'll need to:
* Add "path" attribute parsing to `ConfigCatalogBuilder.readDatasetScan()`.
* Add "path" attribute to `InvCatalog.1.2.xsd`.
* Add unit test(s) to `TestDatasetScanFilter`.
* Update documentation.

I'll try to get to it before the release of 5.0.0. I'll leave this PR open and un-merged until then. Thanks for getting this started, @cofinoa.

cofinoa:
Agree.

I was looking at the TDS4.6 source code to understand the `datasetScan` and it was too obvious to me how to implement that feature (which has been a long standing complain "among TDS admins").

cofinoa:
@cwardgar I have migrate the master changes to the version 5.0.0 branch
https://github.com/cofinoa/thredds/tree/support_path_filter_v5

but I don't know how to add the unit tests to `TestDatasetScanFilter`, basically because of my lack of how to work with them. 

If you provide me some basic instrucitions or guidance I can learn for future PR.

cwardgar:
I'm converting this to an issue (#1032), just to clear out our open pull requests a bit.

PR #1021, opened 2018-02-01
Downgrade jna shipped with chronicle-map

lesserwhirls:
Allows the TDS to be run on systems using GLIBC 2.12. We should be able to get rid of this on the next release of chronicle-map, as their snapshot BOM shows that they pulled back on jna.


cwardgar:
Commit looks good. This just needs a rebase.

PR #1022, opened 2018-02-02
setEarthlocation api change

yuanho:
allow location variable to be an array

### DIFF
--- a/cdm-test/src/test/java/ucar/nc2/dataset/TestCFRadial.java
+++ b/cdm-test/src/test/java/ucar/nc2/dataset/TestCFRadial.java
@@ -43,6 +43,14 @@ public class TestCFRadial {
                 FeatureType.RADIAL, filename, null, buf);
     }
 
+    private RadialDatasetSweep arrayLatLonData() throws IOException
+    {
+        String filename = TestDir.cdmUnitTestDir + "conventions/cfradial/cfrad.20171127_202111.203_to_20171127_202123.085_DOW7_v275_s04_el7.00_SUR.nc";
+        Formatter buf = new Formatter();
+        return (RadialDatasetSweep) FeatureDatasetFactoryManager.open(
+                FeatureType.RADIAL, filename, null, buf);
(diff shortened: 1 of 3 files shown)


lesserwhirls:
Hi @yuanho - do you have a test file that used to cause this to fail?

yuanho:
Yes, where do you want me to upload it?


Yuan

lesserwhirls:
@yuanho - you can probably attach it to this issue if you zip the file.

yuanho:
too big to attach.

Yuan

lesserwhirls:
Dang. Ok, how about RAMADDA or Google Drive?

lesserwhirls:
Ok, I got the file and it's part of our test data now. I am writing some tests for this, and I have a question. The `.isStationary()` method of `CFnetCDF2Dataset` is returning `false` for the example dataset, but the values of `latitude`, `longitude`, and `altitude` arrays each include the same, repeated value...I assume those variables the `latitude`, `longitude`, and `altitude` of the radar? If so, why would `.isStationary()` be `false` if the values of the radar location do not actually change?

Other than `isStationary()`, the only other check is that the dataset can opened as a `Radial` `FeatureType`, which previously it could not.

lesserwhirls:
Hi @yuanho - I'd be happy to help out here, but I'm not sure of the expected behavior. Just let me know if you have any questions!

lesserwhirls:
Hi @yuanho - I think I've at least captured the previous behavior. Let me know if you think this looks ok and we can go ahead and merge it in.

lesserwhirls:
Also, @yuanho, you'll need to "sign" the CLA before we can merge (simple point and click kind of web form). Thanks!

lesserwhirls:
https://artifacts.unidata.ucar.edu/repository/unidata-snapshots/edu/ucar/ncIdv/4.6.12-SNAPSHOT/ncIdv-4.6.12-20180223.180656-18.jar

PR #1025, opened 2018-02-07
Enable and cleanup WMS tests

lesserwhirls:


cwardgar:
Nice. So previously we had no tests enabled for WMS? And also, I gather that the existing, commented-out tests aren't worth saving? I see lots of crufty Windows paths in there.

lesserwhirls:
Yeah, not really worth saving. edal-java and ncwms are tests on their end, so we just need to make sure the hooks are working. I will add more tests once I get the TDS setup to read the wms config file again, but for now, this at least covers the basics.

PR #1026, opened 2018-02-07
 Rewrite NCSS front-end

cwardgar:
The new NCSS pages mostly look and act like the old pages, but the implementation is completely different. The markup, CSS, and Javascript are all brand new. Almost none of the old stuff survived.

We're now using Thymeleaf templates for markup generation. We already had one template for Grid, but now we have ones for Grid-as-Point, Point, and Station as well. I moved several common fragments of markup to their own templates. Those fragments then get included in the main pages. As a result, there is very little duplication of markup.

We're now using modern CSS3 for all of our page layout. No more tables or floating elements; CSS Flexbox and Grid all the way! Note that Flexbox and Grid are recent technologies and are unsupported in older browsers. CSS Grid in particular was not supported in all major browsers [until Nov 2017](https://caniuse.com/#feat=css-grid). Perhaps I shouldn't have chosen to use such bleeding-edge tech, but layout would have been much harder and kludgy-er without it. Meh. People will just have to update their shit.

We're now using ECMAScript 6 WITHOUT jQuery. OpenLayers.js is still a dependency, and unfortunately we remain stuck on the 2.X line (4.X is the new stuff). The new Javascript largely does the same things that the old Javascript did, but in about a quarter of the code, due to better reuse of functions. The one big new feature is the accordion buttons on the Grid and Grid-as-Point pages.

Other notes:
* We are no longer hosting OpenLayers on THREDDS; instead, the NCSS pages grab the library from a CDN, which is best practice. A consequence of this is that the NCSS pages will not function 100% correctly if THREDDS is used in an environment without internet access. However, I've limited the breakage to the "Horizontal Extent" map: it simply won't load. The rest of the page will be functional.
* I developed the pages in Chrome, and that's where they look best. I've noticed one visual quirk in Firefox and two in Safari, but the pages otherwise work fine in those browsers. I don't really have the patience for browser-specific CSS tweaks.
* In this PR, I tried to limit my changes to the NCSS back-end to only what was necessary in order to get the templates working. There's a hell of a lot of cruft there that I want to clean up, but that will follow in a separate PR.
* To review this PR, you'll probably want to check out my branch and run the pages yourself (`./gradlew :tds:appRun`). Let me know if you see anything weird or broken.


lesserwhirls:
So, rather than playing locally, I figured it would just throw it up on thredds-dev:

https://thredds-dev.unidata.ucar.edu/thredds/catalog/catalog.html


1. javascript for url builder not working in safari

   Using Safari (Version 11.0.3 (13604.5.6)):

   On this page (featureType="Point")
   https://thredds-dev.unidata.ucar.edu/thredds/ncss/point/nws/synoptic/ncdecoded/Surface_Synoptic_Point_Data_fc.cdmr/dataset.html

   I noticed the URL builder isn't working. It just says "THIS WILL BE REPLACED BY buildAccessUrl().".

   same for this GRIB featureCollection:

   https://thredds-dev.unidata.ucar.edu/thredds/ncss/grid/grib/NCEP/GFS/Global_onedeg/GFS_Global_onedeg_20180207_1200.grib2/dataset.html

   Chrome (Version 64.0.3282.140 (Official Build) (64-bit)) - builder works

2. Initial Date/time on POINT and STATION issue

   On this page (featureType="Point")

   https://thredds-dev.unidata.ucar.edu/thredds/ncss/point/nws/synoptic/ncdecoded/Surface_Synoptic_Point_Data_fc.cdmr/dataset.html

   and this page (featureType="Station")

   https://thredds-dev.unidata.ucar.edu/thredds/ncss/point/nws/metar/ncdecoded-test/Metar_Station_Data_-_Test_fc.cdmr/dataset.html

   The start dates that are pre-populated seem...odd...`0000-01-01T00:00:00Z`. But, that's probably a backend issue, right?

lesserwhirls:
Also, we get http/https mixed content warnings like this:

~~~
OpenLayers.js:494 Mixed Content: The page at 'https://thredds-dev.unidata.ucar.edu/thredds/ncss/point/nws/synoptic/ncdecoded/Surface_Synoptic_Point_Data_fc.cdmr/dataset.html' was loaded over HTTPS, but requested an insecure image 'http://vmap0.tiles.osgeo.org/wms/vmap0?LAYERS=basic&SERVICE=WMS&VERSION=1.1.1&REQUEST=GetMap&STYLES=&FORMAT=image%2Fjpeg&SRS=EPSG%3A4326&BBOX=-180,-90,0,90&WIDTH=256&HEIGHT=256'. This content should also be served over HTTPS.
~~~

Is it possible to configure the map interface to use a tile server that has https enabled? This will save our .gov users some heartache when it comes to turning on NCSS (that is, they will have to turn it off because of these warnings).

cwardgar:
Nice catches. The first is caused by lack of support for `FormData.entries()` in Safari: https://stackoverflow.com/questions/41028769/safari-is-saying-formdata-entries-is-not-a-function

Grrrrr. I may need to write some browser-specific Javascript. What fun.

cwardgar:
`0000-01-01T00:00:00Z` is a back-end issue, yes. I have: 
```java
      CalendarDateRange calendarDateRange = dsgFeatCol.getCalendarDateRange();
      if (calendarDateRange == null) {
        CalendarDate start = CalendarDate.of(null, 0000, 1, 1, 0, 0, 0);  // 0000-01-01T00:00:00.00Z
        CalendarDate end   = CalendarDate.present();
        calendarDateRange  = CalendarDateRange.of(start, end);
      }
      model.put("calendarDateRange", calendarDateRange);
```
`0000-01-01T00:00:00Z` was sort of an arbitrary choice of default when no start date is available. Is there a more intuitive value? Should I just leave the fields blank?

lesserwhirls:
The default date is ok with me. It's possible the issue isn't backend code, but the dang netCDF files we produce (reverse linked list goodness).

cwardgar:
Okay, I've fixed all outstanding issues that I'm aware of. See individual commits for details. The big change is an upgrade to OpenLayers 4.X, which indirectly solves the Mixed Content problem. Once I set my mind on doing it, the transition actually wasn't that bad.

Sean, can you throw the new build up on thredds-test so that we can give it another round of testing? I'm fairly certain there are other bugs lurking.

lesserwhirls:
Ok, just bounced thredds-dev - very nice!

cwardgar:
Okay, I added one last commit. This PR is ready to merge, unless you discover some other bug.

Issue #1044, opened 2018-03-01
Failing Grid / Coverage Comparison test


lesserwhirls:
We are currently seeing failures when reading in `GRIB1`/`GRIB2` files via the `GridDataset` and `CoverageDataset` APIs and comparing the results (failures in test `ucar.nc2.grib.TestCoordinatesMatchGbx`).

As it stands, 157 `GRIB1` files are tested (of these, four generate failures), and 141 `GRIB2` files are tested (of these, three generate failures).

If one file fails, the entire test fails, and one must dig through the output to see how many actual files fail. In order to keep track of any "new" failing files, these known failures should be ignored for now, such that the tests, as they stand, pass.

This ticket is to document the known failures.

`GRIB1` files (`TestDir.cdmUnitTestDir/formats/grib1/`):
 * QPE.20101005.009.157
 * testproj.grb
 * testproj2.grb
 * wrf_d03_201308080200.grb
 * test.nc (not a `GRIB1` file, but in the `GRIB1` format directory?)

GRIB-2 files (`TestDir.cdmUnitTestDir/formats/grib2/`):
 * ds.pop12.bin
 * shtfl.grib2
 * shtfl.grib2.ncx4 (likely failure path as shtfl.grib2)
 * ds.mint.nc (not a `GRIB2` file, but in the `GRIB2` format directory?)

lesserwhirls:
Pull Request Unidata/thredds#1045 causes the files above to be ignored in the test, allowing us to track any new failures.

PR #1045, opened 2018-03-01
Ignore failing files in Grid/Coverage comparison

lesserwhirls:
See Unidata/thredds#1044 for information on which files are ignored.


lesserwhirls:
I don't *think* so, but I am still looking into it. Each case seems very different.

PR #1047, opened 2018-03-02
Improve and standardize floating-point comparisons

cwardgar:
We have a lot of duplicate code for comparing floating-point numbers, some of it bad. In this commit, I've tried to nuke dupes and concentrate the "core" comparison code into the `ucar.nc2.util.Misc` class. All other comparison code builds off of what's in `Misc`.

Notes:
* The centerpiece is the `Misc.nearlyEquals()` family of methods, formerly named `Misc.closeEnough()`. The previous implementation was actually pretty good, but I've improved it slightly for certain edge cases. See http://floating-point-gui.de/errors/comparison/.
* I've settled on the name "nearlyEquals" for any method that compares floats. Previously we had several different names for these, including "closeEnough", "fuzzyEquals", and "equals".
* Added/updated various ancillary methods/fields in `Misc` for comparing floats. They're used occasionally in our test suite.
* The new comparison code caused a handful of tests to break, usually because the default epsilon was too small for the level of precision that the tests were operating at. I believe I've fixed all of those.
* Added `nearlyEquals()` methods to `LatLonPoint`, `ProjectionPoint`, `LatLonRect`, and `ProjectionRect`. My need for these methods is what caused me to go off on this big refactoring tangent in the first place!
* For those `LatLon*` and `Projection*` classes, I made sure that each also has an `equals()` method that performs exact floating point comparison, which is required for consistency with `hashCode()`.
* Refactored `LatLonPointImmutable` to extend `LatLonPointImpl` (so that it can inherit useful methods), while remaining immutable.

This PR also includes some unrelated tweaks to NCSS, because I was working on NCSS before the floating-point stuff.
* In NCSS, add a label to the lat/lon point subset pane.
* Rewrite `CoverageCoordSys.makeCoorSysName()` to eliminate extraneous whitespace.
* Nuke `CoverageCollection.getRuntimeCoordinateMax()`, which NCSS no longer needs.

I tested these commits on Jenkins. [Here](https://jenkins-aws.unidata.ucar.edu/job/cwardgar/12/testReport/) is the results. Pretty clean. I cannot account for the `TestGribUnits` failures; they don't appear to be related to the `nearlyEquals()` changes.

### DIFF
--- a/cdm-test/src/test/java/ucar/coord/TestCoordinateND.java
+++ b/cdm-test/src/test/java/ucar/coord/TestCoordinateND.java
@@ -42,8 +42,8 @@ public class TestCoordinateND {
     System.out.printf("%s%n", f);
     System.out.printf("reindexed %s%n====================%n", counter.show());
 
-    assert Misc.closeEnough(curr.getSparseArray().getDensity(), 1.0);
-    assert Misc.closeEnough(reindexed.getSparseArray().getDensity(), .826446);
+    assert Misc.nearlyEquals(curr.getSparseArray().getDensity(), 1.0f);
+    assert Misc.nearlyEquals(reindexed.getSparseArray().getDensity(), .826446f);
   }
 
   static public CoordinateND<Short> makeCoordinateND(int rank, int size) {
--- a/cdm-test/src/test/java/ucar/coord/TestSparseArray.java
+++ b/cdm-test/src/test/java/ucar/coord/TestSparseArray.java
@@ -34,7 +34,7 @@ public class TestSparseArray {
     Formatter info = new Formatter(System.out);
     sa.showInfo(info, null);
 
-    assert Misc.closeEnough(sa.getDensity(), 0.906667);
+    assert Misc.nearlyEquals(sa.getDensity(), 0.906667f);
 
   }
 }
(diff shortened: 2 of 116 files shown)


cwardgar:
It occurs to me that we could benefit from some custom JUnit assert methods, because we're doing a lot of stuff like:
```
Assert.assertTrue(Misc.nearlyEquals(timeOffset, ptime.getForecastTime()));
```
Perhaps `Assert.assertNearlyEquals(timeOffset, ptime.getForecastTime())`. `Assert.assertArrayNearlyEquals()` would be really useful too. Not sure if I want to tackle that right now though.

cwardgar:
I'd argue that `Misc` really isn't part of our "core" API. Making changes to `NetcdfFile` ought to be done with care, and possibly a deprecation notice, but for a class like `Misc`? I think limiting the modification to a major release is good enough.

But yeah, it's definitely worthwhile to document it in the release notes.

lesserwhirls:
@DennisHeimbigner - how does this look to you?

DennisHeimbigner:
> I'd argue that Misc really isn't part of our "core" API. 
This is a good point. Are there two sets of equality tests:
one for use inside netcdf-java/thredds and one for use by
our test cases? Also, is all of Misc appropriate to include
in netcdf-java/thredds, or is some part of it test case use only?

cwardgar:
A lot of the methods in `Misc` are only used by our tests, but that's simply because that is the only context where they are useful. But I see no reason why any of it (including `nearlyEquals()`) can't be used in both source sets.

lesserwhirls:
@DennisHeimbigner - if you think this is good, would you mind approving the PR? Thanks!

PR #1055, opened 2018-03-09
Update README.md

lesserwhirls:
sonarcloud changed their badge API - updated to new URL


lesserwhirls:
Yeah, I could have kept the older version, but I like this one better, too. The old badge will stop working on April 2nd: https://about.sonarcloud.io/news/2018/02/05/new-project-badges.html

PR #1056, opened 2018-03-11
 Calculate correct boundaries for coverages that cross longitude seam

cwardgar:
Added `HorizCoordSys.calcConnectedLatLonBoundaryPoints()`, which calculates the lat/lon boundary of the CRS.
* It has optional parameters that limit the resolution.
* It works for projection, latLon1D, and latLon2D CRSs.
* Its boundary points are "connected", meaning that longitude values of adjacent elements are adjusted to eliminate discontinuities resulting from crossing the seam. Consequently, when cross-seam boundaries are turned into WKT polygons and plotted in NCSS's OpenLayers map, they'll be properly interpreted.
* Similar code used to live in `CoverageBoundariesExtractor`, but it was incomplete and/or buggy. I nuked all the old stuff.

Rewrote `HorizCoordSys.calcLatLonBoundingBox()` to properly handle coverages that straddle the seam.
* It builds upon `calcConnectedLatLonBoundaryPoints()`. In fact, its implementation is quite simple once we have that.
* `HorizCoordSys2D` used to have its own implementation, but the `HorizCoordSys` version now handles all CRSs properly. Nuked it.

Added extensive Javadoc for the new/updated boundary methods.
* It was quite a tricky problem, as the docs will show. The ultimate solution turned out to be relatively simple, but it took awhile for me to get there (and realize that it worked).

Added `HorizCoordSysCrossSeamBoundarySpec`, which extensively tests the boundary methods.
* It builds `HorizCoordSys` objects from the new `crossSeam*.ncml` test resources.

Did some other minor refactoring in `HorizCoordSys`, and updated the classes that depend on it.
* Renamed `getIsProjection()` to `isProjection()`.
* Renamed `makeProjectionBB()` to `calcProjectionBoundingBox()`.
* Renamed `makeLatLonBB()` to `calcLatLonBoundingBox()`.
* Converted axis field names to camelCase.

`DtCoverageAdapter` now produces 1D longitude axes with "connected" values. It appears that we used to do that in the analogous `dt.grid` stack, but that the adjustment was lost in the transition to coverage. Hopefully this change doesn't break anything.

Added `LatLonPointNoNormalize`, a container in which latitudes may lie outside of [-90, 90] and longitudes may lie outside of [-180, 180].

### DIFF
--- a/cdm/src/main/java/ucar/nc2/ft2/coverage/HorizCoordSys.java
+++ b/cdm/src/main/java/ucar/nc2/ft2/coverage/HorizCoordSys.java
@@ -36,17 +36,12 @@ package ucar.nc2.ft2.coverage;
 import com.google.common.collect.Lists;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import ucar.ma2.InvalidRangeException;
-import ucar.ma2.MAMath;
-import ucar.ma2.RangeIterator;
+import ucar.ma2.*;
 import ucar.nc2.util.Optional;
 import ucar.unidata.geoloc.*;
 
 import javax.annotation.concurrent.Immutable;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Formatter;
-import java.util.List;
+import java.util.*;
 
 /**
  * Horizontal CoordSys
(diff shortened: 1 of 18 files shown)


cwardgar:
With this done, I finally have what I need to restore NCSS's `datasetBoundaries.xml` endpoint. PR forthcoming.

cdm/src/main/java/ucar/nc2/ft2/coverage/HorizCoordSys.java
```diff
@@ -61,81 +56,88 @@
  * @author caron
  * @since 7/11/2015
  */
+// TODO: This class is trying to do too much: it's handling projection, latLon1D AND latLon2D CRSs. We should break
```
lesserwhirls: This is a good idea - can you open an issue for this, tagged with netCDF-java beta as the milestone?

lesserwhirls:
👍

PR #1058, opened 2018-03-15
Restore NCSS's datasetBoundaries endpoints

cwardgar:
"/datasetBoundaries.xml" is working once again. It returns a coverage's boundary as a WKT polygon. The endpoint also has an "accept" parameter, which can have values "wkt" or "json". "/datasetBoundaries.xml?accept=json" returns the boundary as a GeoJSON polygon. The code for this return type is present in the latest 4.6.X, but commented-out. Now it's back.

Clearly, "/datasetBoundaries.xml" is a terrible name for the endpoint, because it doesn't actually return XML. For backwards compatibility, I'm keeping it around, but I've added "/datasetBoundaries.wkt" and "/datasetBoundaries.json" endpoints as well. It's recommended that people use those instead.

Notes:
* Moved methods that render dataset boundaries as GIS text out of `CoverageBoundariesExtractor` and into `HorizCoordSys`. Nuked the former class, and its test.
* Added unit tests for the new `HorizCoordSys` methods.
* Added Spring integration tests for `NcssGridController.getDatasetBoundaries*` methods. Those methods actually implement the endpoints.
* Updated NCSS Asciidoc to include new endpoints. @lesserwhirls You may want to make a note of this.
* Updated NCSS's inline Javascript code to gracefully handle null values for the lat/lon bounding box, projection bounding box, and calendarDateRange. A dataset might not have a typical coordinate system, and could return "null" for any of those.


lesserwhirls:
nice improvement!

PR #1062, opened 2018-03-20
First 5.0 beta release

lesserwhirls:
Includes copyright and license updates


dopplershift:
🎉

cwardgar:
I'm curious: why are we retaining the "John Caron and ..." copyright? Almost all of John's work was done while he was employed at UCAR, which makes it UCAR's property, right?

Also, must we include the date range (1998-2018) in every source file? That's sure to become stale.

msdsoftware:
Wondering about @cwardgar's questions myself. Mostly due to the code here, but somewhat due to dealing with "copyright x-y" headers in my own code.

lesserwhirls:
So two things:

1. "John Caron and ..." copyright. In the cases I looked at, he made big changes to the code after he left, and he still retains copyright to those changes. I think he only started using that license header after he left Unidata. I am not a lawyer, but here is how I understand things in the general case. When someone submits a PR to Unidata and signs the CLA, they agreed that they have the right (permission from employer, complete ownership, etc.) to submit those changes, to allow Unidata to use those changes into perpetuity, and that those changes are now licensed by Unidata under the terms set fourth in the LICENSE, and that the license can be changed to something more permissive in the future, but not more restrictive, without asking them explicitly (although BSD3 seems pretty open, so I do not foresee this happening). They, however, are still the copyright owners of those changes. So I don't think it would be quite right to just have UCAR/Unidata in the file level copyright block when others have contributed outside of UCAR/Unidata. That said, I think he is the only one who contributed code and modified the copyright block to add his name. Perhaps we should use something like the following in the file level block instead:

~~~java
/*
 * Copyright (c) 1998-2018 University Corporation for Atmospheric Research/Unidata and Contributors
 * See LICENSE for license information.
 */
~~~ 

Then, we have one file level copyright block across the project for code developed here at Unidata. We do have other code that was used from other projects, following the terms set out in their license, and I tried really hard to make sure this PR does not affect any of that code. If someone wants to know copyright info, they can look at the commit history of a particular file to see who shares copyright over the code. Who owns copyright does not determine who can use it, as the code was contributed under the terms of the license at the top of the repo, which sets fourth use. Sorry for the long answer here, but again, this is how I understand things (and this is after consultation with UCAR legal over the years. @dopplershift @ethanrd - thoughts?

2. Dates. According to UCAR legal, the dates should be kept at the file level, but only updated when a major change to the particular file has been made. Define major? Well, what is `is`? I don't know the answer to that. Relicensing seems like a major change, so I updated them all. A bug fix in 2019? Maybe update the date? A doc string typo fix in 2019? Maybe not? A new method? For sure.

dopplershift:
@lesserwhirls 's summary matches my best understanding. The other thing to do is generate a file like `AUTHORS.txt` or `CONTRIBUTORS.txt` and lay out who has contributed to the repository. You can automatically generate this from the git history using:
```sh
git log --format='%aN' | sort -u
```
and making judicious use of a `.mailmap` file to resolve duplicate people down to a single canonical author.

cwardgar:
Perhaps we can look into "UCAR and Contributors" in the future, after generating an appropriate `CONTRIBUTORS.txt`. Of course, that's probably something we'd want to run past UCAR legal.

This looks good for now. :+1:

msdsoftware:
I think I've mentioned this elsewhere, but I'll just throw it out there again to see fi anyone picks it up. Could someone write up a guide for updating 4.6-based code to 5.0?

I took a look at updating the Panoply to use 5.0. Only about 25 or 30 compiler errors. Some easily worked out, but around 20 involving threads.catalog classes looked to be a hairball.

cwardgar:
@msdsoftware We have an [upgrading to v5 doc](https://github.com/Unidata/thredds/blob/5.0.0/docs/website/tds/UpgradingTo5.adoc) that contains most of the big changes. As for the `thredds.catalog` classes, they're still around, but now in `legacy.jar`.

msdsoftware:
Thanks, @cwardgar. I've no idea how I didn't run across that before.

PR #1066, opened 2018-03-23
Enable logging interface thru Nc4prototypes.java

DennisHeimbigner:
1. Added nc_set_log_level to the Nc4prototypes interface.
2. Modified Nc4Iosp.load() so that it
   tries to set the log level to 0 (zero).
   This will provide at lease HDF5 traceback
   plus probably some other (hopefully) minimal
   netcdf-c output.
3. Defined a -Djna.library.loglevel for the JVM
   to control what log level to set. Setting to -1
   will turn off logging.


lesserwhirls:
I'll update our test server with these changes and wait for the next core dump - thanks @DennisHeimbigner

Issue #1065, opened 2018-03-21
Not detecting _FillValue of scaled data correctly


guygriffiths:
This comes from a bug report for EDAL (https://github.com/Reading-eScience-Centre/edal-java/issues/109).

In the test data available at that bug report, the fill values are not being detected properly.  The variable `tmp` is stored as an integer and has a `_FillValue` of `99999` and a `scale_factor` of `1.e-05`.  By the time we get to the `isMissing_()` method in [ucar.nc2.dataset.EnhanceScaleMissingImpl](https://github.com/Unidata/thredds/blob/5.0.0/cdm/src/main/java/ucar/nc2/dataset/EnhanceScaleMissingImpl.java), the value has been scaled to `0.99999`.  However, when that method calls `isFillValue()` on that, it returns `false`, since it is testing against the unscaled value (i.e. 99999).

cwardgar:
Fixed in #1067. Thanks for the report!

guygriffiths:
Perfect, thanks.  Would you be able to cut a `5.0.0-alpha4` release including this?  We're currently depending on `5.0.0-alpha3` and it'd be nice to have this change.

lesserwhirls:
@guygriffiths - sorry for the delay on our end! `5.0.0-beta2` has the fix and is now available as a release on our nexus server.

PR #1067, opened 2018-03-23
Fix scale_factor not being applied to _FillValue

cwardgar:
The bug is described in https://github.com/Unidata/thredds/issues/1065.

Added a regression test and fixed another that broke due to the change.


cwardgar:
It turns out that there is a `setUseNaNs()` method on `VariableDS`. That was handy.

Issue #1048, opened 2018-03-02
NCSS return confusing for lat/lon data


dopplershift:
For a straightforward request for data from models on lat/lon grids (say: http://thredds.ucar.edu/thredds/ncss/grib/NCEP/GFS/Global_0p5deg/GFS_Global_0p5deg_20180302_0000.grib2?var=Temperature_isobaric&disableLLSubset=on&disableProjSubset=on&horizStride=1&time_start=2018-03-02T00%3A00%3A00Z&time_end=2018-03-18T00%3A00%3A00Z&timeStride=1&vertCoord=&accept=netcdf4), the return doesn't seem to be CF-compliant:

```
netcdf GFS_Global_0p5deg_20180302_0000.grib2 {
dimensions:
	time1 = 93 ;
	isobaric = 31 ;
	lat = 361 ;
	lon = 720 ;
variables:
	float Temperature_isobaric(time1, isobaric, lat, lon) ;
		Temperature_isobaric:long_name = "Temperature @ Isobaric surface" ;
		Temperature_isobaric:units = "K" ;
		Temperature_isobaric:abbreviation = "TMP" ;
		Temperature_isobaric:missing_value = NaNf ;
		Temperature_isobaric:grid_mapping = "LatLon_Projection" ;
		Temperature_isobaric:coordinates = "time1 isobaric lat lon " ;
		Temperature_isobaric:Grib_Variable_Id = "VAR_0-0-0_L100" ;
		Temperature_isobaric:Grib2_Parameter = 0, 0, 0 ;
		Temperature_isobaric:Grib2_Parameter_Discipline = "Meteorological products" ;
		Temperature_isobaric:Grib2_Parameter_Category = "Temperature" ;
		Temperature_isobaric:Grib2_Parameter_Name = "Temperature" ;
		Temperature_isobaric:Grib2_Level_Type = 100 ;
		Temperature_isobaric:Grib2_Level_Desc = "Isobaric surface" ;
		Temperature_isobaric:Grib2_Generating_Process_Type = "Forecast" ;
	double time1(time1) ;
		time1:units = "Hour since 2018-03-02T00:00:00Z" ;
		time1:standard_name = "time" ;
		time1:long_name = "GRIB forecast or observation time" ;
		time1:calendar = "proleptic_gregorian" ;
		time1:_CoordinateAxisType = "Time" ;
	float isobaric(isobaric) ;
		isobaric:units = "Pa" ;
[...]

PR #1069, opened 2018-03-24
Don't exclude lat/lon CoverageTransforms from CoverageCollections

cwardgar:
Fixes https://github.com/Unidata/thredds/issues/1048.
Added a regression test.


cwardgar:
Running a Jenkins job just to be safe. Result will be here: https://jenkins-aws.unidata.ucar.edu/job/cwardgar/17/

cwardgar:
Okay, this is okay to merge now: https://jenkins-aws.unidata.ucar.edu/job/cwardgar/18/testReport/

See additional commit for notes.

PR #1073, opened 2018-03-29
use isStride1 to avoid convertSizeT if section.getStride() is empty

### LINKED ISSUE #1072: writeEnumData raises NPE with empty section.getStride()
Hello, 

I have a dataset of netcdf4 files and I try to read/write this dataset using netcdf library version **4.6.11**

Reading the files works but writing fails on 

> Nc4Iosp::writeEnumData()

This is an extract of my test

```java
NetcdfFile ncFile = NetcdfFile.open(inputFile.getAbsolutePath());
final Nc4Chunking chunker = Nc4ChunkingStrategy.factory(Strategy.standard, 0, true);
final FileWriter2 fw = new FileWriter2(ncFile, outputFileName, NetcdfFileWriter.Version.netcdf4, chunker);
    fw.write();
```
The problem comes from Nc4Iosp::writeEnumData(). 
Line 2946 
```java
SizeT[] stride = convertSizeT(section.getStride());
```
the 'stride' reference is null. This is because method 'convertSizeT' does:

```java
if (from.length == 0) return null;
```

And just after in writeEnumData their is a loop trying to use 'stride.length' (which in my case is null)
```java
for(int i = 0; i < stride.length; i++)
```

I will try to provide a fix for this problem.

Regards,
Robin

PS: Unfortunately I cannot provide you the dataset I'm using.

r0perice:
Fixes #1072

### DIFF
--- a/netcdf4/src/main/java/ucar/nc2/jni/netcdf/Nc4Iosp.java
+++ b/netcdf4/src/main/java/ucar/nc2/jni/netcdf/Nc4Iosp.java
@@ -3032,24 +3032,18 @@ public class Nc4Iosp extends AbstractIOServiceProvider implements IOServiceProvi
         SizeT[] shape = convertSizeT(section.getShape());
         boolean isUnsigned = isUnsigned(typeid);
         int sectionLen = (int) section.computeSize();
-        SizeT[] stride = convertSizeT(section.getStride());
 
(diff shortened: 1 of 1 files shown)


cwardgar:
@DennisHeimbigner Do you agree with my analysis? You are the code owner, so please merge this if you do.

PR #1086, opened 2018-04-14
 Reorganize docs and publish Jekyll-generated user guide

cwardgar:
I also updated the docs and schema to clarify that the `documentation` element doesn't support XHTML.

It's best to review the 3 commits individually, because the reorg commit accounts for 99% of this PR's diff and makes the other changes hard to see.

This PR will also serve as a test of the new "thredds-doc" Jenkins job. Once merged, the job should be triggered, which will build the user guide and publish it to Nexus.


lesserwhirls:
I though we decided at our meeting last week that would we keep the line about XHTML, but state the currently the TDS does not support it. The idea was that others use the schema to write and generate catalogs, and they might actually use that "feature" that we've dropped on the floor in terms of TDS support. I could totally be misremember the conversation, though. Other than that, the changes look good to me.

cwardgar:
@lesserwhirls We did decide that; these changes were made before our meeting.

So you recommend leaving the comments in the schema intact? Or changing the text in the doc? Feel free to add a commit to my branch, if you want.

cwardgar:
Merging this PR caused this Jenkins job to run: https://jenkins-aws.unidata.ucar.edu/job/thredds-doc/10/

We now have auto-building and auto-publishing docs!

Issue #1041, opened 2018-02-26
Time Animation not working for netCDF ncWMS request in THREDDS 5.0


chhonker:
[http://localhost:8080/model/wms/testAll/ModelTemperature.nc?service=WMS&version=1.3.0&request=GetMap&LAYERS=TEMP&STYLES=default-scalar/default&WIDTH=1000&HEIGHT=1000&FORMAT=image/gif&CRS=CRS:84&BBOX=-180,-90,180,90&EXCEPTION=application/vnd.ogc.se_xml&time=2004-01-03T12:00:00.000Z/2004-02-22T12:00:00.000Z&elevation=215&animation=true](url)

With the same request, I was able to get gif file with aminated layers in version 4.6.11
But, with the same request, not able to get animation for version 5.0
Is this query correct?

PR #1099, opened 2018-05-28
Fix WMS time parameter not being used with GetMap

geojs:
The bug is described in:
https://github.com/Unidata/thredds/issues/968
https://github.com/Unidata/thredds/issues/1041


lesserwhirls:
Thank you for this contribution @geojs!

PR #1101, opened 2018-06-02
Refactor and polish unsigned conversions

cwardgar:
* `DataType`'s unsigned conversion methods (named `unsignedXToY`) now defer to standard library methods that were added in JDK 8. Nuked our existing implementations.
* `DataType.widenNumber()`, in turn, employs those `unsignedXToY` methods. Greatly expanded its Javadoc.
* Also added `DataType.widenNumberIfNegative()`, which is similar to `widenNumber()`, but only widens when the argument is a negative, integral value. This is an optimization to improve performance.
* Added 4 new tests of the widen methods to `TestDataType`. Actually, they're not exactly new; they're the result of splitting apart the (now-deleted) `TestMAMath.convertUnsigned()`. That test logically had no business being in `TestMAMath`.
* After `TestMAMath.convertUnsigned()` was removed, the corresponding method `MAMath.convertUnsigned()` was no longer being used anywhere. No reason to keep it around. Nuked.
* Similarly, `DataType.unsignedLongToString()` wasn't being used anywhere, except in `TestDataType.testUnsignedLongToString()`. Nuked both.

* `Array.toString()` and `NcDumpW.printArray()` now use `DataType.widenNumberIfNegative()`. Improved nearby comments.

* Nuked `Attribute.isUnsigned()`. That info can be obtained via examining the `DataType` that's available via `Attribute.getDataType()`. This encourages users to work with `DataType` objects directly.
* Simplified `Nc4Iosp.writeAttribute()`, which was needlessly checking both `Attribute.getDataType()` and `Attribute.isUnsigned()`. `DataType` now includes signedness info! No need to check both. Besides, `Attribute.isUnsigned()` is no longer available.

### DIFF
--- a/cdm/src/main/java/ucar/ma2/Array.java
+++ b/cdm/src/main/java/ucar/ma2/Array.java
@@ -1128,16 +1128,19 @@ public abstract class Array {
     StringBuilder sbuff = new StringBuilder();
     IndexIterator ii = getIndexIterator();
     while (ii.hasNext()) {
-      Object data = ii.getObjectNext();
-      if (data instanceof Number && isUnsigned()) {
-        // 'data' is unsigned, but will be treated as signed when we print it below, because Java only has signed
-        // types. If it is large enough ( >= 2^(BIT_WIDTH-1) ), its most-significant bit will be interpreted as the
-        // sign bit, which will result in an invalid (negative) value being printed. To prevent that, we're going
-        // to widen the number before printing it.
-        data = DataType.widenNumber((Number) data);
+      Object value = ii.getObjectNext();
+  
+      if (isUnsigned()) {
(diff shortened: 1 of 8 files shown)


cwardgar:
I think this PR is already okay, but I'm running a Jenkins job to be sure. Result will be here: https://jenkins-aws.unidata.ucar.edu/job/cwardgar/23/

cwardgar:
@DennisHeimbigner I have a PR in the works that addresses that concern. Widening will be performed during dataset enhancement, but only if the `convertUnsigned` option is selected. Data will also be widened (if necessary) when they are printed, as by NCdumpW, for example.

I'm pretty sure those are the only two places, but I could be wrong.

PR #1102, opened 2018-06-05
Make NCSS accessible by default for fresh installation of TDS

cwardgar:
* Enabled NCSS in threddsConfig.xml.
* Nuked most of the default datasets. We now have 3 new ones that can be used to test the GRID, GRID-AS-POINT, POINT, and STATION pages of NCSS.
* The "all" compound service element in catalog.xml now contains all services that we provide.
* Updated test resource for our freshInstallTest to account for the catalog changes.


cwardgar:
While I wasn't able to reproduce the NullPointerException that Hailey was getting, I did notice that NCSS is not easy to test with a fresh installation. At least that is fixed now.

Issue #669, opened 2016-10-28
TDS Beautification


cwardgar:
Let's admit: the pages that TDS generates look like ass. None of us on staff have great front-end design skills, but that doesn't mean that we can't do any better. There's a billion CSS frameworks out there that do most of the work for you; [Bootstrap](http://getbootstrap.com/) is the most popular one right now. It doesn't look too tough to learn. Furthermore, we can get a good head start–as well as plenty of design ideas–from the many [freely available themes](https://startbootstrap.com/).

To me, this issue depends on #444. Trying to apply Bootstrap to our current within-Java HTML generation is a non-starter.

adamsteer:
+1000 - my reason for #444  was that would rather see development focussed on a templating engine  - for example, someplace in the TDS framework sits a (set of) template files which, by default, render TDS pages in plain HTML.

UX/UI folks can go to town on these - add hooks for bootstrap or whatever styling system they like, but all the developers need to do is make sure that the required page elements are all there. Like Jinja2 templating for Flask apps.

That way Java developers don't need to care about styles, or choosing a styling framework, and the UI/UX folks don't need to care about Java. (although maybe bootstrap styles by default would be nice.. and easy). Also get in touch if you need a hand to build templates once an engine is decided upon - HTML used to be my second language.

ccarleton-noaa:
Whatever you do make sure that it doesn't break in an HTTPS-only
environment with HSTS.

kwilcox:
FWIW: it is possible to make TDS look a __bit__ better using the CSS that the existing HTML pages already reference:

* http://thredds.cencoos.org
* http://thredds.axiomdatascience.com/

julienchastang:
I took the liberty of reusing the styling work of Axiom here: http://jetstream.unidata.ucar.edu/thredds/catalog.html

BTW, this is our TDS on the XSEDE Jetstream Cloud. It makes use of the docker work we did recently. IMO the Axiom CSS could be the default CSS of the TDS, minus the logo. That would at least be a step up over what we have now.

dopplershift:
Why for the love of all that is holy is TDS by default using a monospace font for 90% of the content on the page?

<snark>All that's missing is an image of some orange cones, an "under construction banner", and some links to a geocities web ring</snark>

😈 

(Sorry, I couldn't help myself.)

lesserwhirls:
Added in Unidata/thredds#1109 and Unidata/thredds#1110 - yay @haileyajohnson !

Issue #444, opened 2016-02-23
HTML templating


adamsteer:
This is pretty much a wish - relating to style rather than function. Is it possible to get thredds to use a HTML templating system? Being able to customise style sheets is great, but I can't yet see any way of gaining more control without becoming a Java developer. Yes, many clients just use machines to read our catalogs, but many actual humans come and look too.

cwardgar:
This is a long-neglected deficiency of the TDS. Modern webapps use an MVC approach, where the model (the data displayed on the page), the view (the page layout and styling), and the controller (the logic that generates the response) are separate. This loose coupling enables HTML templating technologies to be used for the view component.

Currently, we're doing everything in the controller; there is no discrete model or view. [Here](https://github.com/Unidata/thredds/blob/5.0.0/tds/src/main/java/thredds/core/ConfigCatalogHtmlWriter.java#L126) is where we generate the HTML for catalogs. _Shudder_. JavaServer Pages addressed the problem of writing HTML in Java code in **1999**, but here we are.

I like [Thymeleaf](http://www.thymeleaf.org/) for this. It uses *natural templates*--HTML that can be correctly displayed in browsers and also work as static prototypes. Thymeleaf is the preferred template engine for Spring, so this won't be as much work as it would be otherwise.

cwardgar:
It occurs to me that we're also generating a lot of XML from within Java as well, such as [here](https://github.com/Unidata/thredds/blob/5.0.0/cdm/src/main/java/ucar/nc2/ft/point/writer/FeatureDatasetCapabilitiesWriter.java). That seems bad as well. Luckily, [Thymeleaf can process XML input](http://www.thymeleaf.org/doc/tutorials/3.0/usingthymeleaf.html#what-kind-of-templates-can-thymeleaf-process).

DennisHeimbigner:
What web pages do we generate now?
I can think of: catalogs and ncss. What else? WMS?

cwardgar:
We also generate TDS admin pages and OpenDAP forms in Java. The WMS pages are built by the edal-wms dependency.

lesserwhirls:
Added in Unidata/thredds#1109 and Unidata/thredds#1110

Issue #250, opened 2015-10-22
Replace `HTMLWriter`


dopplershift:
There's got to be a better way than adding hand-coded tags to `StringBuilder`.

JohnLCaron:
Been some changes in 5.0, but still hand-coded.
Could try thymeleaf template...

dopplershift:
Ideas from 4/21/2016 rant session:
- Move current catalog format to a thymeleaf template, preferably outside WAR and in content directory so that downstream users can modify easily (5.0)
- Investigate using modern HTML5 + JS + CSS frontend development with template (5.1)

lesserwhirls:
Added in Unidata/thredds#1109 and Unidata/thredds#1110

PR #1103, opened 2018-06-12
Hajohns/catalog templates

haileyajohnson:
Switch catalog and dataset access pages to Thymeleaf templates - lots left in the "TODO" for the templates, but those will likely become moot when we update the design


tds/src/main/java/thredds/server/viewer/Viewer.java
```diff
@@ -34,6 +33,13 @@
...
    */
-   String getViewerLinkHtml( Dataset ds, HttpServletRequest req);
+  String getViewerLinkHtml( Dataset ds, HttpServletRequest req);
+
+  /**
+   ** @param ds the dataset to view
+   * @param req the request
+   * @return HTML fragment string
+   */
+  ViewerLinkProvider.ViewerLink getViewerLink(Dataset ds, HttpServletRequest req);
```
haileyajohnson: I realized after I started down this path that this is a public interface and will probably break compatibility for some people, but we do need a way to import the raw (not html-ed) link into the templates...

lesserwhirls: I think we're ok. It should be pretty straight forward for people upgrading viewers from 4.6 to 5.0 to implement the new `getViewerLink` method.

tds/src/main/webapp/WEB-INF/templates/catalog.html
```diff
@@ -0,0 +1,58 @@
...
+                <td align="right">&nbsp;<code th:text="${subset.getDataSize()}"></code></td>
+                <td align="right"><code th:text="${subset.getLastModified() != null ? subset.getLastModified() : '&nbsp--'}">--</code></td>
+            </tr>
+
+        </tbody>
+    </table>
+
+    <hr size="1" noshade="noshade">
+
+    <h3><a href="/thredds/catalog.html">Initial TDS Installation (please change threddsConfig.xml)</a> at <a href="http://www.my.site/">My Group</a> see <a href="/thredds/info/serverInfo.html"> Info </a><br>
```
lesserwhirls: Does this get changed if server info in `threddsConfig.xml` has been changed?

tds/src/main/webapp/WEB-INF/templates/commonFragments.html
```diff
@@ -4,24 +4,14 @@
 <!--/* The indentation of the fragments below matches the indentation of the elements that they will replace in the
        target templates. Do not normalize it, or else the generated HTML won't be pretty any more. */-->
 
-<head th:fragment="head(title, scripts)">
-  <title th:replace="${title}">NetCDF Subset Service</title>
```
lesserwhirls: Thymeleaf used to only be used for the netCDF Subset Service, but now thanks to you, we're rocking it out everywhere. However, it would still be good to make sure the NetCDF Subset Service pages have the proper title. Maybe it already does and this is from my lack of understanding the templating system. If so, my bad! If not, we should open an issue to make sure we get that working at some point.

lesserwhirls:
Addresses the following open issues:

Unidata/thredds#250
Unidata/thredds#444
Unidata/thredds#669

w00t!

tds/src/main/java/thredds/server/viewer/Viewer.java
lesserwhirls: BTW - totally awesome that you are considering compatibility :-)

PR #1107, opened 2018-06-17
Hack to get dap4 TestServlet to work both under Intellij and Jenkins/Travis

DennisHeimbigner:
The focus here was two-fold.

First was getting TestServlet to work under Intellij using
Spring+Mocking.  I attempted a variety of approaches to get this
work, but ultimately gave up because the Thredds configuration
code fails in ChronicleMap when executing under Intellij.  So, I
gave up and used a hack as described in Dap4Controller.java.

Second was to experiment with using a lock around the Nc4Prototype
code to guarantee serialized access. This was because I was suspcious
that the Native.synchronizedLibrary operation might not be working
correctly and was causing the HDF5 library to fail. However, looking
at the Jenkins runs, I believe that Native.synchronizedLibrary is
working correctly. We shall see.

Note: TestServlet is still failing under Jenkins because of and HDF5 error.

### DIFF
--- a/dap4/d4lib/build.gradle
+++ b/dap4/d4lib/build.gradle
@@ -7,6 +7,7 @@ apply from: "$rootDir/gradle/any/publishing.gradle"
 dependencies {
     compile project(':dap4:d4core')
     compile project(':httpservices')
+    compile project(':netcdf4')
 
     compile libraries["jna"]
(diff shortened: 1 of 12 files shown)


lesserwhirls:
I'll get you the zip file of our jenkins build machines this evening.

PR #1110, opened 2018-06-30
add documentation for customizing the appearance of TDS

haileyajohnson:


tds/src/main/webapp/WEB-INF/altContent/startup/threddsConfig.xml
```diff
@@ -51,7 +51,6 @@
      <standardCssUrl>tds.css</standardCssUrl>
      <catalogCssUrl>tdsCat.css</catalogCssUrl>
      <datasetCssUrl>tdsDst.css</datasetCssUrl>
-     <openDapCssUrl>tdsDap.css</openDapCssUrl>
```
haileyajohnson: Even though OPeNDAP has it's own stylesheet, there's no hook for user-supplied yet

tds/src/main/webapp/tds.css
```diff
@@ -336,6 +336,7 @@ div.button {
     display: none;
     padding: 5px 10px;
     border: 1px solid #CFCFCF;
+    min-height: 120px;
 }
```
haileyajohnson: This helps prevent elements from shifting when tabs resize, but the only way I can think of to eliminate it entirely is fixed height and scrolling overflow, which is...an option

docs/src/public/userguide/pages/tds_tutorial/tds_configuration/CustomizingTds.md
```diff
@@ -0,0 +1,138 @@
...
+   <catalogCssUrl>catalog.css</catalogCssUrl>
+   <datasetCssUrl>dataset.css</datasetCssUrl>
+</htmlSetup>
+~~~
+
+Values given for `catalogCssUrl` and `datasetCssUrl` will be applied as stylesheets on the Catalog and Dataset
+access HTML pages, respectively. A path supplied to `standardCssUrl` will be used in all generated HTML pages.
+
+Note: Paths to CSS files may be absolute or relative, relative paths must be relative to the webapp URL,
+ i.e. http://server:port/thredds/.
```
lesserwhirls: It might be good to add that if the css files defined in `threddsConfig.xml` are located in `tds.content.root.path/thredds/public` directory, everything should work. So, by putting a file called `standard.css` inside `tds.content.root.path/thredds/public`, no changes to threddsConfig.xml would need to be made and the css file would be used.

That said, how would someone add a css file such that it would show up relative to `http://server:port/thredds/`? Would they need to copy it into the WEB-INF directory of the exploded war file?

tds/src/main/webapp/WEB-INF/altContent/startup/threddsConfig.xml
```diff
@@ -51,7 +51,6 @@
      <standardCssUrl>tds.css</standardCssUrl>
```
lesserwhirls: Should these be changed to match what you have in the docs?

~~~xml
<htmlSetup>
   <standardCssUrl>standard.css</standardCssUrl>
   <catalogCssUrl>catalog.css</catalogCssUrl>
   <datasetCssUrl>dataset.css</datasetCssUrl>
</htmlSetup>
~~~

tds/src/main/webapp/tds.css
lesserwhirls: Scrolling overflow, in general, makes me sad 😢

docs/src/public/userguide/pages/tds_tutorial/tds_configuration/CustomizingTds.md
haileyajohnson: I copied that out of the comments in threddsConfig, I think it was outdated instructions; I've deleted it from both places and replaced it with instructions to use the public directory.

As of now, they do have to put the names of the CSS files they want to use in threddsConfig.xml, but we could instead search the public directory for expected filenames - might make more sense since the only benefit the threddsConfig property adds is the ability to name the files whatever they want

Issue #1113, opened 2018-07-05
Dependency com.google.guava:guava:19.0 has a known vulnerability


davewichers:
This library as specified here: gradle/any/dependencies.gradle:libraries["guava"] = "com.google.guava:guava:19.0" is subject to the known vulnerability: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-10237. "Unbounded memory allocation in Google Guava 11.0 through 24.x before 24.1.1 allows remote attackers to conduct denial of service attacks"

I have no idea if your project uses this library in a way that makes it susceptible to this known vulnerability, but it would be good to upgrade your project to use the latest version of this library to ensure you are not.

You might also want to start using some known vulnerable library tools like OWASP Dependency Check or https://ossindex.net/ (both are free) to help you identify/avoid issues like this in the future.

lesserwhirls:
Thanks @davewichers! We address this shortly.

lesserwhirls:
I've updated our dependency for guava, and will be making a PR soon (running our full test suite). We use OWASP and run it nightly on our jenkins server, but for some reason guava was not being reported. Thanks again for the report!

PR #1114, opened 2018-07-05
update some dependencies

lesserwhirls:


lesserwhirls:
addresses Unidata/thredds#1113

PR #1117, opened 2018-07-16
Hajohns/siphon access

haileyajohnson:
Provide a Jupyter Notebook "viewer" which uses Siphon to access THREDDS datasets

### DIFF
--- a/tds/src/main/java/thredds/server/config/TdsContext.java
+++ b/tds/src/main/java/thredds/server/config/TdsContext.java
@@ -233,6 +233,16 @@ public final class TdsContext implements ServletContextAware, InitializingBean, 
             }
         }
 
+        // jupyter notebooks dir
+        File notebooksDirectory = new File(this.threddsDirectory, "notebooks");
+        if(!notebooksDirectory.exists()) {
+            if(!notebooksDirectory.mkdir()) {
+                String msg = "Couldn't create TDS notebooks directory [" + templatesDirectory.getPath() + "].";
+                logServerStartup.error("TdsContext.init(): " + msg);
+                throw new IllegalStateException(msg);
+            }
+        }
+
     /* places to look for catalogs ??
     List<DescendantFileSource> chain = new ArrayList<>();
     DescendantFileSource contentMinusPublicSource =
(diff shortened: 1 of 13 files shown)


tds/src/main/java/thredds/server/config/TdsContext.java
```diff
@@ -233,6 +233,16 @@ public void afterPropertiesSet()
             }
         }
 
+        // jupyter notebooks dir
```
haileyajohnson: This still needs a way to handle enable/disabling this feature (other than deleting the notebook folder from the content directory). We could treat '/notebook/' as a service and allow/disallow it similarly to the other services?

tds/src/main/java/thredds/server/notebook/NotebookController.java
```diff
@@ -0,0 +1,309 @@
...
+      throw new IllegalArgumentException("Argument 'catalog' cannot be null.");
+    }
+
+    Catalog catalog = getCatalog(catalogName, req);
+    Dataset dataset = catalog.findDatasetByID(datasetId);
+
+    // Get notebook
+    File responseFile = getNotebookFile(dataset, catalog);
+    if (responseFile == null) {
+      // TODO: handle
```
haileyajohnson: There are a bunch of "TODO" comments concerning warning/logging in this controller - I'm not sure how we generally handle warnings, but there are situations here, such as missing fields in the ipynb metadata, which should probably just warn and return a default rather than throwing an error.

tds/src/main/java/thredds/server/viewer/ViewerServiceImpl.java
```diff
@@ -210,6 +225,30 @@ private Access getOpendapAccess(Dataset ds) {
     }
   }
 
+  private static class JupyterNotebookViewer implements Viewer {
+    private static final String title = "Jupyter Notebook viewer";
+
+    public boolean isViewable(Dataset ds) {
+      return true;
+    }
```
haileyajohnson: Right now, TDS treats all datasets as viewable with Jupyter Notebooks as long as there is at least one ipynb file in the content directory, but we don't check for NB compatibility with the dataset until the NotebookController tries to return a file, which means some datasets might link to Jupyter Notebook viewer but not be able to return a valid notebook :/ Not sure whether to move that logic to here (but then it would execute on page load rather than download which seems wonky), or do something like return an empty or mostly empty file.

tds/src/main/java/thredds/server/config/TdsContext.java
lesserwhirls: Yes, given that it's a service, we should allow/disallow it in threddsConfig.xml. However, by default, I think it should be enabled.

tds/src/main/java/thredds/server/notebook/NotebookController.java
lesserwhirls: I think I answered this in email, but to capture it here, basically follow `thredds.server.exception.ServiceNotAllowed` and ` thredds.server.TdsErrorHandling` to get messages to show up in `threddsServlet.log`

tds/src/main/java/thredds/server/config/TdsContext.java
haileyajohnson: Notebook viewers are now treated as a service BUT the point at which allowed services are finalized is pretty late in the startup process, i.e., after the content directory is populated and after viewers are registered, so we always register a `JupyterNotebookViewer` and plop a `notebooks` folder in the content directory, and then we just don't use them if the service is disabled. Not really a problem, just a note.

PR #1127, opened 2018-07-31
fix download bug in safari; fixed 'info' link in footer to work with new /info service

haileyajohnson:
…(unsure what that was originally intended to direct to)

### DIFF
--- a/tds/src/main/webapp/WEB-INF/templates/commonFragments.html
+++ b/tds/src/main/webapp/WEB-INF/templates/commonFragments.html
@@ -37,7 +37,7 @@
     </div>
 
     <div th:fragment="footer-content">
-      <h4><a class="static" th:href="${installUrl}" th:text="${installName}"></a> at <a class="static" th:href="${hostInstUrl}" th:text="${hostInst}"></a></h4>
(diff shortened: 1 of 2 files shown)


tds/src/main/webapp/WEB-INF/templates/commonFragments.html
```diff
@@ -37,7 +37,7 @@
     </div>
 
     <div th:fragment="footer-content">
-      <h4><a class="static" th:href="${installUrl}" th:text="${installName}"></a> at <a class="static" th:href="${hostInstUrl}" th:text="${hostInst}"></a> see <a class="static" th:href="${contextPath} + '/serverInfo.html'"> Info </a></h4>
```
lesserwhirls: We want to keep the info link, as it links to a page with general and contact info for the individual server, as specified in threddsConfig.xml (for example: http://thredds.ucar.edu/thredds/serverInfo.html).

That said, in 5.0, the `ServerInfoController` was changed such that the url should reference the /info service, so the correct url should actually be something like

https://thredds-dev.unidata.ucar.edu/thredds/info/serverInfo.html

Would you mind adding that fix to this PR?

dopplershift:
Still downloads as `jupyter_viewer.ipynb.txt`. Any idea what it's returning for the mime type? My guess is Apple, in its infinite wisdom, will add `.txt` to anything `text/plain`. Apparently, the recommended mime type for the notebooks is [`application/x-ipynb+json`](https://github.com/jupyter/jupyter/blob/master/docs/source/reference/mimetype.md).

dopplershift:
Success!

PR #1128, opened 2018-08-01
fix default notebook for use in JupyterLab and other bugs

haileyajohnson:


lesserwhirls:
Updated! https://thredds-dev.unidata.ucar.edu/thredds/catalog/catalog.html

PR #1132, opened 2018-08-07
make maven pom generation sort-of-work-ish

lesserwhirls:
Currently, the `thredds-parent` maven BOM generated by gradle has an empty dependency list. This PR addresses that, although not completely. Not all dependencies are added (for example, `sping-webmvc` does not show up, although `spring-core` does?). This at least gets a partial working BOM out there, which other projects depend on (like `ncSOS`). Getting a fully filled out BOM will be important for migrating `ncSOS` to v5.0 of the codebase, but for now, this takes us a long way towards that.


lesserwhirls:
Once travis clears, I will merge this in, as I want to get a `beta3` out this morning for ncWMS to use (I have a PR in the wings to address recent API changes that broke `ncWMS`). `beta3` will be short lived, and `beta4` come out in the next day or so, so that we can have a functional `ncWMS` server in TDS 5.0 again.

PR #1135, opened 2018-08-09
Declared simple geometry coverage and added CF constants

katiekopecky:


cschroed-usgs:
Heads up to @lesserwhirls : We didn't take squash merging into account when we designed our branches, CI jobs, and contribution workflow. If there's a way to merge this without squashing commits, that would help us out. Unless I'm mistaken (It has happened before, will happen again :smile: ), I believe squash-merging would require us to re-write history in multiple branches in order to integrate the freshest changes from the Unidata repo into the USGS fork.

If squashing commits is a really high priority, please let us know so that we can try to devise different branches, CI job configurations, and update our workflow.

lesserwhirls:
Greetings @cschroed-usgs! We've typically done merge commits, so each commit in the PR will be preserved. I *think* that will work well with your current workflow.

cschroed-usgs:
@lesserwhirls sweet! That works well for us!

cschroed-usgs:
I suggest that this PR be closed because it is included in #1207 .

PR #1138, opened 2018-08-22
Prevent spaces in string names from being replaced with underscores

c3iotai:
Prevent spaces and forward slashes in string names from being replaced with underscores.

Need this for an application that accesses HDF5 files via Java which are not NetCDF files.


lesserwhirls:
Spaces and forward slashes in names are [explicitly not allowed](https://www.unidata.ucar.edu/software/thredds/v4.6/netcdf-java/CDM/Identifiers.html#CDM) as part of the Common Data Model. That said, in cases where the name has been modified, the HDF5 IOSP could store the original variable name (or other types of names) as an attribute (`hdf5_name`?).

PR #1147, opened 2018-09-10
user control of nc4iosp loading

lesserwhirls:
use the `Nc4Iosp` only if a user has explicitly enabled it in `threddsConfig.xml` or `nj22Config.xml`.


DennisHeimbigner:
I am working on the Nc4Iosp issue in a separate branch. I am returning to
it now that 1134 is fixed

lesserwhirls:
Does this mean there are portions of DAP4 that require netCDF-C?

PR #1150, opened 2018-09-11
Formatting, logging in CatalogBuilder

msdsoftware:
Cleaning up while looking into user problems accessing catalogs.


lesserwhirls:
Empty exception catches...

facepalm

Thank you for cleaning this up.

msdsoftware:
Well, they weren't entirely empty, but I personally find it much much easier to have the logger write stuff out rather than query for the errLog string. Also, digging into one of the problems I was trying to figure out, I needed the printStackTrace that had been commented out.

PR #1152, opened 2018-09-13
Update NetcdfFile intro comments

msdsoftware:
Demonstrate use of try-with-resources when creating NetcdfFile.


lesserwhirls:
Nice doc addition - thanks!

PR #1153, opened 2018-09-13
File decompression uses new class StringKeeper

craigrose:
File decompression uses new class StringKeeper to make sure a file is being handled only
by one thread at a time.  StringKeeper maintains a synchronised list of strings (in this case file paths) across threads.  When a thread wishes to access a file it can check the list to see if the file is already in use and thread wait() until a notifyAll() is received.  notifyAll() is sent when a thread finishes with a file.

This addresses the following issue:

While unzipping files other concurrent requests waiting for the file are unblocked prior to the file being unzipped causing issues trying to read half unzipped files.

The cause of the problem is a race condition between the caching system (FileCache.java) and the decompression in NetcdfFile.java. NetcdfFile.makeUncompressed() adequetely locks files across threads (by catching OverlappingFileLockException) to ensure mutiple threads do not attempt to decompress a file concurrently. However other threads trying to read from files (eg N3header.isValidFile, which results in the "not a valid CDM file" error) will not be aware that a file is still be written and end up reading wrong or no bytes. This is because the FileCache is not checking the file locks established during decompression.

### DIFF
--- a/cdm/src/main/java/ucar/nc2/NetcdfFile.java
+++ b/cdm/src/main/java/ucar/nc2/NetcdfFile.java
@@ -102,6 +102,8 @@ public class NetcdfFile implements ucar.nc2.util.cache.FileCacheable, Closeable 
 
   static private boolean userLoads = false;
 
+  static private StringLocker stringLocker = new StringLocker();
+
   // IOSPs are loaded by reflection
   static {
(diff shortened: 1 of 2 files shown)


lesserwhirls:
Thank you again, and sorry for the delay in getting this merged in!

PR #1156, opened 2018-09-20
Functional interfaces in ToolsUI, formatting

msdsoftware:
Use functional interfaces for setting up action listeners. Also did some code formatting, and some conversion of System.err.println to using logger.

One of these days I want to break this class into multiple classes to make it easier to figure out — one of these days.


lesserwhirls:
Thanks for this PR, @msdsoftware! How I would love to totally redo toolsUI in something like JavaFX, maybe even GroovyFX...but as you say — one of these days.

msdsoftware:
@lesserwhirls, I'm less interested in shifting ToolsUI to another GUI platform than I am in breaking it into more manageable, intelligible chunks. BTW, apparently effective with the release of Java 11 this week, JavaFX is a separate download.

PR #1160, opened 2018-10-16
Char valued _FillValue does not work

DennisHeimbigner:
re: e-support ticket (WWW-347958)

A char valued _FillValue attribute fails complaining of a type
mismatch with the corresponding char valued variable. It turns
out that the char valued attribute get "upgraded" to type
String, so the mismatch.

Fixing this required several sets of changes to the
Nc4Iosp.java#writeAttribute function.

1. At about line 2820, modify the type equality check for
   _FillValue to add a test for the case of variable type is CHAR
   and attribute type is String.
2. At about line 2845, switch case STRING check for _FillValue
   of type char.  If so, then use nc_put_att_text function in
   NC4prototypes.
3. if the file version is NetcdfFileWriter.Version.netcdf4 then
   use nc_put_att_string.
4. Else we have some variant of the netcdf classic format, which does not
   have a string type, so use nc_put_att_text.

Finally add a test case called TestCharFillValue.java.

### DIFF
--- a/cdm-test/src/test/java/ucar/nc2/TestCharFillValue.java
+++ b/cdm-test/src/test/java/ucar/nc2/TestCharFillValue.java
@@ -1,0 +1,83 @@
+/*********************************************************************
+ * Copyright 2018, UCAR/Unidata
+ * See netcdf/COPYRIGHT file for copying and redistribution conditions.
+ ********************************************************************/
+
+/*
+The code that demonstrates the problem can be found below. The issue is due
+to the special checks we do for fill value in Nc4Iosp, specifically around
+line 2822 and line 2841. Thanks for taking a look!
+*/
+
+package ucar.nc2;
+
+import org.junit.Test;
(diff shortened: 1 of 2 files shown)


cdm-test/src/test/java/ucar/nc2/TestCharFillValue.java
```diff
@@ -0,0 +1,83 @@
+/*********************************************************************
```
lesserwhirls: @DennisHeimbigner - sorry for the minor comment here, but could we use the netCDF-Java / TDS copyright block in this?

~~~java
 /*
  * Copyright (c) 1998-2018 University Corporation for Atmospheric Research/Unidata
  * See LICENSE for license information.
  */
~~~

DennisHeimbigner: Sure, go ahead and change it.

PR #1179, opened 2018-10-26
run travis with openjdk in addition to oraclejdk

lesserwhirls:


dopplershift:
I swear it used to be this way, I forget why now that it was removed...maybe it took awhile for openjdk8 became available?

lesserwhirls:
It looks like openjdk was removed here as the build stopped working (looks like due to a travis upgrade?):

https://github.com/Unidata/thredds/commit/e91c692cd93698c3bef465b8d935dde4e6b0f16b

Revisiting the issue dropped through the cracks, but at least it works now.

PR #1192, opened 2018-11-15
beta6

lesserwhirls:


lesserwhirls:
No code change from the last PR - just creating commit point for beta6. Will merge once Travis goes green, and will create release artifacts from Jenkins.

PR #1197, opened 2018-12-07
handle all coordinate transforms

lesserwhirls:
Make sure all coordinate transforms get added to ncss response, even if
they are not needed to compute latitude / longitude. This would include,
for example, the LatLon_Projection, which is listed as the grid_mapping
for some global GRIB datasets, but don't actually "do" a transform (no
x/y -> lat/lon by netCDF-Java), as lat/lon are explicitly constructed
from the GRIB message GSD). This ensures that we are following CF if a
grid_mapping is listed (i.e. must include the transform variable).

See https://github.com/Unidata/python-workshop/issues/372 for details.

### DIFF
--- a/cdm/src/main/java/ucar/nc2/dt/grid/CFGridWriter2.java
+++ b/cdm/src/main/java/ucar/nc2/dt/grid/CFGridWriter2.java
@@ -33,6 +33,8 @@
 
 package ucar.nc2.dt.grid;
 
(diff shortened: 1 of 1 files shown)


lesserwhirls:
Not needed in 5.0, as `CFGridCoverageWriter2` does not suffer from this bug.

PR #1198, opened 2018-12-10
fix constructor copy to pass in correct parameter

lesserwhirls:
We were incorrectly passing in the scan_geometry (`GOES` or `GEOS`) rather than the sweep_angle_axis (`x` or `y`) in the constructor copy. The net effect is that all `GEOS` projections were being treated as `GOES` projections.

### DIFF
--- a/cdm/src/main/java/ucar/unidata/geoloc/projection/sat/Geostationary.java
+++ b/cdm/src/main/java/ucar/unidata/geoloc/projection/sat/Geostationary.java
@@ -199,7 +199,16 @@ public class Geostationary extends ProjectionImpl {
    */
   @Override
   public ProjectionImpl constructCopy() {
-    return new Geostationary(navigation.sub_lon_degrees, navigation.scan_geom, xScaleFactor, yScaleFactor);
+    // constructor takes sweep_angle_axis, so need to translate between
+    // scan geometry and sweep_angle_axis first
+    // GOES: x
+    // GEOS: y
+    String sweepAxisAngle = "x";
+    if (navigation.scan_geom == GEOSTransform.GEOS) {
+      sweepAxisAngle = "y";
+    }
+
+    return new Geostationary(navigation.sub_lon_degrees, sweepAxisAngle, xScaleFactor, yScaleFactor);
   }
 
   @Override


phollemans:
That corrects things in the Geostationary class.  There's another issue in the GEOSTransform class that probably won't be discovered unless some code is directly calling the GEOSTransform constructors.  In these two constructors:

` public GEOSTransform(double subLonDegrees,
                       double perspective_point_height,
                       double semi_minor_axis,
                       double semi_major_axis,
                       double inverse_flattening,
                       String sweep_angle_axis) {
    Geoid geoid;
    if (Double.isNaN(inverse_flattening)) {
      geoid = new Geoid(semi_minor_axis, semi_major_axis);
    } else {
      geoid = new Geoid(semi_minor_axis, semi_major_axis, inverse_flattening);
    }

    init(subLonDegrees, scan_geom, geoid, perspective_point_height);
  }

  public GEOSTransform(double subLonDegrees,
                       double perspective_point_height,
                       double semi_minor_axis,
                       double semi_major_axis,
                       String sweep_angle_axis) {
    Geoid geoid = new Geoid(semi_minor_axis, semi_major_axis);
    init(subLonDegrees, scan_geom, geoid, perspective_point_height);
  }
`

you can see that the sweep_angle_axis parameter isn't being used anywhere, and the call to init() is getting the default value of the scan_geom variable which is "GEOS".  I don't have a specific test case where one of these two constructors is being called, but I'm guessing it will affect any code that tries to use these constructors.  I haven't done a search in the project to see if those constructors are called anywhere.

lesserwhirls:
Thanks @phollemans - see Unidata/thredds#1199 for the more complete fix. These classes need a good cleanup (like you had mentioned, GEOS vs GOES is an eye nightmare), but I'll address that when I port these changes to v5.0.

msdsoftware:
I see you're making commits to the 4.6 master branch. Do the same problem(s) exist with the 5.0 branch?

Wondering as I had a Panoply user ask about its handling of the Geostationary projection and the sweep angle last month. The answer I gave was probably not re-assuring.

msdsoftware:
@lesserwhirls, Looking forward to your porting these to 5.0. I had a Panoply user ask about its handling of Geostationary data and the sweep angle last month. I tried looking at the NJ source but came away somewhat mystified. And without a sample dataset to play with, I wasn't going to dive in any further.

lesserwhirls:
Yeah, these classes are mystifying, that's for sure! I just cut a 4.6.12 release, and I'm in the process of making sure all of the appropriate changes are merged into 5.0 as well, so they'll be there soon. I tested with the file @phollemans pointed me to (sweep angle axis `y`), as well as some GOES 16 files (sweep angle axis `x`), and things appear to line up visually.

PR #1205, opened 2019-01-22
experimental json-ld support

lesserwhirls:
This PR has the overall goal of exposing TDS datasets to [Google's Dataset search](https://developers.google.com/search/docs/data-types/dataset). There are two main part to this PR:

1) allow admins to create a sitemap.xml file under `tds.content.root.path/public` and expose it to HTTP Get requests.
2) map out THREDDS metadata to schema.org [Dataset](https://schema.org/Dataset) objects, encoded as json-ld, and embed them in the `head` of the generated dataset HTML pages.

This is an experimental feature, and is off by default (controlled in `threddsConfig.xml`)


lesserwhirls:
Given that this is experimental and off by default, I'll go ahead and merge it in (once things pass) and get it up on thredds-test. Once we do that, we can can see if it plays nice with google dataset search.

PR #1208, opened 2019-01-31
let EsriShapefile handle an unknown type

JohnLCaron:
Bit of a trial run with a small fix.


rsignell-usgs:
@JohnLCaron !!!

jlcaron:
In the house!

PR #1209, opened 2019-02-01
Write grib ncx index from Grib2CollectionPanel

JohnLCaron:
1) Grib ncx writing from Grib2CollectionPanel was disabled - get that working.

2) Set GribIntvFilter.isZeroExcluded default to false. Setting to true should be done on a case by case basis.
Examining CFSR again, which has lots of zero intervals, it turns out that the "average" stats are populated (as far as i checked), and of the 4 "accumulation" variables, 2 have what appears to be reasonable data.

The 4 Accumulation variables in this dataset (cdas1.t00z.pgrbh00.grib2 from 01/30/2019):
  Convective_precipitation (zero)
  Large-scale_precipitation_non-convective (zero)
  Storm_surface_runoff (not zero)
  Water_runoff (not zero)

So Ive realized this should not be the default. 

The original motivation (as i recall) was that these extra intervals make the coordinates of very large collections non-uniform. Which ones? hmm, not sure but probably including CFSR. Anyway, those will have to be revisited.


grib/src/main/java/ucar/nc2/grib/collection/Grib2CollectionBuilder.java
```diff
@@ -158,7 +158,7 @@ private boolean filterIntervals(Grib2Record gr, FeatureCollectionConfig.GribIntv
     int haveLength = intv[1] - intv[0];
 
     // discard zero length intervals (default)
```
lesserwhirls: No longer the default, so this comment can be removed, right?

lesserwhirls:
@dopplershift - is there a way to tell CLAHub that JohnLCaron and jlcaron are the same user?

@JohnLCaron - now that GRIB-1 and GRIB-2 are handled the same in terms of `excludeZero`, would you mind updating `docs/src/public/userguide/pages/thredds/GribConfigRef.md` as part of this PR?

So this block, which looks to be incorrect currently :grimacing::

~~~md
1. `excludeZero`
 
   * GRIB-1: By default, intervals of length `0` are included.
     You can choose to ignore zero length intervals by setting `excludeZero="false"`.
   * GRIB-2: By default, intervals of length `0` are excluded. 
     You can choose to include zero length intervals by setting `excludeZero="true"`.
~~~

to something like:

~~~md
1. `excludeZero`
 
    By default, intervals of length `0` are included.
    You can choose to ignore zero length intervals by setting `excludeZero="true"`.
~~~

JohnLCaron:
how do i push changes to an existing pull request? do i have to make a new pull request?

grib/src/main/java/ucar/nc2/grib/collection/Grib2CollectionBuilder.java
JohnLCaron: done

dopplershift:
The easiest way is to just fill the silly thing out again--which it looks like has been done.

lesserwhirls:
Awesome - looks good to me!

PR #1220, opened 2019-02-12
cleanup GRIB code

JohnLCaron:
Cleanup only.
All changes should not affect behavior of code.


lesserwhirls:
Looks good to me. The `Objects.equals()` method really cleaned up those gnarly ternary operators. I have a run going on Jenkins against this PR and will let you know how it goes (unless you are able to get back into Jenkins already).

lesserwhirls:
Jenkins looks good 👍

PR #1221, opened 2019-02-13
Various functional interface notation.

msdsoftware:
Working out how to edit a fork locally via Xcode and sync with GitHub. Starting out with numerous cases that can be rewritten in functional interface notation.


msdsoftware:
Worked out the variable name collision bug in the initial pull request. Also replaced println statements in the problem class with logger.info calls, seeing as a logger was declared but otherwise unused.

lesserwhirls:
That looks much cleaner, and hooray for logging messages 👍  Thanks @msdsoftware!

PR #1222, opened 2019-02-13
Try with resource

JohnLCaron:


JohnLCaron:
grib cleanup: use try-with-resource where that is feasible

lesserwhirls:
I think this one just needs a rebase

lesserwhirls:
Oops I forgot - this was included in Unidata/thredds#1223, so we're good here.

Issue #306, opened 2015-11-28
Upgrade protobuf-based indexes to proto3.


JohnLCaron:
Turns out its possible to make indexes backwards compatible (5.0 can read old indexes) but not forward compatible (<5.0 cannot read new indexes). This is because default values are not stored in proto3, so proto2 barfs when "required" fields are missing. 

i think we can modify proto2 to accept new indexes by changing eg

    required fixed32 cdmHash = 1; 

to

    optional fixed32 cdmHash = 1 [default = 0];

this would fix new versions using proto2, but old version would still break.

Can we  expect people to upgrade to 5.0 and not run 4.6 simultaneously?
or if they need to do both, to upgrade to latest 4.6 ?

there are downsides to bumping the gbx/ncx suffix, particularly if you want to continue to allow the old ones in 5.0.
a clean break is easy enough.
still, im thinking of our installs (NCDC, RDA) with a terabyte of index files.

the other option is to bag the proto3 and revert to proto2. 
we did it for the earth engine group who needs to use proto3. 
plus, it has some improvements and will be used heavily in android (i understand).

im leaning towards 

1. keeping proto3.
2. keeping gbx9 files compatible, and upgrading 4.6.x to read both versions.
3. bump to ncx4, with incompatible version; there are some improvements we can make there, i think.

the main downside is that older versions will barf on new gbx9 files. 
could also bump gbx9 (gbx10? gbxa? gbx?), but see about accepting both in 5.0.

JohnLCaron:
Ive discovered that 

1) its possible to be forwards compatible, by modifying the proto2 description as above. this allows proto2 processor to tolerate missing values.

2) its possible for proto3 library to generate proto2 messages.  Im not sure if i realized this and decided against it for a reason TBD, or if its continuing confirmation of advancing senility. 

So im going to try keeping gbx9 in proto2. This solves any compatibility problems.

Then think if ncx should do the same or increment and take advantage of proto3 (ie leave ncx3 in proto2 and ncx4 in proto3).

JohnLCaron:
Im seeing some improvements to be made in ncx:

Maintain seperate GDS. see #309.

adding packing = true will save some space.

lesserwhirls:
@JohnLCaron - I like the idea of keeping the .gbx9 in proto2, especially given the heavy lifting it takes to recreate those, unless there are some really nice advantages to moving those to proto3. 

[...]

PR #1229, opened 2019-02-15
Proto version

JohnLCaron:
upgrade protobuf to 3.6.1
regenerate grib protos by hand.

### DIFF
--- a/gradle/any/dependencies.gradle
+++ b/gradle/any/dependencies.gradle
@@ -184,7 +184,7 @@ libraries["jsoup"] = "org.jsoup:jsoup:1.11.2"
 // Annotations
 libraries["jsr305"] = "com.google.code.findbugs:jsr305:3.0.2"
 
-libraries["protobuf-java"] = "com.google.protobuf:protobuf-java:3.5.1"
+libraries["protobuf-java"] = "com.google.protobuf:protobuf-java:3.6.1"
 
 libraries["visad"] = "edu.wisc.ssec:visad:2.0-20130124"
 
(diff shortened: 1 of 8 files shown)


lesserwhirls:
Will this require a regeneration of .ncx files for those running a 5.0 beta?

JohnLCaron:
Thats a good question, let me make sure that the answer is no.

JohnLCaron:
So this problem affects both grib1Index.proto / Grib1Index.java and grib2Index.proto / Grib2Index.java.

10/16/15 we upgraded protos to proto3.
11/28/15 we added proto_version field so we could deal with any changes between proto2 and proto3.
12/1/15 we reverted to use proto2 and changed the name to version (not proto_version).

Issue #306 captures the thought processes for all that.

This PR was just intended to upgrade the protobuf library. However, when I forced recompiled the protos with the new version of protoc, I got a compile error indicating the GibXIndex.java was still referencing proto_version field. That could only work if the protos checked in on 12/1/15 had never been recompiled.

We could probably assume that no one else recompiled them, else it would have failed and likely they would have reported it. But who knows if there are versions out there with a version field but not a proto_version field. Since they both use field number #4, the binary form is compatible.

The field isnt actually needed, nor used anywhere. So Im going to remove it from the code. Ill reserve that field in the proto so we never use it. If we decide we need something, we can just create a new field with a new field number.

Note that this implies protos are not being recompiled as part of the build process. We might want to look at that harder.

Current best practice is to never use "required" modifier. If we had done that, we could upgrade to proto3 with ease. We can revisit proto3 upgrades another time.

lesserwhirls:
Ok, just finished the Jenkins run and this fixes four more tests - that take us back to where we were with 8 failing tests. Once this is rebased, I'll merge it in.

PR #1231, opened 2019-02-19
move VariableRenamer out of grib package

JohnLCaron:


lesserwhirls:
How much do you use the GribVariableRenamer functionality? Given the history of it, and why it came about, `5.0` seems like a good time for it to go into the new `legacy` package, unless you see a need?

JohnLCaron:
Its not used anywhere except in ToolsUI. If you are sure no one wants it there, we can move it to legacy.

lesserwhirls:
As far as I know, the IDV is the only one using this. If we keep it in `ui`, then `ncIdv` will need to depend on the `ui` module...eventually...`ncIdv` already depends on `legacy`, so we'd be covered there. I'd be in favor of removing it from `ToolsUI` completely.

lesserwhirls:
Hmmm. I tried using the "Update branch" button, which worked, but it added a merge commit. Not sure I like that, given that you would need to update your branch before adding anything (if something came up that you needed to fix). I'd love to be able to handle something minor (potentially) without the need to ping users for one more push on each PR, but I'm not sure there is a way around it :-/

dopplershift:
The "Update branch" button comes from this setting:

image

The problem is that you don't know that different merged branches don't conflict with each other (at the runtime level) until your tests fail in master.

lesserwhirls:
It'd be nice to have the option to do an auto-rebase. Like, check if a rebase can be done cleanly, provide an updated PR, and rerun the tests all without the need for the contributor to do the rebase locally and force push, or without needing me to pull down the PR branch, rebase locally, and force push over someone's PR branch (if enabled). I've force pushed over someones PR branch before, partially because I through github would not let me, but it did and wow did that feel dirty.

PR #1234, opened 2019-02-20
suppress false positives on dependency check

lesserwhirls:


lesserwhirls:
quartz is being misidentified as jenkins, which is a known issue: 

https://github.com/jeremylong/DependencyCheck/issues/1637#issuecomment-451115272

same as Unidata/thredds#1133, but for 5.0

PR #1236, opened 2019-02-21
Begin de-cluttering ToolsUI

msdsoftware:
Trying to clean up ToolsUI.java, starting with moving some code out into separate classes. If this pull request is accepted I'll keep on keeping on.

A problem is that the splash window is appearing blank at ToolsUI launch, but is okay when the Logo item is selected from the Help menu. However this happens with every other compiled copy of ToolsUI that I have on hand.


lesserwhirls:
This looks great to me! On my machine, I see the splash screen for a split second, which is what I see on the main 5.0 branch. I get the full splash screen on 4.6, so not quite sure what's going on there. Would you mind rebasing this PR and I'll get it merged in?

msdsoftware:
@lesserwhirls, A couple issues. I've made further progress with my local copy, but had not yet posted an update to GitHub because of a couple issues to work out. And then I saw that the VariableRenamer commit conflicted with this. I wasn't sure what to do about that. Looks like I need to learn how to do a rebase, as you suggest.

lesserwhirls:
I can do the rebase on my end if you'd like. Once your code is where you'd like it, let me know. I can even make a screencast showing you what I do so that you can see a rebase in action. It's not that they are terribly difficult, but the first few times can be a little disorienting and it really helped me to watch someone do it first. Thank you for your contributions!

msdsoftware:
@lesserwhirls, Why don't you go ahead and do it. I've just reversed all my local updates. Sorting out a couple problems I was running into was looking to take more time to figure out.

What I'd like to do is separate out the various `OpPanel` subclasses from `ToolsUI.java`. But so many of them are calling `ToolsUI` methods that it's going to require extra effort.

msdsoftware:
Added some code to the splash and about window code to handle the possible case of a multi-screen set-up. But unable to figure out the problem with initially blank splash screen despite trying multiple possible fixes.

lesserwhirls:
Just picked up your last commit and added it to Unidata/thredds#1242. Going to go ahead and close this one and get the new one merged in. Thanks @msdsoftware!

Issue #330, opened 2015-12-11
'HDF Error' when subsetting file


dopplershift:
The attached file yields:

```
NetCDF: HDF error
```

It looks like #301 on the surface, but the fix for that does not fix this problem. 

This query string allowed me to reproduce locally (my first attempt):

```
COMPRESS_LEV2_20140201000000-GLOBCURRENT-L4-CURekm_15m-ERAWS_EEM-v02.0-fv01.0.nc?var=eastward_ekman_current_velocity&north=79.8750&west=-140&east=179.8750&south=-79.8750&horizStride=1&time_start=2014-02-01T00%3A00%3A00Z&time_end=2014-02-01T00%3A00%3A00Z&timeStride=1&accept=netcdf4
```

dopplershift:
[COMPRESS_LEV2_20140201000000-GLOBCURRENT-L4-CURekm_15m-ERAWS_EEM-v02.0-fv01.0.nc](https://files.zenhub.io/566b5c0e9b2939d368c188f6)
Uploaded using [**ZenHub.io**](http://www.zenhub.io)

JohnLCaron:
It reads ok in toolsUI.

seems to write ok in Writer tab, though i didnt subset. (5.0).

 I assume you get the error from ncss when writing to netcdf4?

dopplershift:
Even have the problem on 5.0 asking for netcdf4, I get:
```IOException sending File NetCDF: HDF error```

Here's the query:
```
COMPRESS_LEV2_20140201000000-GLOBCURRENT-L4-CURekm_15m-ERAWS_EEM-v02.0-fv01.0.nc?var=eastward_ekman_current_velocity&north=79.8750&west=-140&east=170&south=-79.8750&horizStride=1&time_start=2014-02-01T00%3A00%3A00Z&time_end=2014-02-01T00%3A00%3A00Z&timeStride=1&accept=netcdf4
```

JohnLCaron:
i have a fix for this, do you want it on 4.6? obviously needs to be run
through unit tests, so depends if you have already started the release

cwardgar:
Nope, haven't finalized 4.6.4 yet (barely). So yeah, go ahead and submit a PR. You need a full Jenkins run for this, not just Travis?

JohnLCaron:
on second look, its not as easy to fix on 4.6.
so im leaving it off for now.
will fix on 5.0, retrofit later if called for.

JohnLCaron:
[...]

PR #1238, opened 2019-02-22
Chunksize can be read from an attribute now.

Soman2:
Uncommented the part of the code which allows Chunksize set up from a variable attribute for NetCDF format.

### DIFF
--- a/cdm/src/main/java/ucar/nc2/write/Nc4ChunkingDefault.java
+++ b/cdm/src/main/java/ucar/nc2/write/Nc4ChunkingDefault.java
@@ -73,10 +73,10 @@ public class Nc4ChunkingDefault extends Nc4ChunkingStrategy {
 
   @Override
   public long[] computeChunking(Variable v) {
-    /* check attribute
+
     int[] resultFromAtt = computeChunkingFromAttribute(v);
     if (resultFromAtt != null)
-      return convertToLong(resultFromAtt); */
+      return convertToLong(resultFromAtt);
 
     int maxElements = defaultChunkSize / v.getElementSize();
 


lesserwhirls:
So these lines were commented out as part of a [PR](https://github.com/Unidata/thredds/commit/0e1bef2d3c74a0b558d987ee7dae751238471645) to fix issue Unidata/thredds#330. As part of that, some tests were written to test the fix. This PR does not break any of our tests, so that's certainly a plus. That said, I know that some core netCDF-4 attributes, like `_ChunkSizes`, are kind of special and are handled differently that other generic attributes. @DennisHeimbigner - what's your take?

DennisHeimbigner:
Ideally, this should be consistent wit the netcdf-c library behavior. Without looking in detail,
I am guessing that setting (as opposed to reading) the chunksize from attributes
is only supported by ncgen .cdl files.

DennisHeimbigner:
I am reluctant to merge this PR because it appears to conflict with the behavior 
of the netcdf-c library. In that library, _ChunkSizes is a virtual attribute -- meaning it
does not exist as an explicit attribute in the dataset.  What is the use case for this?

Soman2:
In my project I want to use your library to write down the file with ~100000x60000 size and set up custom ChunkSize( ~4000x4000). As for know I can not set up any ChunkSize.

lesserwhirls:
I see. Currently, the use of the `_ChunkSizes` attribute to control how data are written out is not supported (our documentation is out of date on this). This is probably something we can support again in the future, but it will take some work. I think we see an issue when trying to write a subsetted dataset using the `_ChunkSizes` defined in the non-subsetted dataset, but I'm not totally sure.

What you can do currently, however, is create a class that implements `ucar.nc2.write.Nc4Chunking` for your particular dataset and needs.  Once you've implemented that, you should be able to pass it into  `FileWriter2` or `NetcdfFileWriter` using one of the signatures that accepts `Nc4Chunking`. A bit more work that setting an attribute and being done with it, for sure, but until we get that fixed, this should get you up and running.

PR #1239, opened 2019-02-22
fix broken link to grib paper

lesserwhirls:
These links are good in the 5.0 docs, so just 4.6.


lesserwhirls:
Will merge and update website once the redirect is in place.

PR #1245, opened 2019-02-28
 ToolsUI de-clutter. Peel off menus to separate classes.

msdsoftware:
Further reduce amount of code in ToolsUI.


lesserwhirls:
This is great - thank you @msdsoftware! If you think it's ready to go, I'll update the branch via github and merge it in!

msdsoftware:
@lesserwhirls, Go ahead. I'll post more PRs on this whenever I get another neat chunk done.

BTW: Spent some time working on the blank splash issue. Found that on my laptop, the splash appears about 600 ms after the call is made to create and show it, but if left on screen, it takes around two seconds (two mississippis) before the logo graphic is painted in the window.

lesserwhirls:
Interesting - is it that intensive of a thing to take 2 seconds to paint the window?

msdsoftware:
@lesserwhirls, It shouldn't be. The splash screen in my own apps has no such problem. There has to be something else going on that I don't understand.

PR #1252, opened 2019-03-07
ToolsUI de-clutter. Split all OpPanels out to separate classes.

msdsoftware:
This is a pretty big pull request. It takes all 47 of the OpPanel inner classes in ToolsUI, splits them off as individual classes, ands puts them all in `ucar.nc2.ui.op`. Another 33 associated panels, tables and viewers that were in `ucar.nc2.ui` are also moved down to `ucar.nc2.ui.op`.

Some hooks were added to ToolsUI because some of the OpPanel classes need to interact with ToolsUI. I'll likely re-visit this later to see if there's a cleaner way of handling this interaction.

This PR also provides a fix for the blank ToolsUI splash window. It's not that I am entirely sure what's going on, and what I've done is a bit of a kludge. But it does get the splash logo to render almost immediately as the splash window is shown.

There's also a ton of code format re-styling. Hopefully this improves readability where done.

Upshot is that ToolsUI.java is now 1600 lines, whereas it was around 6000 when I started on this crusade.


lesserwhirls:
This looks great to me! The main toolsUI class was a bit out of hand, to say the least. Thank you for going on the crusade 👍 If you feel this is a good point to merge, I'll go ahead and update it and merge it in.

msdsoftware:
@lesserwhirls, I think it's okay to merge. I wouldn't have posted a pull request if I thought otherwise.

It would be nice if someone else went through to perform a sanity check, but the number of classes I split off to separate files, I realize that's not particularly realistic.

lesserwhirls:
I noticed that you had pushed a few commits since you opened the PR, so I just wanted to make sure. I've taken a look at your changes, pulled them down and played a bit, and it looks good to me. That said, I'm do not do much GUI work, so I might not be the best to review. @hvandam2 - would you be interested in taking a look since you've done some AWT work in the past (am I remembering that right?)?

msdsoftware:
@lesserwhirls, My later commits were pretty much noodling around over the weekend. Rewriting addition of adding listeners to use functional interfaces, and some copyright comments.

The ToolsUI interface should be unchanged as far as the user is concerned, with the exception that I tried to fix the problem with the splash screen initially appearing blank. As noted above, my solution is a bit of a kludge — specifically the creation of a `FontMetrics` object to force some GUI initialization before attempting to display the splash, but also the 2.5 second sleep.

msdsoftware:
BTW, one thing that had me scratching my head. The following appears in `Tools.main`.

```
           ClassPathXmlApplicationContext springContext = new ClassPathXmlApplicationContext(
                                "classpath:resources/nj22/ui/spring/application-config.xml");

```

I see a lot of imports of `org.springframework.*` classes in the `thredds.*` packages, but couldn't see where spring was necessary to ToolsUI. Commenting out that line, compiling and running didn't seem to make any difference, i.e., nothing seemed broken). But perhaps I needed to attempt to connect to a remote server?

ui/src/main/java/ucar/nc2/ui/GetDataRunnable.java
```diff
@@ -10,7 +10,7 @@
 /**
  *
  */
-interface GetDataRunnable {
+public interface GetDataRunnable {
```
hvandam2: This comment applies to the next one also: Is this interface/class being used beyond the enclosing package?  I assume that is why the public modifier is added.

hvandam2: Never mind.  I see why.

msdsoftware: @hvandam2, The `public`is there because NCdumpPanel implements GetDataRunnable.  But NCdumpPanel was moved from being an inner class of ToolsUI to being a separate class in `ucar.nc2.ui.op`, along with all the other classes extending OpPanel.

msdsoftware: Snap! Looks like we ere typing at the same time.

hvandam2: Looks good, I think we should leave the org.springframework classes for a separate change set.  I don't see a need for them, but we should look into it in detail.  I am going to test for a couple of days.

hvandam2: Tested on both mac and windows, everything looks great!@mdsoftware Thank you for the excellent work on this.@lesserwhirls Let's get this merged and in, Howard

PR #1253, opened 2019-03-16
Remove System.out.println in GRIB module.

JohnLCaron:
OK in main, otherwise remove.


lesserwhirls:
Looks good to me - loggers for the win!

PR #1256, opened 2019-03-22
Misc UI class code clean-up.

msdsoftware:
This is a batch of code re-formatting, println-to-logging changes, import cleansing, functional interface re-writes, etc. in various ToolsUI classes. The bulk of this is for readability, in one coder's opinion.

Hardly any re-factoring in here except for one class where an if-then-else-else-else-etc was re-worked as a switch-case.


msdsoftware:
And in case you're wondering about all the `final` insertions, in this pr and previously, it's a habit that arose from shutting up a static code analyzer.

msdsoftware:
I'm confused. More recent 5.0 code commit(s) were merged to this pull request, and now Travis is all unhappy?

lesserwhirls:
The artifacts server was exactly the problem. I have been out all week, and the artifacts server has been bonkers. I restarted the build and it all looks good. Thanks @msdsoftware!

msdsoftware:
TY, @lesserwhirls.

PR #1260, opened 2019-04-09
Ecmwf tables

JohnLCaron:
Use latest WMO tables.
Add ECMWF grib2 tables into resources (not actually used yet)


lesserwhirls:
Overall looks good to me. Did the `.table` files come from `ecCodes`, or did you generate those? I have code that generates parameter tables from the `ecCodes` `.def` files for `GRIB-1`, and those tables are what is used by the `GRIB-1` code for ecmwf.

I had started down the path of using the `.def` files from `ecCodes` for `GRIB-2` to make similar parameter tables, but I ran into an interesting situation. If I recall correctly, there were cases where two parameters would have the same `discipline`, `category`, and `parameter number`, but would have slightly different "names" (e.g. the keys in the `name.def` files). For example, there would be nearly duplicate entries in a given `name.def` file, but with one difference - level information was included in one of those entries which made them distinguishable (and the name would mention something about the level). That did not seem to map very well to the way we do `GRIB-2` parameter tables. I wondered if it was an error in the way those `.def` files were being generated. Have you encountered any weirdness like that? These `.def` files look a bit different, so maybe it's not an issue anymore?

dopplershift:
I think this PR holds a record for the biggest one I've ever seen.

lesserwhirls:
Chrome struggles to load the "Files Changed" page. There are a bunch of def files, tables, from ecCodes, with only a bit of java code changes and a markdown file. 

dump truck

dopplershift:
Yeah, I looked. Just blew my mind that it almost had more lines added than digits available in a 32-bit float.

JohnLCaron:
The .table files are in the ecCodes downdown, I did no processing. I think
it has all versions of grib2 since the beginning (?)

I am going to analyse them to see how different they are from each other
and from WMO. I will probably delete all the other versions, but I wanted
to get them into the repo to make it easier to work, and in case anyone
else wants to look at them.

If you prefer, I could put them in some other place than grib/resources.

lesserwhirls:
If the code will use the `.table` files provided by `ecCodes` directly, then I'd say that's a win and they should live where they are (otherwise maybe they should live in `grib/src/main/sources/`). If you're good to go, then I'm happy to merge things in as-is. Thanks!

Issue #925, opened 2017-09-26
Netcdf writeAttribute NullPointerException


antoinetran:
In netcdf4-4.6.4 and even latest 4.6.10 (https://github.com/Unidata/thredds/blob/v4.6.10/netcdf4/src/main/java/ucar/nc2/jni/netcdf/Nc4Iosp.java):

Writting an Attribute with not null name and null values creates a NullPointerException at values.getStorage(). In my case:
`ret = nc4.nc_put_att_int(grpid, varid, att.getShortName(), Nc4prototypes.NC_INT, new SizeT(att.getLength()), (int[]) values.getStorage());`

I am not sure a null values is Netcdf compliant, but according to the spec https://www.unidata.ucar.edu/software/netcdf/docs/attribute_conventions.html §"missing values", this is something that can happen. I can even write a null values with Netcdf4 Java with ` attribute = new Attribute(name, dataType, false);`

DennisHeimbigner:
The method nc_put_att_int is not part of the Nc4Iosp
interface. I infer then that you are actually writing to
an instance of Nc4prototypes. This is not a good idea
unless you are an expert. I think using the  method
Nc4Iosp.makeAttribute would be a better option.,

DennisHeimbigner:
The code does not do a null check. The question is: should it just silently return,
or should it generate some other kind of exception?

DennisHeimbigner:
The same bug appears in version 5 also.

DennisHeimbigner:
Ok, I believe the correct solution is that if values == null, then treat it like |values|==0.
|values|==0 should work.

antoinetran:
> Ok, I believe the correct solution is that if values == null, then treat it like |values|==0.
|values|==0 should work.

If values is null, it would be better to make it null at the end.

> The code does not do a null check. The question is: should it just silently return,
or should it generate some other kind of exception?

No, because it is authorized to have null values for an Attribute, if I read correctly the Netcdf spec.

DennisHeimbigner:
> No, because it is authorized to have null values for an Attribute, if I read correctly the Netcdf spec.
I am not sure how you are inferring this. The underlying netcdf-c library and spec has no notion of NULL
value. It does have _FillValue (aka "missing value") but that is a different thing altogether.
Can you elaborate on what you mean by null values.

antoinetran:
> I am not sure how you are inferring this. The underlying netcdf-c library and spec has no notion of NULL
[...]

PR #1262, opened 2019-04-17
Allow null valued attributes to be written

lesserwhirls:
Addresses Unidata/thredds#925

Before, Array.getStorage() would throw an NPE if the array was null.
Now, we handle the possibility of the NPE, and pass the null value
to netCDF-C (via JNA) and allow the C library to handle the null
pointer.

### DIFF
--- a/cdm/src/main/java/ucar/nc2/iosp/IospHelper.java
+++ b/cdm/src/main/java/ucar/nc2/iosp/IospHelper.java
@@ -726,10 +726,13 @@ public class IospHelper {
   // convert char array to byte array
 
   static public byte[] convertCharToByte(char[] from) {
-    int size = from.length;
-    byte[] to = new byte[size];
-    for (int i = 0; i < size; i++)
-      to[i] = (byte) from[i]; // LOOK wrong, convert back to unsigned byte ???
+    byte[] to = null;
+    if (from != null) {
+      int size = from.length;
+      to = new byte[size];
+      for (int i = 0; i < size; i++)
+        to[i] = (byte) from[i]; // LOOK wrong, convert back to unsigned byte ???
+    }
     return to;
   }
 
--- a/cdm/src/main/java/ucar/nc2/iosp/hdf5/H5header.java
+++ b/cdm/src/main/java/ucar/nc2/iosp/hdf5/H5header.java
@@ -942,7 +942,7 @@ public class H5header extends NCheader
     // check for empty attribute case
     if (matt.mds.type == 2) {
       if (dtype == DataType.CHAR)
-        return new Attribute(matt.name, ""); // empty char considered to be a 0 length string
+        return new Attribute(matt.name, DataType.STRING); // empty char considered to be a null string attr
       else
         return new Attribute(matt.name, dtype);
     }
--- a/netcdf4/src/main/java/ucar/nc2/jni/netcdf/Nc4Iosp.java
+++ b/netcdf4/src/main/java/ucar/nc2/jni/netcdf/Nc4Iosp.java
@@ -558,8 +558,16 @@ public class Nc4Iosp extends AbstractIOServiceProvider implements IOServiceProvi
           case Nc4prototypes.NC_UBYTE:
             att = new Attribute(attname, DataType.UBYTE);
             break;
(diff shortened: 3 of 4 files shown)


lesserwhirls:
Ok, I added more fixes here, but this uncovered an inconsistency in our HDF5 iosp code. When we read in an HDF5 file (in this case, netCDF-4) and the `H5header` class creates attributes, `CHAR` types are handled differently than other data types for the null case.

The case of a null valued netCDF attribute with a CHAR datatype, results in the creation of a `String` attribute with an empty string as its value. All other attribute types will keep their null value. Specifically, see here:

https://github.com/Unidata/thredds/blob/38aaac1d1a873d3d93b268e4715f5aa43e7ec66f/cdm/src/main/java/ucar/nc2/iosp/hdf5/H5header.java#L942-L947

It seems to me like we should not be making a special case here. If we really want to make the `CHAR` attributes into `String` attributes, then I think we should keep the null value for consistency rather than reassign it to an empty string. For example, ` return new Attribute(matt.name, "")` would become ` return new Attribute(matt.name, DataType.STRING)` on line 945.  Because of the way things currently are, the test for a null char attribute is a different than the others. See here:

https://github.com/Unidata/thredds/pull/1262/files#diff-43f9cdcb162e8122ab6cfaba063bb7baR149

Thoughts @DennisHeimbigner? Can you recall any conversations as to why `CHAR` is special here?

DennisHeimbigner:
IN thinking about this, I have a question. The real issue here is that the attribute's
length is zero (as inferred from a NULL value or a zero length list of values).
So upon reading, should we return NULL or a zero length list of values?

lesserwhirls:
I'd say it depends on the HDF5 attribute's [DataspaceMessage](https://support.hdfgroup.org/HDF5/doc1.8/H5.format.html#DataspaceMessage) value. For a value of 0 or 1, then maybe we should return a zero length list of values. However, since the `DataspaceMessage` value is 2, then I think we should return null (which is consistent with the attribute types with `DataspaceMessage=2`.)

DennisHeimbigner:
So I looked at the netcdf-4 code in the netcdf-c library.
In that code, if the attr length is zero, then it just returns the length
and makes no attempt to do anything about returning a value.
This means that technically, the returned data ptr is undefined
if the attribute length is zero. Should we change that to return NULL?

lesserwhirls:
If we set it to return `new Attribute(matt.name, DataType.STRING)`, then the value associated with the attribute will be null. That's what I am leaning towards.

lesserwhirls:
I just pushed a new commit with those changes.

lesserwhirls:
Ok, I think I have it this time. The changes were causing some tests in our full test suite to fail. Basically, when reading netCDF-3 through the c library, we need to treat empty char attribute values as an empty string value in the CDM - for netCDF-4, we need to treat empty char attribute values as null. I'll run the tests one more time just to be sure.

DennisHeimbigner:
I added a test to TestCharFillValue.java using this creation code
and it fails as advertised. https://github.com/AlexInTransition, Is this code close
to what you are using?
````
try (NetcdfFileWriter ncfw =
                         NetcdfFileWriter.createNew(NetcdfFileWriter.Version.netcdf4, fileName);) {
                Dimension charDim = ncfw.addDimension("charDim", 3);
                Variable charVar = ncfw.addVariable(charVarName, DataType.CHAR,
                        charDim.getFullName());
                Array charArrayFillValue =
                        ArrayChar.makeFromString("", 0);
                Attribute charAttrFillValue;
                    charAttrFillValue = new Attribute("_FillValue",
                            charArrayFillValue);
                charVar.addAttribute(charAttrFillValue);
                ncfw.create();
            } catch (IOException e) {
                e.printStackTrace();
            }
````

DennisHeimbigner:
OK, I am not sure how you want the fix, so I am attaching a zip file
containing a testcase (I modified an exiting one) and a fix to nc4iosp.
[pr1262.zip](https://github.com/Unidata/thredds/files/3168290/pr1262.zip)

lesserwhirls:
Thanks @DennisHeimbigner! I just made a PR with the fixes and tests.

PR #1265, opened 2019-04-24
Refactor WMO tables, separating param, code, and flag with different interfaces.

JohnLCaron:
Get tables from ECMWF eccodes version 2.12.0 (not used in production yet), and compare to WMO tables.


JohnLCaron:
Now that I see its passing tests, let me clean this up a bit before you review it.

JohnLCaron:
ok, good enough to go...

lesserwhirls:
Looks good to me, but I do have one question. The `truth` lib - is that getting us something that `JUnit` does not provide, or do you have future uses of `truth` that you foresee? It seems like the `assertNotNull`, `assertTrue`, and `assertEquals` methods from `org.junit.Assert` would cover what you are checking.

However, in general the capabilities of `truth` look nice from what I've seen in the docs. If we keep it, we should probably change the scope from `compile` to `testCompile` in `dependencies.gradle` so that we don't add anything to the bulk of our artifacts.

JohnLCaron:
Good questions. The fluent syntax is better I think. Its probably better for more complicated things, which the examples I have so far are not. Really, its just that we use it at Google, so Im just following that. 
Heres a link I found comparing to AssertJ: 

https://google.github.io/truth/comparison.html

Its up to you if you prefer to forgo. Not hard at this moment to revert to vanilla JUnit.
Defo should be testCompile.

PR #1277, opened 2019-05-04
Add eccodes Grib2Table, with ecmwf + 10 other centers. 

JohnLCaron:
These are used only for local parameters.

Add functionality in Grib2TableViewer to examine the overriding of standard parameters (overriding not used in production).

Tweek Grib2Table API, eg Grib2Table.getTableValue -> getCodeTableValue
Rename StringUtil2.remove -> StringUtil2.removeAll


JohnLCaron:
Ok this one is ready to review.
It adds eccode localConcept's local params, but doesnt allow overriding standard params. So things arent any worse then they are now.

Ill be looking at the possibility of standard param overrides. probably a big refactor, because we have assumed that params are independent of levels and  times values. 

Look at the localConcepts defs, which are reflected in Iosp/Grib2/Grib2Tables - choose an eccodes, then "show low level details" from context menu.

No plans to touch grib1.

PR #1278, opened 2019-05-07
Grib2 table cleanup2

JohnLCaron:
oops, got isLocal wrong again.


lesserwhirls:
Looks good to me. I think I came to the same spot in wondering how to deal with parameters that also depended on `level` (I don't recall anything depending on `time` at the time). Oh GRIB...

PR #1279, opened 2019-05-09
ToolsUI cleanup

JohnLCaron:
Mostly minor fixes from IntelliJ code inspections.
Restored FeatureScanOpPanel, with minor fix that was throwing an Exception. Not sure how it got removed, but seems to be a one-off.


msdsoftware:
As many changes as I was making a few months go into ToolsUI to break code out into separate classes, _stuff_ may have happened.

JohnLCaron:
yeah, stuff happens ;^)
thanks for your help cleaning it up.

JohnLCaron:
Ok, enough ui for now.

msdsoftware:
Just noticing the indenting changes in OpPanel. Is that an IntelliJ thing to have indents be 2 chars rather than 4?

Asking because I had discovered in the past that indenting of various classes was 2 chars instead of the 4 that I expected.

PR #1282, opened 2019-05-12
Break ui module into uibase (non-cdm) and uicdm (cdm dependent)

JohnLCaron:
Do some standard intelliJ inspection cleanup.
First pass at getting unit tests working.

PR #1285, opened 2019-05-16
Divide the ui icons into base and cdm.

JohnLCaron:
Rearrange the ui resources directories.


lesserwhirls:
Bring out the earth movers! Looks good to me. This also includes Unidata/thredds#1282. BTW - we're getting close on the netCDF-Java breakout - I'll let you know when it's ready.

PR #1322, opened 2020-03-28
http --> https correction

oxelson:
@lesserwhirls 
http --> https protocol correction for unidata website.  Ignored/skipped references containing or pertaining to:

- xmlns
- dtd
- schema
- namespace 
- obvious code invocations


tds/src/main/webapp/WEB-INF/classes/resources/xsl/ncssSobs.xsl
```diff
@@ -161,7 +161,7 @@
             </div>
         </form>
         <hr/>
-        <h3><a href="http://www.unidata.ucar.edu/software/netcdf-java/stream/CdmRemote.html">CDM Remote Documentation</a></h3>
+        <h3><a href="httpss://www.unidata.ucar.edu/software/netcdf-java/stream/CdmRemote.html">CDM Remote Documentation</a></h3>
```
lesserwhirls: Looks like there is an extra `s` in this one. Also, the link should be https://www.unidata.ucar.edu/software/netcdf-java/v4.6/reference/stream/NcStream.html (that one that is there 404s).

lesserwhirls:
Thanks @oxelson! In addition to the extra `s`, we have a new test failure:

```bash
thredds.client.catalog.TestResolve1 > testResolve FAILED
    java.lang.AssertionError: https://www.unidata.ucar.edu/
        at thredds.client.catalog.TestResolve1.testResolve(TestResolve1.java:70)
```

which is due to this:

```java
docUrl.equals("http://www.unidata.ucar.edu/")
```

I'm thinking there might be more like that once we run our full tests on Jenkins.

tds/src/main/webapp/WEB-INF/classes/resources/xsl/ncssSobs.xsl
oxelson: D'oh.  Sorry.  Fixed in commit https://github.com/Unidata/thredds/commit/6b07552d90aeedc063fbb3d4a8115dabfdefa768

oxelson:
@lesserwhirls, fixed the assert docUrl test to use https.  Passing travis now.  Understood about the possibility of more failures on jenkins.

lesserwhirls:
Thanks @oxelson! I'll merge this is and start a run on Jenkins to see how things go.

lesserwhirls:
Actually, going to run current master on jenkins first, to get a baseline.

PR #1337, opened 2020-10-08
Correctly handle GINI negative calibration values

lesserwhirls:
Backport of Unidata/netcdf-java#488

GINI calibration values can be negative (for the data value range), and
GEMPAK uses the left most bit of the stored integer to designate the
signedness. Fixes Unidata/netcdf-java#480.

Minor addition - since the `long_name` attribute was there and so close,
go ahead an add the `standard_name` attribute to the `x` and `y`
coordinate variables.


lesserwhirls:
Travis passed:

https://travis-ci.org/github/Unidata/thredds/builds/733949999

Not sure why the status has not updated though.

PR #1338, opened 2020-10-08
Use the data_word_size from the level 2 message

lesserwhirls:
Backport of Unidata/netcdf-java#514

Previously, the level 2 code used a fixed word size of 8 bits for the
data blocks, with a special exception for the differential phase
product, which was hardcoded to use 16 bits. This PR enables the IOSP
and Level2 Record parser to use the actual word size as encoded in the
message.


lesserwhirls:
Travis passed:

https://travis-ci.org/github/Unidata/thredds/builds/733962676

dopplershift (approved):

Issue #1342, opened 2020-10-22
EOFException on NCDump data


Mikhail-Aoun:
Hi,
We've found an issue when trying to read data from some particular netcdf files. We met an EOFException in the ucar.nc2.iosp.hdf5.H5header$GlobalHeap instantiation algorithm.
The bug can be reproduced with ToolsUI latest maintenance release (Version 4.6.15) and this attached netcdf file as example :
[SGA1-RO_-00-SRC_C_EUMT_20200704124757_G_O_20200704124200_20200704124459_O_N____.zip](https://github.com/Unidata/thredds/files/5421475/SGA1-RO_-00-SRC_C_EUMT_20200704124757_G_O_20200704124200_20200704124459_O_N____.zip)

When you open this file and try to NCDump tm_packets data, it will raise the following exception : 

> byte tm_packets(number_tm_packets=10256, *);
>   :long_name = "Array of TM packets";
>   :_Unsigned = "true";
> 
>  data:
> java.io.EOFException
> 	at ucar.unidata.io.RandomAccessFile.readInt(RandomAccessFile.java:1244)
> 	at ucar.unidata.io.RandomAccessFile.readLong(RandomAccessFile.java:1325)
> 	at ucar.nc2.iosp.hdf5.H5header.readLength(H5header.java:4950)
> 	at ucar.nc2.iosp.hdf5.H5header$GlobalHeap.<init>(H5header.java:4772)
> 	at ucar.nc2.iosp.hdf5.H5header$HeapIdentifier.getHeapObject(H5header.java:4688)
> 	at ucar.nc2.iosp.hdf5.H5header.getHeapDataArray(H5header.java:4525)
> 	at ucar.nc2.iosp.hdf5.H5header.getHeapDataArray(H5header.java:4519)
> 	at ucar.nc2.iosp.hdf5.H5iosp.readData(H5iosp.java:285)
> 	at ucar.nc2.iosp.hdf5.H5iosp.readData(H5iosp.java:206)
> 	at ucar.nc2.iosp.hdf5.H5iosp.readData(H5iosp.java:144)
> 	at ucar.nc2.NetcdfFile.readData(NetcdfFile.java:2026)
> 	at ucar.nc2.Variable.reallyRead(Variable.java:874)
> 	at ucar.nc2.Variable._read(Variable.java:845)
> 	at ucar.nc2.Variable._read(Variable.java:886)
> 	at ucar.nc2.Variable.read(Variable.java:709)
[...]

PR #1343, opened 2020-10-22
Fix for EOFException on NCDump data Unidata/thredds#1342

Mikhail-Aoun:
On GlobalHeap instantiation, we should stop reading file data just after we met a HeapObject.id equals to 0. Otherwise, we could be at the end of the file and a EOFException will be raised. Addresses the github issue Unidata/thredds#1342.


lesserwhirls:
For some reason we have not been getting update from travis-ci, but [the tests passed](https://travis-ci.org/github/Unidata/thredds/builds/737984374). I've also ran your PR against our full test suite and everything looks good. As I said over at Unidata/netcdf-java#533, your work on finding and fixing this bug is greatly appreciated!

PR #1368, opened 2021-08-19
fix rewrite multi sub group ncfile error

zhaosd:
Thank you for your contribution! Please be sure to open your PR against the `4.6.x` branch :-)

### DIFF
--- a/cdm/src/main/java/ucar/nc2/iosp/netcdf3/N3header.java
+++ b/cdm/src/main/java/ucar/nc2/iosp/netcdf3/N3header.java
@@ -682,8 +682,8 @@ public class N3header {
         if (largeFile)
           raf.writeLong(pos);
         else {
-          if (pos > Integer.MAX_VALUE)
-            throw new IllegalArgumentException("Variable starting pos="+pos+" may not exceed "+ Integer.MAX_VALUE);          
+          if (pos > Long.MAX_VALUE)
+            throw new IllegalArgumentException("Variable starting pos="+pos+" may not exceed "+ Long.MAX_VALUE);
           raf.writeInt((int) pos);
         }
 
--- a/cdm/src/main/java/ucar/nc2/Dimension.java
+++ b/cdm/src/main/java/ucar/nc2/Dimension.java
@@ -231,12 +231,15 @@ public class Dimension extends CDMNode implements Comparable {
     Dimension other = (Dimension) oo;
     Group g = getGroup();
     if ((g != null) && !g.equals(other.getGroup())) return false;
-    if ((getShortName() == null) && (other.getShortName() != null)) return false;
-    if ((getShortName() != null) && !getShortName().equals(other.getShortName())) return false;
+    if ((getShortName() == null) || (other.getShortName() == null)) return false;
+
+    String fullName = NetcdfFile.makeValidCdmObjectName(getFullName());
+    String otherFullName = NetcdfFile.makeValidCdmObjectName(other.getFullName());
     return (getLength() == other.getLength()) &&
             (isUnlimited() == other.isUnlimited()) &&
             (isVariableLength() == other.isVariableLength()) &&
-            (isShared() == other.isShared());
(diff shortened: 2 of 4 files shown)


lesserwhirls:
Thank you for your PR! I have one comment, but I also wanted to see if you knew that this repo contains code for netCDF-Java 4.6.x, while the latest release is [5.4.x](https://github.com/unidata/netcdf-java/tree/maint-5.x)?

cdm/src/main/java/ucar/nc2/iosp/netcdf3/N3header.java
```diff
@@ -682,8 +682,8 @@ void writeHeader(int extra, boolean largeFile, boolean keepDataStart, Formatter
         if (largeFile)
           raf.writeLong(pos);
         else {
-          if (pos > Integer.MAX_VALUE)
-            throw new IllegalArgumentException("Variable starting pos="+pos+" may not exceed "+ Integer.MAX_VALUE);          
+          if (pos > Long.MAX_VALUE)
+            throw new IllegalArgumentException("Variable starting pos="+pos+" may not exceed "+ Long.MAX_VALUE);
```
lesserwhirls, line 686: Was this change necessary to support subgroup writing?

zhaosd: Our team rewrite some grib2 files(file size is bigger than 110MB). This can throw the exception and I see the pos type is long. So I make this change. The output nc files is bigger than 550MB.

lesserwhirls: Ah, ok. So your use case is converting grib2 to netCDF using the `netcdf4/src/main/java/ucar/nc2/rewrite/Rewrite.java`  class?

zhaosd: Yes, our code as follow
image

zhaosd: Import code:
import ucar.nc2.rewrite.Rewrite;

zhaosd: Dependency library:

lesserwhirls: Ok, if that's your goal, then try using [ucar.nc2.FileWriter2](https://docs.unidata.ucar.edu/netcdf-java/4.6/javadoc/ucar/nc2/FileWriter2.html). For larger GRIB files, you will likely not be able to use the `netcdf3` format. We have a pure Java writer for the netCDF 3 classic format, which has a limitation of only supporting 32-bit file offsets. I would suggest writing to one of the netCDF-4 formats (`NetcdfFileWriter.Version.netcdf4_classic` or `NetcdfFileWriter.Version.netcdf4`), which would look something something like this:

~~~java
public static void GribToNc(String filePath) throws IOException {
  File file = new File(filePath);
  NetcdfFile gribFile = NetcdfFile.open(file.getPath());
  String ncFilePath = file.getParent() + File.separator +
          file.getName().substring(0, file.getName().lastIndexOf(".")) +
                  ".nc";
  FileWriter2 writer = new FileWriter2(gribFile, ncFilePath, NetcdfFileWriter.Version.netcdf4, null);
  writer.write();
}
~~~

Writing a netCDF-4 format file requires having access to the netCDF-C library (built with netCDF-4 support). For information on how to have netCDF-java use the netCDF-C library for writing, please see https://docs.unidata.ucar.edu/netcdf-java/4.6/userguide/reference/netcdf4Clibrary.html#loading_preferred. This page is a little out of date, and you no longer need to build HDF5 or netCDF-C in any special way, but the loading instructions are still valid.

If you are using a build system like maven or gradle, you will need the `cdm`, `netcdf4`, and `grib` artifacts. If you are building with `netcdfAll.jar`, those are included.

